PR #12, opened 2023-03-22
Datatable component

kellyel:
This PR implements a datatable component that can be used on the site and on score reports. It takes an object of raw data and a constructed object of column definitions. In these definitions, the user can define several behaviors for the columns such as sorting and filtering behavior. 

This PR also implements this table as the table that appears on the query run reports listing. 

Future planned improvements include: 
- optional column display/hiding
- allowing slots for each column


src/components/DataTable.vue
```diff
@@ -0,0 +1,208 @@
...
+  let options = [];
+  _forEach(props.data, entry => {
+    if(!options.includes(entry[field])){
+      options.push(entry[field]);
+    }
+  });
+  return options
+}
+
+function prepareData(entry){
```
richford, line 192:
```suggestion
function prepareDates(entry){
```

Perhaps this function name would be more explicit about what the function is doing.

src/components/DataTable.vue
```diff
@@ -0,0 +1,208 @@
...
+  columns: {type:Array, required: true},
+  data: {type: Array, required: true},
+  allowExport: {type: Boolean, default: true}
+});
+
+let selectedRows = [];
+const refSelectedRows = ref(selectedRows);
+
+const exportCSV = async () => {
+  const csv = Papa.unparse(_flatMap(refSelectedRows.value, prepareData));
```
richford, line 109:
```suggestion
  const csv = Papa.unparse(_flatMap(refSelectedRows.value, prepareDates));
```

See comment below about function name.

src/components/DataTable.vue
```diff
@@ -0,0 +1,208 @@
...
+  const a = document.createElement('a');
+  a.href = URL.createObjectURL(blob, { type: 'text/plain' });
+  a.download = 'runs.csv';
+  document.body.appendChild(a);
+  a.click();
+  document.body.removeChild(a);
+};
+
+const exportFullCSV = async () => {
+  const csv = Papa.unparse(_flatMap(refData.value, prepareData))
```
richford, line 120:
```suggestion
  const csv = Papa.unparse(_flatMap(refData.value, prepareDates))
```

See comment below about function naming.

src/components/DataTable.vue
```diff
@@ -0,0 +1,208 @@
...
+            :options="refOptions[col.field]" 
+            placeholder="Any"
+            :showToggleAll="false" 
+            class="p-column-filter" 
+          />
+          <Calendar 
+            v-if="col.dataType === 'date' && !col.useMultiSelect"
+            v-model="filterModel.value" 
+            dateFormat="mm/dd/yy" 
+            placeholder="mm/dd/yyyy"
```
richford, line 57:
```suggestion
            dateFormat="mm/dd/yyyy" 
            placeholder="mm/dd/yyyy" 
```

Should these two date formats match up?

richford (changes requested):
This is a big improvement in usability! I added some suggestions inline. In addition, I have three other comments:

- I'm worried about name collision with the PrimeVue datatable, which we register globally inside of `main.js`. Can we EITHER drop that global component registration and import only inside of the new `DataTable` component OR rename the component you've written to `RoarDataTable.vue` or some such?
- When I test the demo website, I cannot get any of the date filters besides "Date is" to work. Can you check that out and see if you can reproduce on your end?
- Lastly (probably this is a "phase 2" comment), can we also allow the user to specify a binary `Checkbox` filter type, as we did in the old DataTable for the `completed` column?

richford:
The latest commits look good to me. Is this ready to go from your point of view?

richford (approved):

PR #18, opened 2023-06-22
MVP: Participant View Part I

kellyel:
This is a WIP PR. The dev team has decided to merge this branch back into main as a way of synchronizing our work currently two different branches. This will not happen in the future, as this is only possible since we have no deployment dependent on main right now.

This PR includes a lot of changes, including bugs that are currently being worked on and will be fixed before merging part II.


dguimara (approved):
Looks good

PR #27, opened 2023-08-10
Use pinia subscribe to detect auth status in SignIn.vue

richford:
In my local testing, this PR, along with [roar-firekit PR 32](https://github.com/yeatmanlab/roar-firekit/pull/32) resolves the reauthentication issue that we were experiencing yesterday.

### DIFF
--- a/.firebaserc
+++ b/.firebaserc
@@ -3,7 +3,7 @@
     "default": "gse-roar-admin"
   },
   "targets": {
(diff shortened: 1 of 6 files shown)


richford:
@kellyel, can you test on your end and confirm?

kellyel (approved):
Looks great! ⭐

PR #38, opened 2023-08-18
Add Orgs as dropdown options on the mass uploader

kellyel:
This PR primarily makes the mass uploader more user-friendly in line with requests from partnerships. 

In this PR:
- Remove orgs dropdown section from the mass uploader
- Adds orgs category in MU dropdowns
- Adds support for checking whether an org is valid (against orgs the user is allowed to create for) before submitting.
- If not valid, added to the error users table with relevant message.
- Adds 'ignore this column' as dropdown item.


kellyel:
@richford Since this PR deals with studies, We should wait to merge this until your 'study -> group' PR merges, then I will update this one.

richford (approved):

PR #48, opened 2023-08-21
Add CreateAdministrator component and allow login via email link

richford:
This PR enables creation of admin user accounts and authentication of those user accounts using email link login.

It is WIP because it will require an update to the version of roar-firekit once [the corresponding firekit PR](https://github.com/yeatmanlab/roar-firekit/pull/43) is merged.

PR #49, opened 2023-08-21
Form Validation

kellyel:
This PR implements form validation for the Create Admin form. It also adds a minimum validation (just two text fields) on Create Orgs. The additional validation for Create Orgs can be completed post-MVP, as partnerships requested a different form (org mass-uploader) for org creation.


richford:
Just as with #52, we've elected to roll these changes into #48 in order to avoid the merge/rebase headache.

PR #54, opened 2023-08-22
Pass id and abbreviation to register students methods

richford:
This PR requires that [roar-firekit PR47](https://github.com/yeatmanlab/roar-firekit/pull/47) be merged and a new version of firekit is published before it will work


kellyel (approved):
⭐

PR #64, opened 2023-08-25
CSS adjustments for GameTabs

dguimara:
Adjust font sizing, some spacing, and a few other small tweaks for enhancement


richford (approved):

richford:
Merging now so that it is ready for Carrie to demo. But @dguimara I was wondering if we could keep the image size the same. For this PR it appears that both the text and the image size increased. I'll open another issue for that.

PR #65, opened 2023-08-25
Fix Clever SSO

richford:
This PR:
- [x] Fixes SSO workflows that originate from the ROAR dashboard
- [x] Fixes Clever SSO workflows that originate from the clever dashboard

There is still one failure mode that I haven't resolved. If user logs in from Clever dashboard, but then signs out and attempts to sign in again using Clever, the page will hang.

This PR depends on [roar-firekit PR 49](https://github.com/yeatmanlab/roar-firekit/pull/49)


richford:
@kellyel , this is now ready for review. I rebased it on top of #73 so we should merge that first. It also requires [roar-firekit PR49](https://github.com/yeatmanlab/roar-firekit/pull/49) to be merged first.

So the PR link will fail until we merge [roar-firekit PR49](https://github.com/yeatmanlab/roar-firekit/pull/49), cut a new version of firekit and update package.json in this PR.

If you'd like to test locally before that, you can do so with `npm link` on the `fix/doc-listeners` branch of roar-firekit.

kellyel (approved):
⭐

PR #72, opened 2023-08-27
Fix refresh bug

kellyel:
I believe the issue here is that we were relying on the `$subscribe` method solely. `$subscribe` is fantastic for waiting for changes to firekit to trigger the refresh function. However this fails when firekit is already initialized, such as navigating to a page well after the user has logged in. 

This PR adds a simple check to `onMounted()`, where we check whether firekit is ready. If so, we want to call refresh and start gathering data. If not, simply do nothing and wait for `$subscribe` to take care of it.

### DIFF
--- a/src/components/CreateAdministration.vue
+++ b/src/components/CreateAdministration.vue
@@ -307,6 +307,15 @@ if (allVariants.value.length === 0) {
   assessments.value = [allVariants.value, []];
 }
 
+onMounted(async () => {
+  if (roarfirekit.value.getVariants && roarfirekit.value.isAdmin()) {
+    await refreshAssessments();
+  }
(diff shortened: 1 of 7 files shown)


richford (approved):

richford:
Great!!! Thanks so much for solving this and taking the time to reason through the root cause. I pushed a commit to define `roarfirekit` iin `ListOrgs.vue` and `AdministrationProgress.vue`. This is good to go! I'll merge once I test the preview link again. :shipit:

richford (approved):

richford:
Okay, I spoke too soon: had to add this `onMounted` trick to `CreateAdministration.vue` as well. Now, I believe it's good to go.

PR #74, opened 2023-08-27
Correctly grab administration info in authStore

kellyel:
This PR fixes grabbing the admin data from the participant view. This error manifested in a breakdown of the sequential logic, which relies on the field grabbed from this call

### DIFF
--- a/src/store/auth.js
+++ b/src/store/auth.js
@@ -64,7 +64,7 @@ export const useAuthStore = () => {
       },
       async getAdministration(administration) {
         try {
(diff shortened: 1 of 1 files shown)


kellyel:
It also should be noted that post-MVP, I think I want to analyze how we use auth store and query store. I think Adam and I built a lot of 'get' type functions that overlap in functionality. Adam's are more robust and are properly titled 'query'. But again these are post-MVP goals

richford (approved):

PR #75, opened 2023-08-28
Add variant info tooltips

richford:
This PR adds variant info tooltips to let the user know how to interact with the variant info buttons.


kellyel (approved):
Nice UX change!

PR #81, opened 2023-08-29
Enh/admin sidebar

dguimara:
This PR does two things:

1. A simple enhancement to make the admin sidebar menu collapsible. a future enhancement should look at a new placement for this (or to use a different Primevue component) so that it does not get in the way.  
2. Removes the "Your Info" section of sidebar. Let's rethink what this should say – showing a list of all districts, classes, etc... will be too information intense.


richford:
Closing in favor of #244

PR #91, opened 2023-09-01
Return boolean value

kellyel:
I believe this will fix the incorrect invalid dates message while preserving the guard against single date entries.


richford:
Oh gosh, that makes me feel pretty silly

PR #99, opened 2023-09-02
Abort a run before unmounting the game components

richford:


richford:
This PR resolves https://github.com/yeatmanlab/roar/issues/6 and https://github.com/yeatmanlab/roar/issues/21

richford:
@kellyel, I wound up intercepting the back button and reloading instead. It's hacky, and I think we should find a way to do it without using `window.event`, which is deprecated. But it works for now.

To do it without `window.event`, we'll have to create a `popstate` event listener. But then we'll also have to destroy that listener in all the other possible ways to leave the page.

Zio-4 (approved):

PR #106, opened 2023-09-07
Fix clever first login

richford:
This PR hard-codes a delay on the consent modal to allow the firestore use doc to be updated with assignedAssignments. Hacky but works for now.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,7 +8,7 @@
       "name": "roar-dashboard",
       "version": "0.0.0",
       "dependencies": {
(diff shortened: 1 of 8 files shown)


richford:
Note that the PR preview link will not work because Clever is currently redirecting to localhost. To test it out, you'll have to pull to your local machine and test with localhost.

kellyel (approved):
The code looks great. Will want to test this (maybe via huddle- as I do not have access to the newest, live version of the clever app)

PR #111, opened 2023-09-08
Handle Multiple Admins on Participant

kellyel:
This PR changes the logic of participant view to cope with multiple administrations being assigned to the user. The dropdown appears when more than one administration is detected, and allows the user to swap between the admins at will. 

You can test a user with multiple admins using `labtest1` and `password1`. To test a user with just one admin, use any clever student assigned to the BRSSD admin via clever. 

I tried to test this in most scenarios but I would appreciate if you tested this one a bit more rigorously as essentially every user will be touching this functionality.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,7 +8,7 @@
       "name": "roar-dashboard",
       "version": "0.0.0",
       "dependencies": {
-        "@bdelab/roar-firekit": "^3.12.6",
+        "@bdelab/roar-firekit": "^3.12.7",
         "@bdelab/roar-letter": "^1.1.2",
         "@bdelab/roar-pa": "^1.1.1",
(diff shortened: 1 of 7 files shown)


kellyel:
Completely forgot, this relies on firekit [#64 ](https://github.com/yeatmanlab/roar-firekit/pull/64), so preview will not work until we cut a new minor version of that!

richford (approved):
Looks great! It's a big relief to me to have this in here. I made a few changes:
- bumped the firekit version as requested
- saved `selectedAdmin` in a new game store so that it survives refresh
- added a label to the assignment dropdown

richford:
I tested pretty extensively on my end as well.

richford:
And tested again just to be sure. Merging. Thanks!

PR #116, opened 2023-09-11
Add a key to router view

richford:
Resolves https://github.com/yeatmanlab/roar/issues/45

@kellyel, can you take a look to make sure this doesn't have any unintended consequences. I tested with username/password students and some Clever users (make sure to test with Demo district users because we have live Clever users now.

For the above linked issue, I was able to reproduce on Chrome but not Brave. Logging out and then trying to log in **as another user** did not work. Same user was okay. With this change, it forces a replacement of the `<router-view>` element/component every time a navigation event occurs. See https://dev.to/brylie/reload-router-view-when-vue-route-changes-3cck

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -14,7 +14,7 @@
   <div>
     <Toast />
     <Navbar v-if="!navbarBlacklist.includes($route.name)" />
-    <router-view />
+    <router-view :key="$route.fullPath" />
   </div>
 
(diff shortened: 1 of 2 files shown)


kellyel (approved):
I have tried logging in and out several times in several ways. this change does not have unintended consequences, as for as I can see. Approved.

PR #126, opened 2023-09-14
Implement chunking for register student

kellyel:
We encountered a rate limit with creating more than x number of students a second from a single IP. So this PR introduces a timed chunking system for registration. Currently, 10 students will start the create process every second. I am open to discussion on the actual rate.

### DIFF
--- a/src/pages/RegisterStudents.vue
+++ b/src/pages/RegisterStudents.vue
@@ -422,7 +422,6 @@ async function submitStudents(rawJson) {
 }
 
 function delay(milliseconds){
-  console.log('delay function has been called')
   return new Promise(resolve => {
       setTimeout(resolve, milliseconds);
   });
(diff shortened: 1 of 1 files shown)


src/pages/RegisterStudents.vue
```diff
@@ -317,102 +318,114 @@ function submitStudents(rawJson) {
...
+      }).catch((e) => {
+        toast.add({ severity: 'error', summary: 'User Creation Failed', detail: 'Please see error table below.', life: 3000 });
+        addErrorUser(user, e)
+        if(processedUsers >= totalUsers){
+          activeSubmit.value = false;
+        }
+      })
     })
-  })
+    await delay(1000)
```
richford, line 420: I suggest padding to reduce possible overlap due to cloud function queueing instability.
```suggestion
    await delay(1250)
```

richford (changes requested):

richford (approved):

PR #140, opened 2023-09-21
Fix for game image resizing issues

dguimara:
A few changes to fix the image resizing/cropping problem for game images.


jyeatman:
As far as I can tell this fixed the issue (though the games in my admin were old). Taggine @richford to review and merge

richford:
@dguimara, I tested this with username `carrie-test-lab` (same password). Your fix is good for smaller screens, however, the size is still variable for larger screens.

PR #147, opened 2023-09-29
Use emulator ports only in npm run dev:emulate

richford:


kellyel (approved):
Looks great! :shipit:

PR #155, opened 2023-10-10
Add class column to the progress and scores table 

kellyel:
This PR adds the initial version of classes on the progress and scores tables, as well as porting the schools column over to the progress table. 

Upon further conversation, I am not sure we are quite happy with the implementation of classes, namely what should be displayed in the cell when multiple classes are listed for a user. The solution might also include having one thing listed, and have another included when the datatable is exported.


kellyel:
Outdated

PR #172, opened 2023-10-26
Reroute to sign-in if email-link auth is taking too long

richford:
This PR mitigates the failure of email-link sign in. I tried to replicate this issue on my localhost and could only get it to reproduce perhaps 10% of the time. This PR does not actually fix the issue but rather tries to mitigate the failure mod by setting a timer and redirecting to sign-in if the auth workflow is taking too long. This reuses the pre-existing error message.

This isn't an ideal solution but is a good enough bandaid for now.


kellyel (approved):
Decent bandaid solution for now!

PR #173, opened 2023-10-26
Paginate the progress table

kellyel:
This PR
- Adds pagination to the progress table
- Implements export selected and export all
- Refactor scores fetchers to general assignment fetchers- make including scores optional
- Sync keys between the two pages
  - assignmentCounter should use same key
  - scores and assignments use different keys- including scores and not including scores respectively

### DIFF
--- a/src/helpers/query/assignments.js
+++ b/src/helpers/query/assignments.js
@@ -203,7 +203,7 @@ export const assignmentCounter = (adminId, orgType, orgId) => {
   })
 }
 
-export const scoresPageFetcher = async (adminId, orgType, orgId, pageLimit, page, skinny = false, paginate = true) => {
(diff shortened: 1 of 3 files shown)


src/helpers/query/assignments.js
```diff
@@ -203,7 +203,7 @@ export const assignmentCounter = (adminId, orgType, orgId) => {
   })
 }
 
-export const scoresPageFetcher = async (adminId, orgType, orgId, pageLimit, page, skinny = false, paginate = true) => {
+export const assignmentPageFetcher = async (adminId, orgType, orgId, pageLimit, page, includeScores = false, skinny = false, paginate = true) => {
```
richford, line 206: It appears that `includeScores` is not used. Shouldn't the `scoresPromises` be wrapped in `if (includeScores)` or something like that?

richford (changes requested):

richford (approved):

PR #175, opened 2023-10-27
Use display percentile and standard score in table but base the color…

richford:
… coding on the raw percentile


kellyel (approved):
Looks good, thank you for taking care of this!

PR #177, opened 2023-10-30
Roar Vocab

ksmontville:
Added Roar Vocab to Roar Dashboard.


src/router/index.js
```diff
@@ -67,6 +67,12 @@ const routes = [
     component: () => import("../components/tasks/Multichoice.vue"),
     meta: { pageTitle: "Multichoice" }
   },
+  {
+    path: "/game/vocab",
+    name: "Vocab",
+    component: () => import("../components/tasks/SWR.vue"),
```
kellyel, line 73: I think this wants to point at `../components/tasks/Vocab.vue`

kellyel (approved):

PR #184, opened 2023-11-02
[Enh] CSS Adjustments

ksmontville:
Adjust css on admin sidebar button so that text is centered, and remove some unused css classes.


kellyel:
Thanks

kellyel (approved):

PR #186, opened 2023-11-03
WIP: Use tanstack/query to fetch parent orgs in the create-orgs page

richford:


ksmontville (changes requested):
I think that the styling looks a little awkward when there is only a single school to assign to a new class.

I propose retaining the dropdown styling regardless of the length of returned list of schools (I can commit my changes for this if you'd like).

Screenshot 2023-11-08 133824

Screenshot 2023-11-08 133900

richford:
You're absolutely right. Can you commit your changes for this (to both the districts and schools dropdowns)? Thanks!

PR #188, opened 2023-11-06
Implement defaulting behavior for displayNames

kellyel:
Implements defaulting behavior when tasks don't have a proper entry in the displayNames object

### DIFF
--- a/src/pages/ScoreReport.vue
+++ b/src/pages/ScoreReport.vue
@@ -449,7 +449,8 @@ const displayNames = {
   "multichoice": { name: "Multichoice", order: 6 },
   "anb": { name: "ANB", order: 7 },
   "mep": { name: "MEP", order: 8 },
-  "morphology": { name: "Morphology", order: 9 }
+  "mep-pseudo": { name: "MEP-Pseudo", order: 9},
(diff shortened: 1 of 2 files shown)


src/pages/ScoreReport.vue
```diff
@@ -443,6 +447,9 @@ const displayNames = {
   "sre": { name: "Sentence", order: 5 },
   "letter": { name: "Letter", order: 1 },
   "multichoice": { name: "Multichoice", order: 6 },
+  "anb": { name: "ANB", order: 7 },
+  "mep": { name: "MEP", order: 8 },
+  "morphology": { name: "Morphology", order: 9 }
```
richford, line 452:
```suggestion
  "anb": { name: "ANB", order: 7 },
  "mep": { name: "MEP", order: 8 },
  "mep-pseudo": { name: "MEP-Pseudo", order: 9},
  "morphology": { name: "Morphology", order: 10 },
```

richford (changes requested):

richford (changes requested):
We need to port these same changes over to the progress view

richford (approved):

PR #228, opened 2023-12-12
Lock assigments after the task is completed

kellyel:
Corrects an issue where RockRest students are able to play games after the task has already been completed


lucasxsong (approved):
LGTM

PR #234, opened 2023-12-14
Enh/score report charts

lucasxsong:
image

Adds two histograms to an update tabview score report. 
- grouped histogram by support level, displayed over grade.
- histogram with binned percentiles, displayed over grade. 

Makes structural changes to score report to reduce wordiness: putting task descriptions, histograms, and subscore graphs into tabbed components.


src/components/reports/chartSpecs.js
```diff
@@ -0,0 +1,212 @@
...
+    }
+    return value;
+  } else {
+    throw new Error('Index out of range');
+  }
+}
+
+function returnSupportLevelValues(scores, mode) {
+  let gradeCounts = returnGradeCount(scores);
+  let values = [];
```
richford, line 85: Same here. Can be const
```suggestion
  const gradeCounts = returnGradeCount(scores);
  const values = [];
```

src/components/reports/chartSpecs.js
```diff
@@ -0,0 +1,212 @@
...
+  // generates values for bar chart
+  for (let grade of gradeCounts) {
```
richford, line 87: Same as above
```suggestion
  for (const grade of gradeCounts) {
```

src/components/reports/chartSpecs.js
```diff
@@ -0,0 +1,212 @@
...
+
+  return gradeCount;
+}
+
+function returnValueByIndex(index, grade, mode) {
+  if (index >= 0 && index <= 2) {
+    // 0 => needs extra support
+    // 1 => needs some support
+    // 2 => at or above average
+    let valsByIndex = [
```
richford, line 55:
```suggestion
    const valsByIndex = [
```

src/components/reports/chartSpecs.js
```diff
@@ -0,0 +1,212 @@
...
+    }
+    if (mode === 'count') {
+      value = {
+        category: grade.grade,
+        group: valsByIndex[index].group,
+        color: valsByIndex[index].color,
+        value: grade?.support_levels[index],
+      };
+    }
+    return value;
```
richford, line 77: Suggestion: return the value in each `if` block rather than saving to the `value` variable. Also, it doesn't appear that `valsByIndex[index].color` is defined so I removed those lines in this suggestion.

```suggestion
    if (mode === 'percentage') {
      return {
        category: grade.grade,
        group: valsByIndex[index].group,
        value: (grade?.support_levels[index] / grade.totalStudents) * 100,
      };
    }
    if (mode === 'count') {
      return {
        category: grade.grade,
        group: valsByIndex[index].group,
        value: grade?.support_levels[index],
      };
    }
    return value;
```

richford (changes requested):

richford:
Another stylistic change request:
- [ ] Can you format the mouseover percentages to whole numbers and add a "%" sign.

PR #253, opened 2023-12-24
Optimize REST batchGet queries

richford:
Many of the batchGet requests had duplicated entries. This PR first uses `_uniq` to reduce the size of the batchGet document lists (sometime reducing to a third of the original size). This also required some downstream changes to the fetchers since the batchGet document array were no longer the same length as the arrays for assessments/runs/scores. Practically, this meant using `_find` instead of `_zip`.

This PR is based on top of #248


kellyel (approved):
Excellent!

PR #261, opened 2024-01-06
Fix/admin show details wrap

lucasxsong:
this pr slightly modifies the structure of the administration card component to render groups under the administration as a full container width. whereas small width screens would cause the previous table to clip, this update makes the table a bit more responsive. 

image


richford:
Task linked: [CU-86870748j Admin Card Show Details Improvements](https://app.clickup.com/t/86870748j)

kellyel (approved):
Looks great!

PR #273, opened 2024-01-11
Update Score Reports and Distribution Charts

lucasxsong:
Makes some changes to the score reports including...

- Change at a glance charts from bar charts to pie charts
- Graphs and tabs are ordered in the preferred order (Phoneme, Word, Sentence)
- TaskReport DistributionByGrade can now be facetted by schools for district level reports
- Improve graph consistency
- Improve graph readability (increase label size, increase graph size)

This PR also slightly optimizes some of the fetchers in the query folder. (task, run)


src/pages/ScoreReport.vue
```diff
@@ -802,6 +840,14 @@ const runsByTaskId = computed(() => {
   });
 });
 
+const sortedTaskIds = computed(() => {
+  const taskIdsInOrder = ['pa', 'swr', 'sre', 'letter', 'cva', 'fluency', 'mep', 'morph', 'multichoice', 'vocab'];
```
kellyel, line 844: This list shouldn't be needed, I recommend utilizing `taskDisplayNames` imported from `helpers/report.js`. For each task there's a taskId, display name, and ordering number. The reason I recommend this is so all of our ordering is in once place and consistent.

src/pages/ScoreReport.vue
```diff
@@ -771,25 +787,47 @@ function scoreFieldAboveSixth(taskId) {
   return 'percentile';
 }
 
+const descriptionsByTaskId = {
```
kellyel, line 790: Not essential, but this could be another candidate for being moved to the helper file. Just to reduce bloat in this file and make the text easily accessible for editing.

src/pages/ScoreReport.vue
```diff
@@ -5,46 +5,75 @@
     </aside>
     <section class="main-body">
       <div>
-        <div class="flex-col p-4 drop-shadow-lg">
+        <div class="">
```
kellyel, line 8:
```suggestion
        <div>
```

kellyel (changes requested):
Overall I think this is looking really good! I've made a few suggestions below

richford (changes requested):
Requesting changes that we talked about in person:
- linting/formatting
- Using `@` and absolute component paths on import
- Removing commented out html in templates.

src/pages/ScoreReport.vue
lucasxsong: Good recommendation! I will make the change.

richford (approved):
Thanks for the new changes. I approve. Let's wait for @kellyel to sign off on the changes as well.

kellyel (approved):
Looks good, Thanks!

PR #283, opened 2024-01-12
Filter out tasks on the exclude lists

kellyel:
This PR:
- Adds CVA as a named task
- creates a task exclude list for score reports. Items added will not be shown on score reports.


kellyel:
Related to https://github.com/yeatmanlab/roar/issues/150 and https://github.com/yeatmanlab/roar/issues/151

richford (approved):

PR #291, opened 2024-01-16
Make distribution graphs super-admin only again

kellyel:
This PR changes Score Reports so that the new graphs are a feature for super admins only. The graphs above the table are hidden, and the graphs below the table are replaced with the old text when a non-super admin views the report.
Screenshot 2024-01-16 at 2 18 57 PM
Screenshot 2024-01-16 at 2 18 49 PM


lucasxsong (approved):
LGTM!

PR #301, opened 2024-01-23
Moving Menu on TOP

Emily-ejag:
Adding PvMenu, deleting all Administrator Sidebars


richford (changes requested):
@Emily-ejag, I love the change! But I'm experiencing an issue where the menu bar disappears when I refresh the page. Is this happening for you?

src/components/NavBar.vue
```diff
@@ -18,30 +26,50 @@
...
+
+const { data: userClaims } = useQuery({
+  queryKey: ['userClaims', authStore.uid],
+  queryFn: () => fetchDocById('userClaims', authStore.uid),
+  keepPreviousData: true,
+  enabled: initialized,
+  staleTime: 5 * 60 * 1000, // 5 minutes
+});
+
+ const isSuperAdmin = computed(() => Boolean(userClaims.value?.claims?.super_admin));
```
richford, line 52:
```suggestion
const isSuperAdmin = computed(() => Boolean(userClaims.value?.claims?.super_admin));
```

src/components/NavBar.vue
```diff
@@ -7,8 +7,16 @@
         </div>
       </router-link>
       <div class="login-container">
-        <!-- <i class="pi pi-bars menu-icon" @click="toggleMenu" /> -->
-        <!-- <PvMenu ref="menu" id="overlay_menu" :model="dropdownItems" :popup="true" /> -->
+        <div v-if="isSuperAdmin">
```
richford, line 10: Also, I think we can remove the surrounding div that restricts this to super admins. The `getSidebarActions` functions used below takes care of restricting the menu items that partner admins can see. But I think we still want them to see it.

src/components/NavBar.vue
```diff
@@ -18,30 +26,50 @@
...
+});
+
+ const isSuperAdmin = computed(() => Boolean(userClaims.value?.claims?.super_admin));
+
+
+const isAtHome = computed(()=>{
+  return router.currentRoute.value.fullPath === '/'
+});
+const dropDownActions = computed(() => {
+  const rawActions = getSidebarActions(authStore.isUserSuperAdmin, !isAtHome.value);
```
richford, line 59: Let's use the computed value that you defined above.
```suggestion
  const rawActions = getSidebarActions(isSuperAdmin.value, !isAtHome.value);
```

richford (changes requested):

richford:
I think that implementing the suggestions in my second review might resolve the refresh issue that I brought up in my first review.

Emily-ejag:
Yes, that solved it :)

richford:
Task linked: [CU-8687342dk 1. Moving the menu button to the nav bar (using PvMenu instead of Menu to maintain consistency with the rest of the site)](https://app.clickup.com/t/8687342dk)

src/components/NavBar.vue
richford, line 10: I still think that we need to remove this div because we want partner admins to see the menu as well.

richford (changes requested):

src/components/NavBar.vue
Emily-ejag: changed!

richford (approved):

PR #302, opened 2024-01-23
Fix/dist charts ordering and toggles

lucasxsong:
This PR adds some updates to the distribution charts, namely:
- Sort school facetted chart by lowGrade attribute -- elementary schools should come first, then middle schools, and finally high schools. 
- Add toggle to facetted chart to display data based on raw score as well as percentile. Toggle defaults to raw score.
- Conditionally render `At a Glance` score legend for administrations that have `At a Glance` charts.

### DIFF
--- a/src/assets/styles/components/selectbutton.scss
+++ b/src/assets/styles/components/selectbutton.scss
@@ -38,3 +38,15 @@
 .p-selectbutton.p-invalid > .p-button {
   border-color: var(--red-100);
 }
+
+.select-button .p-button {
+  margin: 0px;
+}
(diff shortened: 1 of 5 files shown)


richford:
Thanks @lucasxsong! Would it be possible to make the "view rows by" button apply to both distribution charts (i.e., the score distribution by grade and the support level distribution by grade)?

Likewise, can the "view by" (percent | count) button that currently applies to the support level distribution plots also apply to the score distribution by grade/school plots?

richford (approved):

PR #303, opened 2024-01-23
Compact view on Dashboard

Emily-ejag:
Making Dashboard view more compact


richford (changes requested):
When I expand the results for an administration, it jumbles up the administration details section. Perhaps this won't happen if we merge #301 first and then rebase this. If that's the case, let's wait on #301. But otherwise, I think this will have to be adjusted.

Screenshot 2024-01-25 at 4 31 57 AM

Emily-ejag:
Yep that was a <div> that I had to delete, thank you for noticing

richford (approved):

PR #305, opened 2024-01-24
Add support for differing support levels for different grades

kellyel:
This PR adds infrastructure to allow changes to the support level criteria. In particular, this PR changes students grades 6 and above to be categorized based on raw score with the cutoffs of 
SWR: 400 - 550
SRE: 47 - 73


richford:
Task linked: [CU-8686y191g Change support categorization for grades >= 6](https://app.clickup.com/t/8686y191g)

src/helpers/reports.js
```diff
@@ -57,9 +57,41 @@ export const getSupportLevel = (percentile) => {
       support_level = 'Needs Extra Support';
       tag_color = supportLevelColors.below;
     }
+  } else if (rawScore !== undefined && grade >= 6) {
+    const { above, some } = getRawScoreThreshold(taskId);
+    if (rawScore >= above) {
+      support_level = 'At or Above Average';
+      tag_color = supportLevelColors.above;
+    } else if (rawScore > some && percentile < above) {
```
richford, line 65: I think that both of these variables should be `rawScore`, correct?
```suggestion
    } else if (rawScore > some && rawScore < above) {
```

src/helpers/reports.js
```diff
@@ -43,10 +43,10 @@ export const supportLevelColors = {
   below: '#c93d82',
 };
 
-export const getSupportLevel = (percentile) => {
+export const getSupportLevel = (grade, percentile, rawScore, taskId) => {
   let support_level = null;
   let tag_color = null;
-  if (percentile !== undefined) {
+  if (percentile !== undefined && grade < 6 && !['swr, pa'].includes(taskId)) {
```
richford, line 49: I don't understand why we check the `taskId` here as well. This would suggest that we don't want to give support categories to elementary aged students who have a percentile score for SWR or PA.

Whatever the answer to the above question is, shouldn't this line also be
```suggestion
  if (percentile !== undefined && grade < 6 && !['swr', 'pa'].includes(taskId)) {
```

richford (changes requested):

src/helpers/reports.js
richford: I think this one is still unresolved.

src/helpers/reports.js
richford: I think the explicit check on task IDs can be removed. My reasoning is that this is the same situation as elementary school students who take letter. We don't have norms for letter so we don't assign them a category. When they enter this function, they won't have `percentile` defined, so they will not get a support category. And I believe the logic is the same for middle and high schoolers. Does that make sense?

richford (changes requested):
Hi Elijah,

I think those two comments are still unresolved. I see that you merged and pushed but they didn't appear to change.

richford (approved):

PR #307, opened 2024-01-24
Update letter public facing name

lucasxsong:
Updating ROAR-Letter to ROAR-Letter Names and Sounds across administration card, subscore tables, score tables, and task reports

### DIFF
--- a/src/components/RoarDataTable.vue
+++ b/src/components/RoarDataTable.vue
@@ -347,7 +347,7 @@ dateFields = _map(dateFields, (col) => col.field);
 let toolTipByHeader = (header) => {
   if (header === 'Word') {
     return 'Assesses decoding skills at the word level. \n\n  Percentile ranges from 0-99 \n Raw Score ranges from 100-900';
-  } else if (header === 'Letter') {
+  } else if (header === 'Letter Names and Sounds') {
     return 'Assesses decoding skills at the word level. \n\n Percentile ranges from 0-99 \n Raw Score ranges from 0-90';
   } else if (header === 'Phoneme') {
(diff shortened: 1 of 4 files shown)


richford:
@lucasxsong, thanks for taking this on. You're right (in private communication) that the column width for letter is now unattractively long. Can you line break after "Names" (i.e., `Letter Names<br />And Sounds`)? Maybe we should do the same for other multi-word assessment names. We could also set a max-width for the assessment columns using the `style` attribute of the `PvColumn` component.

@kellyel, your thoughts?

kellyel:
image
I can confirm that the above changes work. Here, I've added `style="max-width: 11.25rem; text-wrap: wrap"` to the PvColumn in RoarDataTable. The columns are still resizable. I think this looks okay, though the rows getting wrapped makes them much taller. If we make these changes, maybe we want to center the color tags for scores as well?

Another option here is to apply wrapping and width changes to the template header, so that wrapping changes apply only to the headers. I think I like this more, but the solution needs a bit more tinkering. Applying wrap to the header div works nicely but then the 'Letter Names and Sounds' text compresses itself to 4 lines so we would need to apply some sort of width to that as well. Happy to chat about this sometime today or tomorrow!

lucasxsong:
image
I've added some CSS logic that appends the width attribute along with the text-wrap attribute for only the Letter names and sounds header.

richford:
@lucasxsong, thanks. Can we set `max-width` instead of `width` and use relative units like `em` instead of `px`.

richford:
@lucasxsong, that's great! Thanks! Can you rebase on main and then this is good to go!

kellyel (approved):
Looks excellent!

PR #310, opened 2024-01-25
Remove the need to toggle norms when filtering

kellyel:
This PR allows the query function for filtering scores to access both elementary and secondary results with one query.


src/helpers/query/assignments.js
```diff
@@ -317,16 +317,85 @@ export const getFilteredScoresRequestBody = ({
...
+                      {
+                        fieldFilter: {
+                          field: { fieldPath: 'userData.schoolLevel' },
+                          op: 'EQUAL',
+                          value: { stringValue: 'high' },
+                        },
+                      },
+                    ],
+                  },
+                },
```
richford, line 364: Likewise, here we should include `postsecondary` in the composite filter.

richford (changes requested):

richford (approved):

PR #311, opened 2024-01-26
Filter on user fields

kellyel:
This PR allows users to filter on fields other than scores in the score report, as well as sorting.


src/helpers/query/assignments.js
```diff
@@ -81,20 +85,33 @@ export const getAssignmentsRequestBody = ({
...
+      // const userFilters = (filters.map(filter => {
+      //   console.log('filter ->', filter)
+      //   return {
+      //     fieldFilter: {
+      //       field: { fieldPath: filter.field },
+      //       op: "EQUAL", // Need more logic here later for different constraints
+      //       value: { stringValue: filter.value }
+      //     }
+      //   }
+      // }))
```
richford, line 111:
```suggestion
```

src/helpers/query/assignments.js
```diff
@@ -609,17 +647,42 @@ export const assignmentPageFetcher = async (
...
+    //     aggregationQuery: false,
+    //     pageLimit: pageLimit.value,
+    //     page: page.value,
+    //     paginate: paginate,
+    //     filter: filters,
+    //   });
+    //   console.log('Request Body', requestBody);
+    //   return adminAxiosInstance.post(':runQuery', requestBody).then(async ({ data }) => {
+    //     console.log('Found Data:', data);
+    //   });
```
richford, line 665:
```suggestion
```

richford (changes requested):
This looks great @kellyel! Thanks for all of your hard work. I have two minor requests and one major one.

- Sorting by grade results in no data table to be displayed.
- I made two suggestions for removing commented out code.

kellyel:
Feedback addressed:
- Change wording of filters in dropdown
- Allow filters to be applied to scores from any view (color, percentile, standard, raw)
- Prevent user from filtering on blacklisted columns (letter)
- Visually remove option to filter on blacklisted columns. I am as of right now unable to remove the button itself, however I have disabled sorting on that column (no action happens), and I was able to remove the icon for those columns. the only indication that there is a button there is the hover event. I am not 100% happy with this, and will return to the problem later.

richford (approved):

PR #317, opened 2024-01-26
Make grade range options more clear when filtering

kellyel:
This PR replaces the checkbox for national norms with radio buttons encouraging the user to choose between the two grade ranges. The default is K-6.


lucasxsong (changes requested):
This looks great and the additions certainly help with making it easier to understand the filter for scores! Just have a small nit: could we add a bit of margin-left to the filter icon so the outline does not clip with the header? 

image

Emily-ejag:
I think after the margin it looks good!

PR #319, opened 2024-01-28
Progress bar same length calculation

Emily-ejag:
Making calculation to get the max value of the bar for all the progress bars


richford:
Thanks @Emily-ejag, one drawback of this approach is that it changes the numbers we see in the tooltip when you hover of the data. However, your PR made me realize we could fix it by explicitly setting the min/max in the chart options, which is what I did in #325. So I'm closing in favor of #325. But thanks for diagnosing this error.

Emily-ejag:
Good!

PR #320, opened 2024-01-29
Compress of Expand Tables in Dashboard

Emily-ejag:
Button to compress/expand the rows on a table for preference

### DIFF
--- a/src/assets/styles/components/datatable.scss
+++ b/src/assets/styles/components/datatable.scss
@@ -84,12 +84,7 @@
   color: var(--text-color);
   transition: none;
 }
-.p-datatable .p-datatable-tbody > tr > td {
-  text-align: left;
-  border: 1px solid var(--surface-c);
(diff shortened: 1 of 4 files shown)


richford (changes requested):
Looking good! I noticed that the compress/expand button is missing from the list-users table (the one accessed by clicking on the "users" button in the last column of the list-orgs table.

I think it'd be nice if all of the tables had the same compress/expand button. Can you add it there. Then I think we're good to go.

Emily-ejag:
The issue was that exporting was not allowed in the list-users table, so I moved the v-if out of the` span` and added the buttons

richford:
Hi Emily, I'm still not seeing the expand/compress button in the list users table on the PR preview link.

Emily-ejag:
The button is there. It was showing in my localhost. But after the rebase is in the PR link now

richford (approved):

PR #337, opened 2024-01-31
Filter and Sort changes

kellyel:
This PR addresses changes requested to the filter / sorting experience on the score reports.


Emily-ejag:
The table on the score report is working perfectly!
Do you know if the table below each task also needs filtering?
image

Also, on tables from list-orgs I see the filter button, but I can't filter
image

Same case with list-users
image

kellyel:
Ah yes, tables other than the score report will not require filtering at this time. I will update that, one moment.

Emily-ejag (approved):

PR #340, opened 2024-02-01
Moving support level dropdown view

Emily-ejag:
Moving dropdown closer to the table and to the right as requested


kellyel (approved):
Looks good!

PR #353, opened 2024-02-03
WIP: Update outdated packages

richford:


kellyel (approved):
⭐

Emily-ejag:
I noticed some of the styling going off on the checkmarks: 
image

Other example on list Orgs: 
image

Score Reports: 
image

I noticed it is double once you check it another checkbox appears

Emily-ejag (changes requested):
Checkbox styling off, and repeating checkbox

richford:
I think the issue here is that I updated primevue but we have legacy scss files in the `src/assets/styles/components` directory. These were copied over from a previous version of PrimeVue and I'm guessing that they need to be updated as well. Two options:

- Figure out the updates to the scss files and merge those with the adjustments that we previously made to these files.
- Wait until "the end of Q1 2024" when
> [PrimeVue v4 will be introduced with a modern theming architecture based on design tokens and css variables.](https://primevue.org/theming/#architecture)

which will make custom theming and colors much easier.

richford:
I don't have the combination of bandwidth/expertise to do option 1. @Emily-ejag, if you feel like you can do option 1 without diverting too much of your time, then go for it. It would be nice to merge this without waiting on the theming updates from PrimeVue. But not a big deal if that doesn't work out.

PR #359, opened 2024-02-07
Adding Parent SignUp

Emily-ejag:
Adding Register Parent, Register student, their styles and routes


src/components/auth/RegisterStudent.vue
```diff
@@ -0,0 +1,665 @@
+<!-- eslint-disable vue/valid-v-for -->
```
richford, line 1: I'm not sure which `v-for` is causing this error, but in general, I think we should try not to ignore these errors and track down which one is causing the error.

richford (changes requested):

richford:
@Emily-ejag and @kellyel, this is looking really good! In addition to my comments above, I think we need to figure out what happens when no org code is provided. Since the first incremental delivery of the parent signup is without org code, I think we should
- Create a new group called `ROAR families` and add every new family that doesn't have an org code to this group.
- Let's add a field called `families` in the Firestore doc for this group so that we can keep track of which families are  in it.
- Create a default administration for all families without an org code. Assign the ROAR families group to this administration. And when a new family is created, assign that family to the administration as well.
- Disable the org code input box for now until it is actually supported.

src/components/auth/RegisterParent.vue
```diff
@@ -0,0 +1,242 @@
...
+  accept: false,
+});
+const passwordRef = computed(() => state.password);
+
+
+
+const ParentEmail = (value) => {
+  if (!value.includes('@')) return true;
+
+  const emailRegex = /^(?!.*@.*@)[a-zA-Z0-9]+([._]?[a-zA-Z0-9]+)*(@[a-zA-Z0-9]+([.-]?[a-zA-Z0-9]+)*(\.[a-zA-Z]{2,7})+)?$/;
```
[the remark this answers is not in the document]
Emily-ejag: I think we don't even need this, we just have to add on the input the type email and the form will do all the validation for us since we are using html5.

src/store/query.js
```diff
@@ -20,6 +20,13 @@ export const useQueryStore = () => {
       };
     },
     actions: {
+      async getOrgs(orgType) {
+        if (roarfirekit.value?.app?.db) {
+          return roarfirekit.value.getOrgs(orgType);
+        } else {
+          return []
+        }
+      },
```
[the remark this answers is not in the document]
Emily-ejag: yeah, you are right, we don't need it

src/pages/Register.vue
```diff
@@ -0,0 +1,160 @@
+<!-- eslint-disable vue/multi-word-component-names -->
+// eslint-disable-next-line vue/multi-word-component-names
```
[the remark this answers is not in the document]
Emily-ejag: We have a RegisterUser already, I don't think we are using it, do you want me to delete it from auth and from pages, or you want me to change the name of Register to like RegisterParent

src/components/auth/RegisterStudent.vue
Emily-ejag: Yeah, you are right, the indexes were mixing up. I fixed it

richford:
@kellyel and @Emily-ejag, I merged the corresponding firekit PR. You probably want to update firekit in this PR to v4.8.0.

richford (changes requested):
Looking good overall. I left an inline comment about the API key for Recaptcha. In addition to that, I have some requested enhancements:
- [x] Create a confirmation page that the family registration routes to after successful completion
- [x] Create an error page that the family registration routes to if there's an error in registering
- [x] Add tabs to sign-in page to switch between signing in and registering as a parent. I assume this would mean renaming the pages/SignIn.vue component to something like `LandingPage.vue` or `SignUpOrIn.vue` or something like that. And then it could call different components in a `PvTabView`.
- [ ] Disable the org code for now. I.e., put it in a `v-if="false"` block. The org code doesn't work right now so we should just not show it.
- [x] Check username availability on the client side before submitting the form. This will stop users from attempting to register emails or usernames that already exist. [See this function](https://github.com/yeatmanlab/roar-dashboard/blob/6e8006a4b8ce5d8f0a99e57d655dd3642a254ac4/src/components/auth/SignIn.vue#L117) for an example of how we do that in the SignIn.vue component. It will have to be modified slightly but you can use something like that to check for username/email availability in form validation.

richford (changes requested):
We had an in-person discussion about this (@kellyel, @Emily-ejag, and me) and decided on the following enhancements
- [x] Use [Primevue's Steps component](https://primevue.org/steps/) to enable going back and forth between the parent and child registration pages
- [x] Add "Register family" as a menu item for super-admins.
- [x] If the user is authenticated and a super-admin, add a checkbox to indicate whether or not this family should have the `testData` boolean flag
- [x] Consider switching to Primevue's dialog component rather than writing your own.

src/components/auth/RegisterParent.vue
```diff
@@ -0,0 +1,312 @@
...
+const closeErrorDialog = () => {
+  isDialogVisible.value = false;
+};
+
+const props = defineProps({
+  isRegistering: { type: Boolean, default: true },
+});
+
+console.log(props);
+
```
richford, line 184:
```suggestion
```

src/pages/RegisterFamilyUsers.vue
```diff
@@ -0,0 +1,242 @@
...
+  } else {
+    return RegisterStudent;
+  }
+}
+
+watch([parentInfo, studentInfo], ([newParentInfo, newStudentInfo]) => {
+  if (newParentInfo && newStudentInfo) {
+    const rawParentInfo = toRaw(newParentInfo);
+    const rawStudentInfo = toRaw(newStudentInfo);
+    console.log('both student and parent info present');
```
richford, line 156:
```suggestion
```

src/pages/RegisterFamilyUsers.vue
```diff
@@ -0,0 +1,242 @@
...
+      };
+    });
+    authStore.createNewFamily(
+      rawParentInfo.ParentEmail,
+      rawParentInfo.password,
+      parentUserData,
+      studentSendObject,
+      isTestData.value,
+    );
+    console.log('firekit function called');
```
richford, line 193:
```suggestion
```

richford (changes requested):
Amazing! I suggested some changes. Mostly just removing console.log statements and also changing the name of the drop-down button.

Also, could you please
- [x] move the back button to be next to the submit button on the student registration page, ideally both on bottom, with the back button fully left-justified and the submit button fully right-justified.
- [ ] Use a [PrimeVue icon](https://primevue.org/icons/#list) for the left arrow in the back button. I believe we want `pi-arrow-left`.

Lastly, @kellyel, I tested it out one last time and all of the appropriate Firestore documents were created except for the sub-group for the default administration. I recall that this was working before but it isn't any longer. Can you look into the firebase cloud function?

richford (commented):
@Emily-ejag and @kellyel, I think this is good to go for internal review. I'm clicking "comment" rather than "approve" on this review since I'd like to do internal review using the PR preview link. After internal review, then we can approve and merge. Here are the steps I'd recommend to get started with internal review:

- [x] Create a default ROAR families administration
- [x] Test the workflow once more to confirm that you can
  - [x] Create a family
  - [x] Log in as a child and take the default administration
  - [x] Log in as a parent and view scores for your children
- [x] Delete all testing families, users, userClaims, and subgroups created during development.
- [x] Send the PR preview link to the `#roar-staging-testing` Slack channel. Remind everyone to log in as a super admin first so that they can check the test data checkbox. Create a checklist of things that you'd like internal reviewers to check.
- [x] Respond to review comments/suggestions
- [ ] Delete all test data created during internal review
- [ ] Reimplement Captcha (this is a blocker for deployment to production)
- [ ] Approve, merge, celebrate!!!

Thanks so much for all of your hard work on this!

richford (approved):

PR #366, opened 2024-02-09
WIP: exporting flags

Emily-ejag:
Exporting on the CSV files


src/pages/ScoreReport.vue
```diff
@@ -894,6 +901,12 @@ const exportAll = async () => {
         ? _get(assessment, 'scores.computed.composite')
         : _get(assessment, `scores.computed.composite.${rawScoreKey}`);
       tableRow[`${taskDisplayNames[taskId]?.name ?? taskId} - Support Level`] = support_level;
+      if (assessment.reliable !== undefined && !assessment.reliable && assessment.engagementFlags !== undefined) {
+        tableRow[`${taskDisplayNames[taskId]?.name ?? taskId} - Flags`] = _lowerCase(
```
lucasxsong, line 905: Would it be possible to get the reliability/engagement columns to be grouped with the scores? Right now the columns start with displaying all the score data first and then group all the engagement flags together at the end. I feel that it would be more clear if the engagement flags/reliability is presented with the scores. 

While we're at it, would it be possible to add a column the indicates reliability? (e.g.`${taskId} - Reliable`). The current implementation does not have much indication on whether or not a score is reliable, and I feel that that is important in context of adding the flags. Also, what do you think of changing the column name from `${taskId} - Flags` to `${taskId} - Engagement Flags` to better match the language in the table?

lucasxsong (changes requested):

src/pages/ScoreReport.vue
```diff
@@ -839,6 +840,12 @@ const exportSelected = (selectedRows) => {
...
+          Object.keys(assessment.engagementFlags).join(' ,'),
+        );
+        console.log('Flags: ', _lowerCase(Object.keys(assessment.engagementFlags)));
```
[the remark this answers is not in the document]
Emily-ejag: TRUE, thanks

src/pages/ScoreReport.vue
Emily-ejag: Good point, let me add reliable next to the reliable ones, and change the column header

src/pages/ScoreReport.vue
```diff
@@ -839,6 +840,17 @@ const exportSelected = (selectedRows) => {
         ? _get(assessment, 'scores.computed.composite')
         : _get(assessment, `scores.computed.composite.${rawScoreKey}`);
       tableRow[`${taskDisplayNames[taskId]?.name ?? taskId} - Support Level`] = support_level;
+      if (assessment.reliable !== undefined && !assessment.reliable && assessment.engagementFlags !== undefined) {
+        const reliabilityKeys = Object.keys(assessment.engagementFlags);
+        if (reliabilityKeys.length > 0) {
+          const reliabilityString = reliabilityKeys.map((key) => _lowerCase(key)).join(', ');
```
ksmontville, line 846:
```suggestion
          const reliabilityString = reliabilityKeys.map((key) => _capitalize(key)).join(', ');
```
Not a critical change, but I wonder if we should capitalize these using the lodash util to capitalize the first letter of a string? Or write a function to capitalize each word to match the rest of the document. The all-lowercase reliability flags look award against evereything else, in my opinion.

image

src/pages/ScoreReport.vue
```diff
@@ -894,6 +906,17 @@ const exportAll = async () => {
...
```
ksmontville, line 912:
```suggestion
          const reliabilityString = reliabilityKeys.map((key) => _capitalize(key)).join(', ');
```
Same here.

kellyel (approved):

src/pages/ScoreReport.vue
```diff
@@ -839,6 +840,17 @@ const exportSelected = (selectedRows) => {
         ? _get(assessment, 'scores.computed.composite')
         : _get(assessment, `scores.computed.composite.${rawScoreKey}`);
       tableRow[`${taskDisplayNames[taskId]?.name ?? taskId} - Support Level`] = support_level;
+      if (assessment.reliable !== undefined && !assessment.reliable && assessment.engagementFlags !== undefined) {
```
ksmontville, line 843:
```suggestion
      if (!assessment.reliable && assessment.engagementFlags !== undefined) {
```
I believe that the first check may be redundant for our purposes since we don't care if `assessment.reliable` is null or undefined, just that it's falsy, but correct me if I am wrong.

ksmontville (approved):

src/pages/ScoreReport.vue
Emily-ejag: I think I prefer to keep them both for now, because those are 2 co-dependent things

src/pages/ScoreReport.vue
Emily-ejag: Cool, let me try this

Emily-ejag: I think _lowerCase is better in this case, because otherwise it will export like this: ResponseTimeTooFast, AccuracyTooLow and not as separate words

ksmontville: I should actually have written this:

`const reliabilityString = reliabilityKeys.map((key) => _capitalize(_lowerCase(key))).join(', ');`

Which gives this result, so at least the first letter of each key is capitalized:

image

lucasxsong (approved):

PR #370, opened 2024-02-12
Restore letter tab

richford:


src/pages/ScoreReport.vue
```diff
@@ -953,8 +954,9 @@ const runsByTaskId = computed(() => {
       computedScores[run.taskId] = [run];
     }
   }
+  console.log(computedScores);
```
lucasxsong, line 957: I think we can remove this console.log?

lucasxsong (changes requested):
Just an errant console.log, but besides that it looks good.

lucasxsong (approved):

src/pages/ScoreReport.vue
richford: Good catch, thanks!

PR #371, opened 2024-02-12
Individual Score Reports

lucasxsong:


src/pages/IndividualReport.vue
```diff
@@ -0,0 +1,410 @@
...
+});
+
+// const {data: classData } = useQuery({
+//   queryKey: ['classes', props.userId],
+//   queryFn: () => fetchDocById('classes', studentData.value.classes.current[0]),
+//   enabled: studentData.value !== undefined,
+//   keepPreviousData: true,
+//   staleTime: 5 * 60 * 1000,
+// })
+
```
richford, line 152: Recommend deletion if no longer needed
```suggestion
```

src/pages/IndividualReport.vue
```diff
@@ -0,0 +1,410 @@
+<template>
+  <div v-if="!studentData" class="flex flex-column justify-content-center align-items-center loading-container">
+    <AppSpinner style="margin-bottom: 1rem" />
+    <span>Loading Your Individual Roar Score Report...</span>
+  </div>
+
+  <div
+    v-else
+    id="individual-report-export"
+    ref="IndividualReportDownload"
```
richford, line 10: Style nitpick: recommend camel case with lowercase first character
```suggestion
    ref="individualReportDownload"
```

src/pages/IndividualReport.vue
```diff
@@ -0,0 +1,410 @@
...
+
+const expanded = ref(false);
+const exportLoading = ref(false);
+
+const setExpand = () => {
+  console.log('expand called');
+  expanded.value = !expanded.value;
+};
+
+const IndividualReportDownload = ref(null);
```
richford, line 199: Style nitpick: recommend camel case with lowercase first character
```suggestion
const individualReportDownload = ref(null);
```

src/pages/IndividualReport.vue
```diff
@@ -0,0 +1,410 @@
...
+  sre: 'ROAR-Sentence',
+  vocab: 'ROAR-Picture Vocabulary ',
+  multichoice: 'ROAR-Multiple Choice',
+  morph: 'ROAR-Morphology',
+  cva: 'ROAR-Written Vocabulary',
+  letter: 'ROAR-Letter',
+  comp: 'ROAR-Comprehension',
+  syntax: 'ROAR-Syntax',
+  fluency: 'ROAM-Fluency',
+};
```
richford, line 286: Is this duplicating any of the titles in the helpers/reports.js file? If so, can you consolidate? If not, can you move this into there and import?

src/pages/IndividualReport.vue
```diff
@@ -0,0 +1,410 @@
...
+//
+//   // Use the filter method to find matching elements
+//   const matchingClasses = studentData.classes.current.filter(className =>
+//     administrationData?.classes.includes(className)
+//   );
+//   classId.value = matchingClasses;
+//   classMatched.value = true;
+//   return matchingClasses;
+// }
+
```
richford, line 318: Recommend deletion
```suggestion
```

src/pages/IndividualReport.vue
```diff
@@ -0,0 +1,410 @@
...
+function getGradeWithSuffix(grade) {
+  if (getGrade(grade) < 1) {
+    return grade;
+  } else if (getGrade(grade) === 1) {
+    return grade + 'st';
+  } else if (getGrade(grade) === 2) {
+    return grade + 'nd';
+  } else if (getGrade(grade) === 3) {
+    return grade + 'rd';
+  } else if (getGrade(grade) >= 4 && getGrade(grade) <= 12) {
```
richford, line 297: I think we technically allow up to grade 13 in the database.
```suggestion
  } else if (getGrade(grade) >= 4 && getGrade(grade) <= 13) {
```

src/pages/IndividualReport.vue
```diff
@@ -0,0 +1,410 @@
...
+        } else {
+          return -1;
+        }
+      })
+      .join(', ') + '.'
+  );
+});
+
+const studentFirstName = computed(() => {
+  if (studentData.value.name.first == undefined) return studentData.value.username;
```
richford, line 264: This might need conditional chaining
```suggestion
  if (studentData.value.name?.first == undefined) return studentData.value.username;
```

richford, line 264: Also I'd recommend adding a comment about why we're using `==` instead of `===`. The reason to add the comment is that linters often flag the use of `==` and months from now, someone else might come in and change it to `===`, not realizing that `==` is actually what we want. The comment would protect against that somewhat. I guess we could also add an eslint-ignore comment too.

src/components/reports/IndividualScoreReportTask.vue
```diff
@@ -0,0 +1,404 @@
...
+//   }
+//   if (taskId === 'sre') {
+//     if (getGrade(grade) < 6) {
+//       return 'tosrecSS';
+//     } else {
+//       return 'sprStandardScore';
+//     }
+//   }
+// };
+
```
richford, line 323:
```suggestion
```

src/components/reports/IndividualScoreReportTask.vue
```diff
@@ -0,0 +1,404 @@
...
+  });
+
+  // Ensure scores are in consistent order
+  return formattedScoresArray.sort((first, second) => {
+    return first[0].localeCompare(second[0]);
+  });
+};
+
+let supportColor = null;
+// let remainderColor = null
```
richford, line 261:
```suggestion
```

src/components/reports/IndividualScoreReportTask.vue
```diff
@@ -0,0 +1,404 @@
...
+    cva: 'This test measures your students’ knowledge of words that are often used in the books they read at school',
+    letter:
+      'This test measures how well your student knows the names of letters and which letters are used to spell each sound',
+    comp: 'Temporary description for comp',
+    phonics:
+      'This test measures phonics knowledge by testing how well your student can match the sounds of a word to the spelling',
+    syntax: 'This test measures how well students understand sentences that vary from simple to complicated',
+    fluency: 'Temporary description for fluency',
+  },
+};
```
richford, line 203: I'd love to move these names and descriptions into the helpers/reports.js file so that they are commonly available to all of the report components/pages.

src/components/reports/DistributionChart.vue
```diff
@@ -0,0 +1,138 @@
+<template>
```
richford, line 1: If this component part of a work in progress? I don't see it being imported/used anywhere.

richford (changes requested):
This looks really good, @lucasxsong!!! It's really nice to see. I requested some changes. Some are very minor, like linting and deleting obsolete commented out code. Then I have some questions about the following components/pages:
- components/reports/DistributionChart.vue: is this a work in progress related to @jasetran's enhancement request? I didn't see it being imported or used anywhere, but maybe I missed it.
- components/reports/IndividualScoreReportTask.vue and pages/IndividualReport.vue: How much of the report names, descriptions, colors, etc. can we (or should we) move into the helpers/reports.js. My preference is to have as much of that centralized as makes sense. Likewise, how much of the PDF conversion code can be put in a common file and then imported into both the score report and individual score report pages?

richford (approved):

lucasxsong:
On the road to ROAR at home...

PR #379, opened 2024-02-20
wip/Dashboard Internationalization

ksmontville:
This PR builds out the foundations for translating the ROAR Dashboard into multiple languages.


src/router/index.js
```diff
@@ -15,7 +15,13 @@ const routes = [
...
     name: 'Home',
     component: () => import('../pages/HomeSelector.vue'),
-    meta: { pageTitle: 'Dashboard' },
+    meta: {
+      pageTitle: {
+        'en-US': 'Dashboard',
+        en: 'Dashboard',
+        es: 'Tablero',
+      },
+    },
```
richford, line 24: This is the only file so far that does not use translations from the external files in `translations/*/*.json`. Could we change the router files to also use translations from that directory? I'm just thinking that it would be nice if all of the translation content was in that folder.

src/router/index.js
```diff
@@ -140,7 +146,13 @@ const routes = [
...
     name: 'SignIn',
     component: () => import('../pages/SignIn.vue'),
-    meta: { pageTitle: 'Sign In' },
+    meta: {
+      pageTitle: {
+        'en-US': 'Sign In',
+        en: 'Sign In',
+        es: 'Iniciar Sesión',
+      },
+    },
```
richford, line 155: See comment above about moving all translation content to the `translations` directory.

src/router/index.js
```diff
@@ -152,7 +164,13 @@ const routes = [
...
       return { name: 'SignIn' };
     },
-    meta: { pageTitle: 'Sign Out' },
+    meta: {
+      pageTitle: {
+        'en-US': 'Sign Out',
+        en: 'Sign Out',
+        es: 'Cerrar Sesión',
+      },
+    },
```
richford, line 173: See comment above about moving all translation content to the `translations` directory.

src/pages/HomeSelector.vue
```diff
@@ -130,6 +130,7 @@ watch(isLoading, async (newValue) => {
 const { idle } = useIdle(20 * 60 * 1000); // 10 min
 const confirm = useConfirm();
 const timeLeft = ref(60);
+const { t } = useI18n();
```
richford, line 133: This seems like the only component where you explicitly import `useI18n` and declare the `t` variable. Is this necessary since you used `app.use(i18n)` in the main.js file?

richford (changes requested):

src/router/index.js
ksmontville: I am looking for a better solution to hard coding this way.

However, I wrote it this way because the i81n instance is only defined with a <script setup> function, so we are unable to directly insert translations into the router file since it lacks a setup function.

src/pages/HomeSelector.vue
ksmontville: Similar to the other issue, the string lliteral syntax {{ $t('someTranslation.item') }} is only defined within the HTML template. In the script, we have to create a specific instance of i81n-vue and perform the translation that way.

I am of course open to a better solution.

src/router/index.js
richford: Could we do something like
```
import pageTitles from `../translations/pageTitles.json` assert { type: `json` };
```

and then reference the `pageTitles` object in the code.

richford: Or perhaps they should be located in the individual language sub-folders.

richford (approved):

richford:
@ksmontville, LGTM. I just approved. The title says "WIP" but whenever you feel like this is ready, I think we could merge it.

src/main.js
```diff
@@ -69,6 +69,7 @@ import VueGoogleMaps from 'vue-google-maps-community-fork';
 
 // Internal Roar components
 import RoarDataTable from '@/components/RoarDataTable.vue';
+import LanguageSelector from './components/LanguageSelector.vue';
```
richford, line 72:
```suggestion
import LanguageSelector from '@/components/LanguageSelector.vue';
```

src/router/index.js
```diff
@@ -1,6 +1,7 @@
 import { createRouter, createWebHistory } from 'vue-router';
 import { useAuthStore } from '@/store/auth';
 import _get from 'lodash/get';
+import { pageTitlesEN, pageTitlesUS, pageTitlesES, pageTitlesCO } from '../translations/exports';
```
richford, line 4:
```suggestion
import { pageTitlesEN, pageTitlesUS, pageTitlesES, pageTitlesCO } from '@/translations/exports';
```

src/components/LanguageSelector.vue
```diff
@@ -0,0 +1,62 @@
...
+        <country-flag :country="country.option.code" class="mr-2" size="small" />
+        <span>{{ country.option.name }}</span>
+      </div>
+    </template>
+  </PvDropdown>
+</template>
+
+<script setup>
+import { computed } from 'vue';
+import { languageOptions } from '../translations/i18n.js';
```
richford, line 31:
```suggestion
import { languageOptions } from '@/translations/i18n.js';
```

src/main.js
```diff
@@ -1,8 +1,8 @@
 import { createApp } from 'vue';
 import { createPinia } from 'pinia';
 import { createHead } from '@vueuse/head';
-import router from '@/router/index.js';
-import App from '@/App.vue';
+import router from './router/index.js';
+import App from './App.vue';
```
richford, line 5: Why change from `@` to `.` here? I kind of prefer the `@` to the relative `.`.

src/main.js
```diff
@@ -78,6 +79,11 @@ import 'primeflex/primeflex.scss'; // primeflex
 import './assets/styles/theme-tailwind.css'; // base theme (pulled from Primevue)
 import './assets/styles/theme.scss'; // ROAR theme
 
+// translations
+import { i18n } from './translations/i18n.js';
```
richford, line 83:
```suggestion
import { i18n } from '@/translations/i18n.js';
```

richford (approved):
Very small nitpicky changes requested, all having to do with the use of `@` in import statements. They are so small that I'm approving anyway as I'd be fine merging without them.

richford:
One thing for us to discuss is the translation of assent forms. Let's discuss at the all-hands meeting today. That might be out of scope for this PR.

PR #387, opened 2024-02-22
Register users page with upload / download of CSV

Zio-4:


src/router/sidebarActions.js
```diff
@@ -1,61 +1,93 @@
...
-    }
-    return true;
-  });
-  const actionsWithHomeLink = actions.filter((action) => {
-    if (!includeHomeLink && action.buttonLink.name === 'Home') {
-      return false;
-    }
-    return true;
-  });
+export const getSidebarActions = (isSuperAdmin = false, isAdmin = false, includeHomeLink = true) => {
```
kellyel, line 67: We discussed reformatting the arguments of this function to be a single object with keys `isSuperAdmin`, `isAdmin`, and `includeHomeLink`

kellyel: We also discussed implementing the filter for project in the sidebar actions, I think you already started that though

src/router/index.js
```diff
@@ -136,6 +136,12 @@ const routes = [
     component: () => import('../pages/RegisterStudents.vue'),
     meta: { pageTitle: 'Register Students', requireAdmin: true, requireSuperAdmin: true },
   },
+  {
+    path: '/register-users',
+    name: 'Register Users',
+    component: () => import('../pages/LEVANTE/RegisterUsers.vue'),
```
kellyel, line 142: We discussed implementing project-specific route guardsbased on the env variable, so that the projects would not have access to outside routes easily.

src/pages/LEVANTE/RegisterUsers.vue
```diff
@@ -0,0 +1,493 @@
...
+    });
+  });
+  return columns;
+}
+
+function getKeyByValue(object, value) {
+  return Object.keys(object).find((key) => object[key] === value);
+}
+
+function areArraysEqual(arr1, arr2) {
```
kellyel, line 229: For simplicity, this function could be substituted with [lodash's isEqual](https://lodash.com/docs/4.17.15#isEqual)

kellyel (approved):
Looks good!

PR #388, opened 2024-02-23
Cypress Network Tests

ksmontville:
This PR adds support for testing the dashboard and the apps under various simulated network conditions, such as 3G mobile connections or high latency networks.


kellyel (approved):

richford (commented):
Thanks @ksmontville. And thanks for adding the workflow to run on release. I'm wondering if we should instead make this a blocking job in https://github.com/yeatmanlab/roar-dashboard/blob/main/.github/workflows/firebase-hosting-production.yml so that if these tests fail, then it would block the other jobs and halt the deployment.

I can see benefits to both. But maybe we don't want to block deployment if a test fails on 2G for example. This isn't a request change but just a comment. What do you think. Should we run the network tests for informational purposes on each release? Or should we make passing the network tests a condition of release?

richford (approved):

PR #399, opened 2024-03-01
Remove multi-sort indexes

kellyel:
This PR removes indexes we added to enable server-side multi-sort for score reports. We've decided to go a different direction, so we no longer need these indexes clogging up our admin project. 

There were a few indexes we wanted to keep from [this commit](https://github.com/yeatmanlab/roar-dashboard/commit/ffeba8de58e49d0a707aaf59c8f2eff91ef16aa4). We want to keep:
- The "default sort" for each org type, which is an index with `grade` and `userData.name.last`
    - Both are ASCENDING
    - Districts will also have schools ASCENDING, before grade and last name.
- Single column sorts for each org type
    - Both ASCENDING and DESCENDING variants of each [`userData.name.first`, `userData.name.last`, `userData.username`, `userData.grade`, and `userData.assessmentPid`]
    - These will still be needed to operate the single column sorts for each org type


richford (commented):
Correct me if I'm wrong here but we need at least one index for each org type that includes all of the possible columns to enable multi-filter, right? Order doesn't matter for filtering like it does for sorting so we don't have the combinatoric explosion that we had before. But I think we need just one index for multi-filter. Is that right?

kellyel:
We support multi-filtering for only a select few fields at the moment. In the table, we can filter by a single column. This is already included in the above indexes. However we also support filtering for grade alongside any field, so that requires an additional index for each field (for example, when filtering on the `last name` column, we also need to support `last name` && `grade`). The only organization that differs is district, which requires us to support filtering `grade` and `school` at any time.  So for that organization's indexes, I also added each column + `readOrgs.schools`, as well as column + `readOrgs.schools` + `grade`. All of these indexes have been added in my recent commit.

richford (approved):

PR #403, opened 2024-03-04
Fixing issue 197

Emily-ejag:
Unclear note showing up in password box [https://github.com/yeatmanlab/roar/issues/197](https://github.com/yeatmanlab/roar/issues/197)


richford (changes requested):
Can we add that language to the translation JSON file (and maybe also in Spanish)?

PR #409, opened 2024-03-07
Components for Conditional Assignment

lucasxsong:
C.f.C.A


src/components/CreateAdministration.vue
Emily-ejag, line 1: Hi Lucas, I think the component you created needs to be under Components and not tasks, since those are only for the games :)

richford (approved):

PR #412, opened 2024-03-08
Allow Email login

kellyel:
This PR adds the ability to log in using a password for email users.

### DIFF
--- a/.github/workflows/deploy-firebase-test-cypress.yml
+++ b/.github/workflows/deploy-firebase-test-cypress.yml
@@ -50,6 +50,8 @@ jobs:
       PARTNER_ADMIN_ID: ${{ secrets.PARTNER_ADMIN_ID }}
       PARTICIPANT_USERNAME: ${{ secrets.PARTICIPANT_USERNAME }}
       PARTICIPANT_PASSWORD: ${{ secrets.PARTICIPANT_PASSWORD }}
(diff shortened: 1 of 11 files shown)


richford (changes requested):
Code review and testing of the PR preview link look good! Could I ask you to add or modify copies tests as well? I realize that the cypress test user will not be able to log in with email link. But maybe we could just write a test that toggles the password box and verifies that it is enabled/disabled.

lucasxsong (commented):
This looks really good! I was testing on staging and noticed that this error prevented me from authing with email, but pulling down the changes and running the login flow on localhost worked. 

image

I had a small design suggestion: would it be possible to style the disabled password input box to look more like the disabled example in the Primevue docs? The current disabled input box works well in dissuading users from trying to enter a password, but I feel that the visual hinting provided by the darkened input background might help users in quickly identifying what mode they are in (email/pw or email/link). What do you think? 

image

kellyel:
Thanks for the review, Lucas.
Yes, I noticed that too. However that seems to be the case on all staging links right now. It seems firebase does not recognize the staging links as a valid uri. I just don't think that's been set up yet. It might be possible given that we use the firebase hosting github action to generate these. 

Yes, I agree with you. I can darken that background right now.

richford (approved):

PR #426, opened 2024-03-14
wip/Optional Assessments on Individual Reports

ksmontville:
This PR builds out the functionality for displaying optional assessments on individual reports, including exporting reports to PDF.


ksmontville:
image

src/pages/IndividualReport.vue
```diff
@@ -235,6 +235,31 @@ const exportToPdf = async () => {
     (exportLoading.value = false);
 };
 
+const optionalAssessments = computed(() =>
+  administrationData.value.assessments.filter((assessment) => assessment.optional),
+);
```
richford, line 240: I believe we want to grab the optional status from the assignment doc rather than the administration doc because the administration defined conditions that might make the assessment optional. Those conditions are then evaluated in the cloud function that syncs up assignments and the result of that evaluation is written as a boolean to the assignment doc.

So we will likely have to add another `useQuery` to grab the assignment at `/users/${props.userId}/assignments/${props.administrationId}`. Then this `optionalAssessments` can be computed from `assignmentData` rather than `administrationData`.

src/pages/IndividualReport.vue
```diff
@@ -235,6 +235,31 @@ const exportToPdf = async () => {
...
+};
+
+// Watch for changes in taskData and update the taskData with the new optional tasks
+watch(
+  () => taskData.value,
+  () => {
+    updateTaskData();
+  },
+  { deep: true },
+);
```
richford, line 261: Is there an advantage to using `queryClient.setQueryData` over another `computed` variable?

richford (changes requested):
Looks good. Thanks @ksmontville. I left some comments/requests below. Also, while viewing the individual score report, I noticed that we are running two nearly identical queries, one for `taskData` and one for `rawTaskData`. I'd like to dig into this and see if we can get rid of one. I'm happy to cowork to figure that out. Or maybe @lucasxsong can chime in to tell us the original motivation for that.

lucasxsong:
The two queries for run information should be refactored to just one, as the data returned from `rawTaskData` encompasses `taskData`. The queries were there originally when I had adopted the code from Elijah and Kyle, but I had put off refactoring those queries to work on more pertinent tasks.

src/pages/IndividualReport.vue
ksmontville: I looked into this a bit, and I found that the tanstackquery has its own internal reactivity and state management considerations when attempting to directly manipulate a returned query. The computed properties that I tried to create would not effectively update the taskData, and so I had to restort to tanstackquery's interal methods.

richford (approved):

PR #438, opened 2024-03-19
Enh/variant card handles

kellyel:
This PR implements a button alternative to the click and drag functionality of the task picker. This is for accessibility reasons and for cypress test reasons. 

Comments welcome on the UX for the button. I chose to implement this as a right-hand-side button rather than left-hand-side. the left would be more consistent with location for 'card actions', however I couldn't come up with a design that worked for the existing style and only have 1 button rather than 3.


Emily-ejag (approved):
looks good!

richford (approved):

PR #448, opened 2024-03-24
Set up sandbox projects

richford:
This PR adds firebase configurations for the development Firebase projects `gse-roar-admin-dev` and `gse-roar-assessment-dev`. These projects are intended to have the same functionality as the production projects but with only the ROAR demo district and the Clever ROAR demo district data.

It also adds variants to the `npm run dev` and `npm run build` scripts that define which data source the dashboard will use. Scripts ending in `:live` will use production data, while scripts ending in `:sandbox` will use development data:

- `npm run dev` starts a development server using sandbox data.
- `npm run dev:live` starts a development server using live data.
- `npm run build` builds the project using live data.
- `npm run build:sandbox` builds the project using sandbox data.


Emily-ejag (approved):

lucasxsong (approved):

kellyel (approved):
Looks good!

PR #452, opened 2024-03-25
Individual Score Report Bug Fixes

lucasxsong:
- Fix bug where Phoneme score wasn't displaying for kindergarteners
- Fix bug where support colors were incorrect for grades >= 6
- Update Score Text

### DIFF
--- a/src/helpers/reports.js
+++ b/src/helpers/reports.js
@@ -203,7 +203,7 @@ export const getSupportLevel = (grade, percentile, rawScore, taskId) => {
       tag_color: 'white',
     };
   }
-  if (percentile !== undefined && grade < 6) {
+  if (percentile !== undefined && (grade < 6 || grade === 'K')) {
     if (percentile >= 50) {
       support_level = 'Achieved Skill';
       tag_color = supportLevelColors.above;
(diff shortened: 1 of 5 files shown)


src/helpers/reports.js
```diff
@@ -203,7 +203,7 @@ export const getSupportLevel = (grade, percentile, rawScore, taskId) => {
       tag_color: 'white',
     };
   }
-  if (percentile !== undefined && grade < 6) {
+  if (percentile !== undefined && (grade < 6 || grade === 'K')) {
```
richford, line 206: Can we use the [`getGrade` function](https://github.com/yeatmanlab/roar-utils/blob/94cad5ebfde63ecaabe21bc1fc4695c3274dbeca/src/utils.js#L232) from roar-utils to handle lots of different kinds of grade input?

richford (changes requested):
Thanks @lucasxsong! These changes look great. My only suggestion is that we use the `getGrade` utility function from roar-utils throughout in order to parse more types of grades.

Even in `IndividualScoreReportTask.vue` can we use `getGrade` in the computed `grade` variable?

richford (approved):

PR #455, opened 2024-03-26
Optional Variant URL in Firestore

ksmontville:
This PR adds a checkbox control to appending the variant url to the variant parameters when registering a new game variant through the dashboard.

### DIFF
--- a/src/components/tasks/RegisterTask.vue
+++ b/src/components/tasks/RegisterTask.vue
@@ -222,7 +222,7 @@
 </template>
 
 <script setup>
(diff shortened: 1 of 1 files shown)


richford (changes requested):
Looking good! Can you also only write the `taskURL` field (in `handleNewTaskSubmit`) if `isExternalTask.value` is true?

ksmontville:
> Looking good! Can you also only write the `taskURL` field (in `handleNewTaskSubmit`) if `isExternalTask.value` is true?

Done, and made some additional refactors as well.

richford (approved):

PR #457, opened 2024-03-26
Implement spanish consent modal

kellyel:


src/pages/HomeSelector.vue
```diff
@@ -82,7 +82,7 @@ const isAdmin = computed(() => {
   return true;
 });
 
-const consentType = computed(() => (isAdmin.value ? 'tos' : 'assent'));
+const consentType = computed(() => (isAdmin.value ? 'tos' : i18n.locale.value === 'es' ? 'assent-es' : 'assent'));
```
richford, line 85: Requested change:
- to avoid a nested ternary
- to allow Spanish localizations (otherwise `es-co` results in the English assent form)
```suggestion
const consentType = computed(() => {
  if (isAdmin.value) {
    return 'tos';
  } else {
    return i18n.locale.value.includes('es') ? 'assent-es' : 'assent'
  }
});
```

richford (changes requested):

richford (approved):

Emily-ejag (approved):

PR #468, opened 2024-03-29
Update user info to navbar 

lucasxsong:


richford (changes requested):
Thanks @lucasxsong! I like where you're going with this. A few issues that I noticed:

- The layout doesn't update for me when I change the window width, only if I change the window width and then refresh.
- When the width is narrow, I wind up with a sign out button and a menu button that are different sizes.
- If I log in with my super-admin account, sign out, and then log in as a demo student. My super admin email still shows for a second or two before updating. I suspect the user info needs to be cleared in the auth store's `signOut` method.
- When I log in as a demo user, I don't get any user info, just the fall back "User" string. But I think we should at least have the demo user's email. It may just not be propagating to the auth store correctly.
- Now that we're bike shedding, I wonder if we should have different menu bar setups for participants and admins:
  - For participants, maybe the display name text should say something like "Hi <Display Name>" and the order of preference for display should be `name.first || username || email || "student"`. `name.first` isn't in the auth store but we have access to it in Firestore so we can (re)use a tanstack query. For participants, I think it's appropriate to keep the sign out button at the top level.
  - Then for admins, the display order could go `email || username || "administrator"`. Put the sign out button in a dropdown menu opened by clicking on the profile info.
  - For all emails, if the email ends in `@roar-auth.com`, remove that from the display
- I like the expansion of the menu into separate dropdowns for different actions. But let's workshop the top-level actions. "Create" and "register" both seem very similar to me. I don't have an immediate suggestion, but let solicit bike-shedding from the rest of the dev team.

lucasxsong:
Thanks for the review and the detailed feedback @richford. I've updated the PR with changes that address most of the suggestions that you brought up. 
- The navbar now dynamically changes with width change instead of requiring a refresh.
- The ROAR Logo width has been made dynamic to prevent the size difference between the buttons
- @roar-auth.com will be removed from any emails
- The ghost storage of the old authStore data in a new session has been fixed by clearing userData after signOut.
- Slightly tuned the conditionals for displayName (add name.first for participants) and added a greeting for participants. 
- Update the top level links to organizations, users, and register task. 

Let me know what you think of the changes now.

richford (changes requested):
Lucas, these changes are great! Some more comments/change requests/bikeshedding:

- For the menu options, I'm not sure it makes sense to put "Create administration" under "Organizations". Can we try the following:
```
- Administrations
  ├── Create administration
  └── Register task
- Organizations
  ├── List organizations
  └── Create organization
- Users
  ├── Register students
  └── Register administrator
```
- Can we use Primevue's `MenuBar` component (https://primevue.org/menubar/) to organize the different menus? It seems like it has a lot of nice responsiveness built-in.

PR #474, opened 2024-04-01
LEVANTE admin abilities: creating orgs, administrations, registering …

Zio-4:
…users

### DIFF
--- a/src/components/CreateAdministration.vue
+++ b/src/components/CreateAdministration.vue
@@ -261,6 +261,10 @@ const submit = async () => {
           toast.add({ severity: 'success', summary: 'Success', detail: 'Administration created', life: 3000 });
           administrationQueryKeyIndex.value += 1;
 
+          // TODO: Invalidate for administrations query.
+          // This does not work in prod for some reason.
+          // queryClient.invalidateQueries({ queryKey: ['administrations'] })
+
           router.push({ name: 'Home' });
         });
       } else {
--- a/src/components/NavBar.vue
+++ b/src/components/NavBar.vue
@@ -13,7 +13,7 @@
               </router-link>
             </template>
             <template #menubuttonicon>
-              <PvButton icon="pi pi-bars" @click="toggleMenu" label="Menu" />
+              <PvButton icon="pi pi-bars" label="Menu" @click="toggleMenu" />
             </template>
             <template #end>
               <div class="flex gap-2 align-items-center justify-content-center mr-3">
@@ -49,9 +49,9 @@ import { ref, computed, onMounted, onUnmounted } from 'vue';
 import { useRouter } from 'vue-router';
 import { storeToRefs } from 'pinia';
 import { useAuthStore } from '@/store/auth';
-import _get from 'lodash/get';
 import _isEmpty from 'lodash/isEmpty';
 import _union from 'lodash/union';
+import _get from 'lodash/get';
 import { getSidebarActions } from '@/router/sidebarActions';
 import { fetchDocById } from '@/helpers/query/utils';
 import { useQuery } from '@tanstack/vue-query';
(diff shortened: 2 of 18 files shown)


src/components/CreateAdministration.vue
```diff
@@ -259,7 +260,12 @@ const submit = async () => {
 
         await roarfirekit.value.createAdministration(args).then(() => {
           toast.add({ severity: 'success', summary: 'Success', detail: 'Administration created', life: 3000 });
+          // Used to rerender component. However, this does not invalidate the query. Deprecate?
           administrationQueryKeyIndex.value += 1;
+          // invalidate the query
+          // Can be further specified if many other queries use this key
+          queryClient.invalidateQueries({ queryKey: ['administrations'] });
+          // test comment
```
richford, line 268: Further investigation needed on prod???

src/App.vue
```diff
@@ -13,7 +13,7 @@
   </AppHead>
   <div>
     <PvToast />
-    <NavBar v-if="!navbarBlacklist.includes($route.name)" />
+    <NavBar v-if="!navbarBlacklist.includes($route.name) && isFirekitInitialized" />
```
richford, line 16: Great! Love it! Can you remove the subscribe patterns in the rest of the components?

Or maybe a better approach is to pull this into a separate PR so that it doesn't hold up your LEVANTE-specific enhancements.

src/components/NavBar.vue
```diff
@@ -71,12 +71,6 @@ const { data: userClaims } = useQuery({
   staleTime: 5 * 60 * 1000, // 5 minutes
 });
 
-const isAdmin = computed(() => {
-  if (userClaims.value?.claims?.super_admin) return true;
-  if (_isEmpty(_union(...Object.values(userClaims.value?.claims?.minimalAdminOrgs ?? {})))) return false;
-  return true;
-});
-
```
richford, line 79: Can you test this on both admin view and participant view, and Google SSO login vs username/password?

src/router/index.js
```diff
@@ -372,16 +374,35 @@ router.beforeEach(async (to) => {
...
+      // Currently no difference between admin and super admin
+      next();
+    } else {
+      // Block routing if user lacks permissions
+      next({ name: 'Home' });
+    }
+  } else {
+    // Allow all users to access the route
+    next();
   }
```
richford, line 406:
```suggestion
  const isLevante = import.meta.env.MODE !== 'LEVANTE'
  if (isUserSuperAdmin) {
    next();
    return;
  } else if (isUserAdmin) {
    // Levante dashboard has opened some pages to administrators before the ROAR dashboard
    // So if isLevante, then allow regular admins to access any route with requireAdmin = true.
    // and if ROAR, then prohibit regular admins from accessing any route with requireSuperAdmin = true.
    if (isLevante && requiresAdmin) {
      next();
      return;
    } else if (requiresSuperAdmin) {
      next({ name: 'Home'})
      return;
    }
    next();
    return;
  }
  
  // If we get here, the user is a regular user
  if (requiresSuperAdmin || requiresAdmin) {
    next({ name: 'Home' });
    return;
  }
  
  next();
  return;
```

src/router/sidebarActions.js
```diff
@@ -68,10 +68,15 @@ export const getSidebarActions = ({ isSuperAdmin = false, isAdmin = false, inclu
...
     return sidebarActionOptions.filter((action) => {
       if (action.project === 'LEVANTE' || action.project === 'ALL') {
-        if ((action.requiresSuperAdmin && !isSuperAdmin) || (action.requiresAdmin && !isAdmin)) {
+        if (
+          (action.requiresAdmin && isAdmin) ||
+          (action.requiresSuperAdmin && isSuperAdmin) ||
+          (!action.requiresAdmin && !action.requiresSuperAdmin)
+        ) {
+          return true;
+        } else {
```
richford, line 77: Add comment, please

richford (changes requested):

richford (approved):

PR #480, opened 2024-04-04
Add updateVideoCompletion components

lucasxsong:

### DIFF
--- a/src/components/GameTabs.vue
+++ b/src/components/GameTabs.vue
@@ -61,7 +61,7 @@
               <VideoPlayer
                 :options="returnVideoOptions(game.taskData?.tutorialVideo)"
                 :onVideoEnd="updateVideoCompleted"
-                :onVideoStarted="updateVideoStarted"
+                :onVideoStart="updateVideoStarted"
                 :taskId="game.taskId"
               />
             </div>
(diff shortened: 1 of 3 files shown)


richford (approved):
Conditional approval upon tests passing.

Emily-ejag (approved):

src/components/GameTabs.vue
```diff
@@ -93,6 +98,22 @@ const taskCompletedMessage = computed(() => {
   return t('gameTabs.taskCompleted');
 });
 
+const updateVideoStarted = async (taskId) => {
+  try {
+    await authStore.roarfirekit.updateVideoMetadata(selectedAdmin.value.id, taskId, 'started');
+  } catch (e) {
+    console.log('Error while updating video completion', e);
```
richford, line 105: Catching and then not throwing anything will "swallow" this error. This might be what we want, but I just want to make sure. Is the intention to swallow this error in order to prevent it from disrupting gameplay?

In any case, maybe we want to use `console.error` instead.

src/components/GameTabs.vue
```diff
@@ -93,6 +98,22 @@ const taskCompletedMessage = computed(() => {
...
+  }
+};
+
+const updateVideoCompleted = async (taskId) => {
+  try {
+    await authStore.roarfirekit.updateVideoMetadata(selectedAdmin.value.id, taskId, 'completed');
+  } catch (e) {
+    console.log('Error while updating video completion', e);
```
richford, line 113: Same here.

richford (changes requested):

richford (approved):

PR #491, opened 2024-04-05
Add test and demo data checkboxes

lucasxsong:
Adds test and demo checkboxes to:
- register task
- create administrator
- create organization


richford (changes requested):
I really like the visual changes. It's looking good. The code for the administration, administrator, and org forms all looks good. For the task/variant forms, I noticed the following:
- Bug: If you click on the text for any of the boxes, it selects/deselects the first box. Clicking on the actual checkbox results in the correct behavior, but clicking on the text itself always toggles the first box.
- I think we need to store separate test/demo variables for the task and the variant. If the task is test or demo, it should force the variant to be test or demo. But if the task is live, then the variant could still optionally be test or demo.

richford (approved):

PR #497, opened 2024-04-09
Add Administration Search

lucasxsong:
image

All administrations are being queried and stored client side (~150) in this update. Navigation between pages is noticeably snappier, and search results also return quickly. I observed a small increase in load time (2.5s to load on roar.education to 3.5s locally)


src/helpers/query/administrations.js
```diff
@@ -261,3 +261,56 @@ export const administrationPageFetcher = async (
...
+    const administrations = _uniqBy(flattened, 'id')
+      .filter((a) => a[orderField] !== undefined)
+      .sort((a, b) => {
+        if (orderDirection === 'ASCENDING') return 2 * +(a[orderField] > b[orderField]) - 1;
+        if (orderDirection === 'DESCENDING') return 2 * +(b[orderField] > a[orderField]) - 1;
+        return 0;
+      });
+    return administrations;
+  }
+};
```
richford, line 316: Could we achieve the same thing by setting `page = 0` and `pageLimit = Number.MAX_SAFE_INTEGER` in the old `administrationPageFetcher`.

src/pages/HomeAdministrator.vue
```diff
@@ -1,32 +1,61 @@
...
+            <div class="flex flex-column gap-1">
+              <small id="search-help" class="text-gray-400">Search by administration name</small>
+              <div class="flex align-items-center">
+                <PvInputGroup>
+                  <PvInputText
+                    id="search"
+                    v-model="searchInput"
+                    @keyup.enter="onSearch"
+                    placeholder="Search administrations"
+                  >
```
richford, line 19: Could we use Primevue's autocomplete component (https://primevue.org/autocomplete/) to suggest things as the user types?

richford (changes requested):
@lucasxsong I love it! I made two comments that imply some requested changes:
- Can we avoid writing a new `administrationFetcher` function by passing in `page = 0` and `pageLimit = Number.MAX_SAFE_INTEGER` to the existing `administrationPageFetcher`?
- Can we use [Primevue's autocomplete component](https://primevue.org/autocomplete/) to suggest things as the user types?

richford:
Additionally, could you add a Cypress test for searching by name?

richford (approved):
LGTM. Approved!. Let's wait for Cypress tests to pass and get a second approval from either @Emily-ejag or @ksmontville.

Emily-ejag (approved):

Emily-ejag (changes requested):
It is not working for PartnerAdmins therefore the tests are failing

richford (changes requested):
I was thinking about this a bit more and I'd like to request an additional refactor. Currently, with `onSortChange` we initiate a new query. However, I think we should remove the `orderBy` from the query key so that we don't invalidate the query by sorting. After all, we already have all of the administration data locally. So I think we should just sort locally without running another query.

- Can you change the `onSortChange` function to edit `filteredAdministrations` in place?
- Sorting and searching can also interact. When a user sorts and then searches, we should return the results in their previously requested sort order. Similarly, when a user searches, then sorts, we should only sort the administrations returned by the search pattern.
- Lastly, once @Emily-ejag merges #498, then partner admins will see the public facing name stored in the `publicName` field. Given that the Cypress tests are failing here, I think it might be best to merge her PR first. Can you work with @Emily-ejag to edit `onSortChange` to sort by `publicName` if the user is a partner admin and by `name` if the user is a super admin?
- Similarly, can you search by `publicName` if the user is a partner admin and `name` if the user is a super admin?
- In the two cases above, if an administration doesn't have the `publicName` field, it should fall back on the `name` field.

cypress/e2e/super-admin/searchAdministrations.spec.cy.js
```diff
@@ -0,0 +1,15 @@
+const roarDemoAdministrationName = Cypress.env('testPartnerAdministrationName');
+
```
ksmontville, line 2:
```suggestion
const timeout = Cypress.env('timeout')
```

cypress/e2e/super-admin/searchAdministrations.spec.cy.js
```diff
@@ -0,0 +1,15 @@
...
+describe(
+  'The admin user can navigate to the view administration page, ' + 'and can see search for an administration',
+  () => {
+    it('Activates the admin sidebar, then searches for an administration.', () => {
+      cy.login(Cypress.env('superAdminUsername'), Cypress.env('superAdminPassword'));
+      cy.navigateTo('/');
+      cy.get("[data-cy='search-input'").click().type(roarDemoAdministrationName).type('{enter}');
+      cy.get('body').contains('You searched for');
+      cy.get('div').contains('Dates');
```
ksmontville, line 11:
```suggestion

```

ksmontville (changes requested):

richford (approved):

Emily-ejag (approved):

ksmontville (approved):

PR #500, opened 2024-04-10
update individual score reports

lucasxsong:
as per https://github.com/yeatmanlab/roar/issues/246


richford (approved):
Code looks good, all but one test are passing, and I have visually confirmed that this satisfies the requests in https://github.com/yeatmanlab/roar/issues/246.

However, I did notice that some Propel students are marked as having unreliable runs when I didn't expect it. For example, can you take a look at PID pro-gr6+-fd79e334. They have unreliable runs listed in their individual score report. But their scores are not highlighted as unreliable in the group score report. Their runs on firestore are marked as unreliable but they do not have any flags. What gives.

This is a problem on production too so it should not block this PR. I'm approving.

PR #507, opened 2024-04-15
Fixing size of Create Administration inputText

Emily-ejag:
increasing to 50vh instead of 20


package.json
```diff
@@ -61,6 +61,7 @@
     "primeflex": "^3.1.3",
     "primeicons": "^5.0.0",
     "primevue": "^3.32.1",
+    "sharp": "^0.33.3",
```
richford, line 64: Why did we add this?

src/components/CreateAdministration.vue
```diff
@@ -24,10 +24,10 @@
               <PvInputText
                 id="administration-public-name"
                 v-model="state.administrationPublicName"
-                style="width: 20vh"
+                style="width: 50vh"
```
richford, line 27: I'd rather not specify the width here and instead use the `formgrid` class to make things more responsive.
```suggestion
```

Can you combine the `administration-name`, `administration-public-name`, and `dates` input using something like `class="field col-12 md:col-6"`? See https://primeflex.org/formlayout#advanced for a more complete example. I think for sufficiently wide screens, the three should all be on the same line. But as the screen gets narrower, they should each be on their own line.

src/components/CreateAdministration.vue
```diff
@@ -24,10 +24,10 @@
...
                 data-cy="input-administration-name"
               />
-              <label for="administration-public-name" class="w-full">Public Administration Name</label>
+              <label for="administration-public-name" style="width: 50vh">Public Administration Name</label>
```
richford, line 30: If we implement the section above, then this change can be reverted back to how it was.
```suggestion
              <label for="administration-public-name" class="w-full">Public Administration Name</label>
```

richford (changes requested):
Thanks Emily! You're right that this needs fixing. I put in some requested changes to use the `formgrid` class rather than hardcoding the width.

package.json
Emily-ejag: It gave an error for running that I needed to install that, but I can definitely take it out

ksmontville (approved):

richford:
@Emily-ejag I pushed some commits. I was thinking something more along those lines.

ksmontville (approved):

lucasxsong (approved):

richford (approved):

PR #522, opened 2024-04-29
refactor individual score report tasks

lucasxsong:
Refactors individual score report to move most of the conditional rendering logic in `<template>` to an updated `computedTaskData` map.


src/components/reports/IndividualScoreReportTask.vue
```diff
@@ -1,18 +1,15 @@
 <template>
   <div v-for="task in computedTaskData" :key="task" class="align-self-start">
     <div
-      v-if="rawOnlyTasks.includes(task.taskId) && isNaN(getRawScore(task.taskId))"
+      v-if="rawOnlyTasks.includes(task.taskId) && !task.rawScore"
```
Emily-ejag, line 4: I am curious on what would you like to evaluate here, before you had not a number. but in this case it will evaluate to true if it is 0 as well. Is there any chance rawScore is 0?

kellyel (approved):
⭐

PR #533, opened 2024-05-03
Display percent correct in score table and update export csv to attach percent correct, num attempted, and num correct Data

lucasxsong:


src/pages/ScoreReport.vue
```diff
@@ -673,6 +673,18 @@ const computeAssignmentAndRunData = computed(() => {
           tags: scoreFilterTags,
         };
 
+        // if task is a raw score only task, add percentage correct, num attempted, and num correct to the scores object
+        if (rawOnlyTasks.includes(taskId)) {
```
kellyel, line 677: If I remember correctly, we want to exclude frase from the new behavior. So maybe the logic could be 
```js
if (taskId === 'sre-es'){
// Original raw-only behavior
} else if (rawOnlyTasks.includes(taskId) {
// New percent-correct behavior
}
```
Taking that a bit further, maybe a new percent-correct blacklist in `reports.js`? so we could check if the task is IN rawOnlyTasks and NOT IN percentCorrectBlacklist to decide whether to display true raw score or percent correct. Let me know your thoughts

lucasxsong: @kellyel, Thanks for the suggestion. What do you think about instead using a `rawPercentCorrectWhitelist`, to reduce on conditional overhead?

kellyel (approved):
LGTM!

PR #534, opened 2024-05-05
Public facing names to participant view

Emily-ejag:
adding public facing names to participant view and fix box position

### DIFF
--- a/src/pages/HomeParticipant.vue
+++ b/src/pages/HomeParticipant.vue
@@ -7,9 +7,9 @@
       </div>
       <div v-else>
         <h2 v-if="adminInfo?.length == 1" class="p-float-label dropdown-container">
-          {{ adminInfo.at(0).name }}
+          {{ adminInfo.at(0).publicName || adminInfo.at(0).name }}
         </h2>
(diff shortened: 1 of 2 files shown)


src/pages/HomeParticipant.vue
```diff
@@ -25,14 +25,14 @@
...
           >
-            <div class="assignment-select-container flex flex-row justify-content-between">
-              <div class="flex flex-column">
+            <div class="assignment-select-container flex flex-row justify-content-between justify-content-start">
+              <div class="flex flex-column align-content-start justify-content-start w-3">
                 <PvDropdown
                   v-model="selectedAdmin"
                   :options="adminInfo ?? []"
-                  option-label="name"
+                  option-label="publicName"
```
ksmontville, line 35: We should somehow account for the case where publicName is undefined so that the list rendering does not throw an error when it searches for the `publicName` key but cannot find it.

ksmontville (approved):

src/pages/HomeParticipant.vue
Emily-ejag: you are right, I added a v-if clause

richford (approved):

PR #535, opened 2024-05-06
Score Table Updates

lucasxsong:
- Updates raw score tags to white background
- Makes score font bold for easier readability/higher visual weight
- Fixes bug where students were getting hexa-duped in progress table


Emily-ejag (approved):
I think the logic is good, I just think if they finished letter already the title should say something different than "Scores Under Development" when hover over, unless that is intended
image

PR #543, opened 2024-05-08
Register users UI polish

Zio-4:


src/pages/LEVANTE/RegisterUsers.vue
```diff
@@ -286,6 +308,9 @@ async function submitUsers() {
   }
 
   try {
+    // showErrorTable.value = true;
+    // throw new Error('test');
```
ksmontville, line 312:
```suggestion
```
Feel free to remove these unless you are leaving them in for debugging or some other purpose.

ksmontville (approved):

src/pages/LEVANTE/RegisterUsers.vue
Zio-4: Yea will keep these for now. Still need to add UI feedback for errors.

PR #545, opened 2024-05-09
Increase Node Memory

ksmontville:
Increase heap size to 8 MB when building project.


ksmontville:
Our packages are getting quite large again!

image

Zio-4:
Is this because the ROAR tasks still hold their assets (media / CSVs / translations)?

richford (approved):

ksmontville:
> Is this because the ROAR tasks still hold their assets (media / CSVs / translations)?

Most likely has something to do with it. We have made some efforts to reduze chunk size on the dashboard, but the same thing needs to happen to the apps as well.

PR #546, opened 2024-05-09
Adding audio buttons child survey

asengupta3:


src/helpers/audio.js
```diff
@@ -0,0 +1,46 @@
...
+      // Asynchronously decode the audio file data in request.response
+      this.context.decodeAudioData(
+        request.response,
+        (buffer) => {
+          if (!buffer) {
+            alert('error decoding file data: ' + url);
+            return;
+          }
+          this.bufferList[index] = buffer;
+          if (++this.loadCount === Object.keys(this.urlListMap).length) this.onload(this.bufferList);
```
Zio-4, line 24: For readability, could we increment on the previous line and then compare?

src/pages/LEVANTE/UserSurvey.vue
```diff
@@ -10,6 +10,42 @@ import { useRouter } from 'vue-router';
...
+    'SchoolFun',
+    'SchoolGiveUp',
+    'SchoolHappy',
+    'SchoolSafe',
+    'TeacherLike',
+    'TeacherListen',
+    'TeacherNice',
+  ];
+
+  const baseURL = `https://storage.googleapis.com/road-dashboard/child-survey/${parsedLocale}/shared/`;
```
Zio-4, line 43: Instead of hardcoding these, could we get them directly from the bucket? You can do a request to list out all the objects in a bucket and them parse them by the folder they live in. 

```suggestion
  `https://storage.googleapis.com/storage/v1/b/${bucketName}/o`
```

Zio-4 (changes requested):

src/pages/LEVANTE/UserSurvey.vue
Zio-4: Adding on to this, this way we could also use this for teacher and parent questionnaires.

asengupta3: I think I have set it up in the way where this is configurable. But not sure how to pass that in. Maybe we should do it as a separate PR?

Zio-4: Agreed. This is good for now.

kellyel (approved):

Zio-4 (approved):

PR #557, opened 2024-05-13
Refactor Subscore Datatables to use client-side data. 

lucasxsong:
Also fixes a bug with frase showing `NaN` for non-completed assignments.


kellyel (approved):
LGTM!

PR #571, opened 2024-05-14
Only show table columns where entries have values + small report styling changes

lucasxsong:


richford (approved):

richford (changes requested):
The code changes look good to me. However, two of the tests are failing and they are related to score reporting, so I'd prefer to get those to pass before merging.

PR #572, opened 2024-05-14
Add App Version to Run Doc

ksmontville:
This PR adds the app version number to the startAssessment() call in the Vue task component, which commands roar firekit to append the app version number to the user run doc.


Emily-ejag (approved):

src/components/tasks/TaskFluency.vue
```diff
@@ -23,12 +24,15 @@ const props = defineProps({
...
 
 const taskId = props.taskId;
+const { version } = packageLockJson.packages['node_modules/@bdelab/roam-fluency'];
 const router = useRouter();
 const gameStarted = ref(false);
 const authStore = useAuthStore();
 const gameStore = useGameStore();
 const { isFirekitInit, roarfirekit } = storeToRefs(authStore);
 
+console.log(version);
```
lucasxsong, line 34: Is this console.log for development or do we intend to keep it in production?

lucasxsong (approved):
This looks good to me, just added a comment for what I thought was an errant console.log

richford (changes requested):
LGTM. I'll just echo @lucasxsong's comment about the console.log statement. Also, I think this PR is blocked until a new version of roar-firekit is deployed, right?

ksmontville:
> LGTM. I'll just echo @lucasxsong's comment about the console.log statement. Also, I think this PR is blocked until a new version of roar-firekit is deployed, right?

It can be merged but it won't have any effect on th run doc until the Firekit changes go in as well.

src/components/tasks/TaskFluency.vue
ksmontville: Don't need it, thanks!

richford (approved):

PR #573, opened 2024-05-14
Calculating age in months for MEFS, adding lng query string

Zio-4:


src/helpers/ageInMonths.js
```diff
@@ -0,0 +1,15 @@
+export function calculateAgeInMonths(birthMonth, birthYear) {
+  // Get the current date
+  const currentDate = new Date();
```
lucasxsong, line 3: Would we be able to use the `getAgeData` in [roar-utils/utils.js](https://github.com/yeatmanlab/roar-utils/blob/main/src/utils.js) to calculate the ageInMonths?

lucasxsong (approved):

PR #589, opened 2024-05-17
Adding VariantId and VariantName

Emily-ejag:
this pr adds the variant id and variant name, and if the admin has it, then it shows on view params

### DIFF
--- a/src/components/CardAdministration.vue
+++ b/src/components/CardAdministration.vue
@@ -41,22 +41,25 @@
             @click="toggleParams($event, assessmentId)"
           />
         </span>
-        <PvOverlayPanel v-for="assessmentId in assessmentIds" :key="assessmentId" :ref="paramPanelRefs[assessmentId]">
-          <div v-if="props.assessments && (assessment = getAssessment(assessmentId))">
-            <div v-if="assessment.variantId">Variant ID: {{ assessment.variantId }}</div>
-            <div v-if="assessment.variantName">Variant Name: {{ assessment.variantName }}</div>
-          </div>
-          <PvDataTable
-            v-if="props.assessments && (assessment = getAssessment(assessmentId))"
-            striped-rows
-            class="p-datatable-small"
-            table-style="min-width: 30rem"
-            :value="toEntryObjects(params[assessmentId])"
-          >
-            <PvColumn field="key" header="Parameter" style="width: 50%"></PvColumn>
-            <PvColumn field="value" header="Value" style="width: 50%"></PvColumn>
-          </PvDataTable>
-        </PvOverlayPanel>
+        <div v-if="showParams">
+          <PvOverlayPanel v-for="assessmentId in assessmentIds" :key="assessmentId" :ref="paramPanelRefs[assessmentId]">
+            <div v-if="getAssessment(assessmentId).variantId">
+              Variant ID: {{ getAssessment(assessmentId).variantId }}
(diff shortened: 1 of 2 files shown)


src/components/CardAdministration.vue
```diff
@@ -41,19 +41,22 @@
...
-              table-style="min-width: 30rem"
-              :value="toEntryObjects(params[assessmentId])"
-            >
-              <PvColumn field="key" header="Parameter" style="width: 50%"></PvColumn>
-              <PvColumn field="value" header="Value" style="width: 50%"></PvColumn>
-            </PvDataTable>
-          </PvOverlayPanel>
-        </div>
+        <PvOverlayPanel v-for="assessmentId in assessmentIds" :key="assessmentId" :ref="paramPanelRefs[assessmentId]">
+          <div v-if="props.assessments && (assessment = getAssessment(assessmentId))">
```
richford, line 45: Is the `props.assessments` necessary in the `v-if` here. The `PvOverlayPanel` already has a `v-for` over the `assessmentIds`, which is a computed value that depends on `props.assessments` being defined. Also, my understanding is that one does not have to explicitly reference `props.` in the template portion.

src/components/CardAdministration.vue
```diff
@@ -41,19 +41,22 @@
...
+            <div v-if="assessment.variantId">Variant ID: {{ assessment.variantId }}</div>
+            <div v-if="assessment.variantName">Variant Name: {{ assessment.variantName }}</div>
+          </div>
+          <PvDataTable
+            v-if="props.assessments && (assessment = getAssessment(assessmentId))"
```
richford, line 50: Same comment here about the `v-if`

src/components/CardAdministration.vue
richford: Actually, I don't really understand this `v-if`. Why is it needed? If the point is to set the value of a variable called `assessment` to `getAssessment(assessmentId)`, why not just use `getAssessment(assessmentId)` below.

Alternatively, we could adjust the `v-for` loop above to loop over `assessments` instead of `assessmentIds`.

richford (changes requested):

src/components/CardAdministration.vue
```diff
@@ -41,19 +41,22 @@
...
-              class="p-datatable-small"
-              table-style="min-width: 30rem"
-              :value="toEntryObjects(params[assessmentId])"
-            >
-              <PvColumn field="key" header="Parameter" style="width: 50%"></PvColumn>
-              <PvColumn field="value" header="Value" style="width: 50%"></PvColumn>
-            </PvDataTable>
-          </PvOverlayPanel>
-        </div>
+        <PvOverlayPanel v-for="assessmentId in assessmentIds" :key="assessmentId" :ref="paramPanelRefs[assessmentId]">
```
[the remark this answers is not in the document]
Emily-ejag: True, I re-added it

src/components/CardAdministration.vue
Emily-ejag: Yeah, no need here

src/components/CardAdministration.vue
Emily-ejag: you are right, It was to check first if we have the id and name, and assigned to a variable. But yes, I can take it out

richford (approved):

PR #609, opened 2024-05-24
[LHT-328]: Sanitizing and validating user rows

asengupta3:
This code should:
1. Remove any unnecessary columns that do not belong to allFields
2. It should also throw errors if, multiple ids exist for the same user, no ids exist for the user and if there are duplicate ids across all users

### DIFF
--- a/src/pages/LEVANTE/RegisterUsers.vue
+++ b/src/pages/LEVANTE/RegisterUsers.vue
@@ -170,16 +170,33 @@ const sanitizeAndValidateSubmittedUsers = (users) => {
     const idList = [user['childId'], user['teacherId'], user['parentId']].filter((a) => 0 || !!a);
     if (idList.length > 1) {
       addErrorUser(user, `Multiple ids present for the user: ${user.id}`);
-      throw new Error(`Mutliple ids found for user, please see the table below for more details`);
+      // throw new Error(`Mutliple ids found for user, please see the table below for more details`);
     }
     if (userIdMap[idList[0]]) {
-      addErrorUser(user, `Duplicate id present with id: ${idList[0]}`);
-      throw new Error(`Duplicate ids found for user, please see the table below for more details`);
-    } else {
-      userIdMap[idList[0]] = true;
+      addErrorUser(
+        user,
+        `Duplicate id present for the user: ${user.id}, duplicate id: ${idList[0]} shared with user: ${
+          userIdMap[idList[0]].userId
+        }`,
+      );
+      // throw new Error(`Duplicate ids found for user, please see the table below for more details`);
+    } else if (!!idList[0]) {
+      // Avoid undefined
+      userIdMap[idList[0]] = {
+        userId: user.id,
+      };
     }
     return acc;
   });
+
+  if (errorUsers.value.length) {
+    toast.add({
+      severity: 'error',
+      summary: 'Error: There is some issues with user validation. Please check below for more details',
+      life: 5000,
+    });
+    return [];
+  }
 
   return filteredFieldsUsers;
 };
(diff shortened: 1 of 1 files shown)


src/pages/LEVANTE/RegisterUsers.vue
```diff
@@ -155,6 +155,35 @@ const allFields = [
...
+    if (userIdMap[idList[0]]) {
+      throw new Error(`User Row validation:, Duplicate id present for ${JSON.stringify(user)}, with id: ${idList[0]}`);
+    } else {
+      userIdMap[idList[0]] = true;
+    }
+    return acc;
+  });
+
+  return filteredFieldsUsers;
+};
```
Zio-4, line 186: This is great. Could you change the error handling to add the row/user to the error table and pass in the message there? Additionally, we want the toast notification message to be more informative rather than hold the actual error values. So maybe an error would be something like "Multiple ids found for some row(s)". Then the error table would display the rows/users with the problems. See the example below.
Screenshot 2024-05-29 at 4 58 40 PM

src/pages/LEVANTE/RegisterUsers.vue
```diff
@@ -155,6 +155,35 @@ const allFields = [
...
+    const acc = {};
+    Object.keys(user).forEach((uKey) => {
+      if (allFieldsIdsList.includes(uKey)) {
+        acc[uKey] = user[uKey];
+      }
+    });
+    // Allowing 0 as a valid id
+    const idList = [user['childId'], user['teacherId'], user['parentId']].filter((a) => 0 || !!a);
+    if (idList.length === 0) {
+      throw new Error(`User Row validation:, No Child, Teacher or Parent Id found for: ${JSON.stringify(user)}`);
```
Zio-4, line 172: Sorry this is on me, I should have given you more context. We allow the site researchers to register children only. They don't always have the parent/teacher relationships beforehand. In this case child, teacher, and parent id columns would be left blank.

Zio-4 (changes requested):

src/pages/LEVANTE/RegisterUsers.vue
```diff
@@ -155,6 +155,34 @@ const allFields = [
...
+    });
+    // Allowing 0 as a valid id
+    const idList = [user['childId'], user['teacherId'], user['parentId']].filter((a) => 0 || !!a);
+    if (idList.length > 1) {
+      addErrorUser(user, `Multiple ids present for the user: ${user.id}`);
+      throw new Error(`Mutliple ids found for user, please see the table below for more details`);
+    }
+    if (userIdMap[idList[0]]) {
+      addErrorUser(user, `Duplicate id present with id: ${idList[0]}`);
+      throw new Error(`Duplicate ids found for user, please see the table below for more details`);
```
Zio-4, line 177: There seems to be a problem with this logic. If I try to register only child users this error is thrown.

Input:
Screenshot 2024-06-06 at 12 28 00 PM

Error:
Screenshot 2024-06-06 at 12 28 42 PM

src/pages/LEVANTE/RegisterUsers.vue
```diff
@@ -155,6 +155,34 @@ const allFields = [
...
+    Object.keys(user).forEach((uKey) => {
+      if (allFieldsIdsList.includes(uKey)) {
+        acc[uKey] = user[uKey];
+      }
+    });
+    // Allowing 0 as a valid id
+    const idList = [user['childId'], user['teacherId'], user['parentId']].filter((a) => 0 || !!a);
+    if (idList.length > 1) {
+      addErrorUser(user, `Multiple ids present for the user: ${user.id}`);
+      throw new Error(`Mutliple ids found for user, please see the table below for more details`);
```
Zio-4, line 173: This error message is a little confusing. For the bad row input, it's hard to decipher what actually needs to be changed. There are 3 things wrong with this input. 1. The parent id does not exist. 2. It is a Teacher so they should not have the teacher id column filled out. 3. They have month and birth year which a Teacher should not have. There is a lot of validation needed here but we should split it into multiple PRs. For now, let's just change the message to something like "One or more columns contain incorrect values. Please review and correct the data."

Input:
Screenshot 2024-06-06 at 12 30 02 PM

Output: 
Screenshot 2024-06-06 at 12 30 13 PM

Zio-4 (changes requested):

asengupta3:
Thanks @Emily-ejag for flagging this. This has been in a limbo for a while. @Zio-4 should we close this out and then deal with it in the fork or open a new ticket?

Zio-4:
Yes. Closing this since it's no longer relevant.

PR #615, opened 2024-05-28
Game route path changes + vocab translations

Zio-4:
These changes were required so that the LEVANTE version of Vocab would launch correctly. The two tasks have the same Id so it would default to the ROAR vocab task component.

### DIFF
--- a/src/router/index.js
+++ b/src/router/index.js
@@ -164,7 +164,7 @@ const routes = [
     meta: { pageTitle: 'Fluency-Alpaca ES' },
   },
   {
-    path: '/game/:taskId',
+    path: '/game/core-tasks/:taskId',
     name: 'Core Tasks',
     component: () => import('../components/tasks/TaskLevante.vue'),
     props: true,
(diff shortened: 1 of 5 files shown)


src/translations/es/es-componentTranslations.json
richford, line 1: Are you missing the corresponding changes in the /src/translations/es/co/es-co-componentTranslations.json file?

richford (changes requested):

src/translations/es/es-componentTranslations.json
Zio-4: No, these are general Spanish translations. Not Colombia specific.

src/router/index.js
Emily-ejag, line 1: maybe this will affect the current administrations, I created one with trog and it is working fine on production, but on the pr link it is not found: 
image

Emily-ejag: Unless I manually add /core-tasks: 
image

Emily-ejag (changes requested):

src/router/index.js
```diff
@@ -164,7 +164,7 @@ const routes = [
     meta: { pageTitle: 'Fluency-Alpaca ES' },
   },
   {
-    path: '/game/:taskId',
+    path: '/game/core-tasks/:taskId',
```
Emily-ejag, line 167: This one:
```suggestion
    path: '/game/core-tasks/:taskId',
```

src/router/index.js
Zio-4: Fixed. This won't break current administrations since it's just a routing change.

Emily-ejag (approved):

PR #645, opened 2024-06-06
Fix: Blank Screen 

kellyel:
## Proposed changes
I believe this will fix the issue of the games tabs not appearing after each game.
Fixes: https://github.com/yeatmanlab/roar/issues/281

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist

## Further comments
I understand that this change was originally made to decrease the package size of the app. However I feel that in this case, it is better to reverse the change to prevent this issue from occurring. 

My theory as to what is happening is as follows. Those two components are being imported on the onMounted hook. Vue tries to reuse components whenever possible, thus will not call the onMounted hook in every case [[1]](https://stackoverflow.com/questions/57699473/vuejs-hook-mounted-is-not-called-when-page-is-re-open-by-vue-router). There seem to be some cases where when the user navigates back to a page, the onMounted hook is not fired. This is supported by the fact that in all cases we have evidence of, those two components are the only ones missing with no other apparent errors.

### DIFF
--- a/src/pages/HomeParticipant.vue
+++ b/src/pages/HomeParticipant.vue
@@ -100,6 +100,8 @@ import { useQuery } from '@tanstack/vue-query';
 import { fetchDocById, fetchDocsById, fetchSubcollection } from '../helpers/query/utils';
 import { getUserAssignments } from '../helpers/query/assignments';
 import ConsentModal from '../components/ConsentModal.vue';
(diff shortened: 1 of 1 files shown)


Emily-ejag (approved):

lucasxsong (approved):
Nice! This was also the fix to the videoPlayer not showing up on first render. Glad you caught this.

PR #665, opened 2024-06-14
Update SRE version to 1.15.10

richford:
This PR updates the version of `@bdelab/roar-sre` to 1.15.10.


Emily-ejag:
no need to update

PR #733, opened 2024-08-07
 Fix session timeout handler

maximilianoertel:
## Proposed changes

This PR fixes the faulty behaviour of the existing session idle detection timer, initially implemented using the `useIdle ` composable. 

The issues previously reported in the corresponding ticket were primarily caused by how browsers treat inactive tabs. More precisely, when a tab becomes inactives, browsers throttle JS execution, directly impact the timer set by `useIdle`. To circumvent this, `useIdle` supports  the use of visible change event listeners. Unfortunately, when a user returns to the tab after being active on another site for a while, the `useIdle` timer treats the visible change as activity and resets the idle timer immediately. 

To avoid these issues, this PR disables the visibility change listener and manually implements corresponding event listeners. Simultaneously, we store an internal last active timestamp when the user leaves the active tab to allow us to calculate the true idle time upon the user's return and trigger the desired business logic. 

In order to facilitate development and testing, the timer can be configured using two environment variables: 
- `VITE_AUTH_SESSION_TIMEOUT_IDLE_THRESHOLD` The duration in ms until we consider a user to be idle, defaults to 15min.
- `VITE_AUTH_SESSION_TIMEOUT_COUNTDOWN_DURATION` The duration of the countdown displayed in the dialog, defaults to 60 seconds. 

Whilst previously only implemented on the Homepage, the session timer is now implemented at the root of the web application to ensure the timer runs on all pages. 

Screenshot 2024-08-07 at 23 16 31

Docs are updated in PR yeatmanlab/roar-docs/pull/14

## Types of changes
- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist

n/a

## Further comments
Additionally, this PR introduces these (minor out-of-scope) changes (that, if required, can be moved to a separate PR): 
* A `.nvmrc` file to facilitate local development practices
* A `.editorconfig` file to enhance development experience 

Finally, it might be worth considering adding a small alert / card within the sign-in page to inform a user they have been signed-out due to inactivity. 

Fixes yeatmanlab/roar/issues/241


Emily-ejag:
Great solution, I have one question. When I tried in production (using the PR link) I noticed that after the 15 min it took me to this page: 
image
 Is this the correct behavior? Or should It take you to the signing page?

ksmontville (approved):

maximilianoertel:
> Great solution, I have one question. When I tried in production (using the PR link) I noticed that after the 15 min it took me to this page: image Is this the correct behavior? Or should It take you to the signing page?

Thanks for catching this @Emily-ejag! From what I noticed: 
- This only happens on the homepage, on other pages (i.e. `/register-students`) you are signed-out correctly. 
- On the homepage, after seeing the "You have no assignments" empty state, if you refresh or try to navigate to another page, you'll be redirected to the sign-in page as expected. 

Can you confirm this on your end?

Emily-ejag (approved):

PR #735, opened 2024-08-07
Update dashboard with `publicName` and `studentFacingName`

lucasxsong:
## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist


src/helpers/reports.js
```diff
@@ -11,114 +11,224 @@ import { getGrade } from '@bdelab/roar-utils';
...
+    extendedName: 'Single Word Recognition',
+    order: 6,
+  },
+  sre: {
+    name: 'Sentence',
+    publicName: 'ROAR - Sentence',
+    studentFacingName: 'Sentence',
+    extendedTitle: 'ROAR - Sentence',
+    extendedName: 'Sentence Reading Efficiency',
+    order: 7,
```
ksmontville, line 66: Is it possible to use your taskDictionary() to define some of these fields?

One idea might be to start storing all of these fields in the task document intself, and then write a function(s) that grabs the documents and sets these fields. 

This would make it easier to modify task documents in the future and not have to change so many hard-coded values.

Just an idea for the future, nothing we need to do immediately.

ksmontville (approved):

src/helpers/reports.js
lucasxsong: These fields are not being used to display titles on the front-end – all public names are pulled from the Firestore entry. Those entries are a relic from before and I thought I would leave them in for reference.

PR #736, opened 2024-08-08
Introduce Vitest for unit and snapshot testing

maximilianoertel:
## Proposed changes
This PR adds [Vitest](https://vitest.dev/) to the project for the purpose of progressively introducing unit and snapshot testing to the project. 

For compatibility purposes, this PR also upgrades the Vite dependencies to their latest version as well as upgrades from `@vueuse/head` to `@unhead/vue`. The latter was necessary due to some incompatibilities with the latest Vite version and the sunset notice provided on the `@vueuse/head` repo.

Once the first tests have been written, we should consider extending the CI workflows to run the tests against PRs as well as measure and publish test coverage. 

## Types of changes
- [ ] Documentation Update
- [x] Repository Maintenance

## Checklist

n/a

## Further comments
n/a


cypress.config.cjs
lucasxsong, line 1: Does the`.cjs` file extension stand for CommonJS? What are the differences between that and a regular `.js` extension?

src/App.vue
```diff
@@ -10,7 +10,8 @@
     <!-- Twitter -->
     <meta name="twitter:title" content="ROAR Web Query" />
     <meta name="twitter:description" content="A web-based tool to query ROAR assessment data!" />
-  </AppHead>
+  </Head>
```
lucasxsong, line 13: Thanks for catching this, I didn't realize that our `AppHead` component was just using the default `Head` component from @vueuse

vitest.config.js
```diff
@@ -0,0 +1,35 @@
...
+  viteConfig,
+  defineConfig({
+    test: {
+      globals: true,
+      environment: 'happy-dom',
+      exclude: [...configDefaults.exclude],
+      root: fileURLToPath(new URL('./', import.meta.url)),
+      watch: false,
+      setupFiles: ['./vitest.setup.js'],
+      // coverage: {
```
lucasxsong, line 18: Is the desired behavior to run our tests without coverage? Or are these configuration options just temporarily disabled?

lucasxsong (approved):

vitest.config.js
maximilianoertel: Eventually, we'll definitely want to enable this but since there aren't any tests at the moment nor anywhere to publish the test coverage, I figured I'd leave it here commented out. 

Once we have some tests in place, we should adjust the CI pipeline to execute the tests and coverage measurement. For the latter, we can publish the stats in the PRs as a comment and ideally publish those to a tool like CodeClimate (or similar).

cypress.config.cjs
maximilianoertel: The `.cjs` extension does indeed stand for CommonJS and using that extension ensures that the file/module is treated as such. The `.mjs` extension is the ESM equivalent.

This was necessary as the newer Vite version deprecated the CJS build of Vite's Node API ([docs](https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated)). As this project primarily uses ESM, I set `"type": "module"` on the `package.json` and had two options for the Cypress config: either convert the file to use ESM or add the `.cjs` extension. Since the Cypress documentation uses CJS syntax, I figured it would be best to keep it that way and simply use the correct file extension. 

Hope that clarifies your questions, but please let me know if I missed something

Emily-ejag (approved):

PR #738, opened 2024-08-09
Edit Orgs

kellyel:
## Proposed changes
This PR adds a modal and form for editing organizations. 

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist

### DIFF
--- a/src/components/EditOrgsForm.vue
+++ b/src/components/EditOrgsForm.vue
@@ -55,6 +55,7 @@ import { useAuthStore } from '@/store/auth';
 import { storeToRefs } from 'pinia';
 import { fetchDocById } from '@/helpers/query/utils';
 import { useQuery } from '@tanstack/vue-query';
+import PvChips from 'primevue/chips';
 import _isEmpty from 'lodash/isEmpty';
 const props = defineProps({
   orgType: { type: String, required: true },
@@ -96,7 +97,6 @@ const localOrgData = ref({
 });
 
 const setupOrgData = (orgData) => {
-  console.log('setupOrgData', orgData);
   let org = {
     name: orgData?.name ?? '',
     abbreviation: orgData?.abbreviation ?? '',
@@ -152,7 +152,6 @@ onMounted(() => {
 watch(
   () => localOrgData,
   (orgData) => {
-    console.log('emitting new orgData', orgData);
     emit('update:orgData', orgData.value);
   },
   { deep: true, immediate: false },
--- a/src/components/EditOrgsForm.vue
+++ b/src/components/EditOrgsForm.vue
@@ -1,0 +1,207 @@
+<template>
+  <div class="flex flex-column gap-3">
+    <div class="form-container">
+      <div class="form-field w-full">
+        <label :class="{ 'font-light uppercase text-sm': !editMode }">Name</label>
+        <div v-if="!editMode" :class="{ 'text-xl': !editMode }">{{ serverOrgData.name ?? 'None' }}</div>
+        <PvInputText v-else v-model="localOrgData.name" />
+      </div>
+      <div class="form-field w-full">
+        <label :class="{ 'font-light uppercase text-sm': !editMode }">Abbreviation</label>
+        <div v-if="!editMode" :class="{ 'text-xl': !editMode }">{{ serverOrgData.abbreviation ?? 'None' }}</div>
+        <PvInputText v-else v-model="localOrgData.abbreviation" />
+      </div>
+    </div>
+    <div class="form-field">
+      <label :class="{ 'font-light uppercase text-sm': !editMode }">Address</label>
+      <div class="p-inputgroup">
+        <span class="p-inputgroup-addon">
+          <i class="pi pi-map"></i>
+        </span>
+        <GMapAutocomplete
+          :placeholder="localOrgData.address?.formattedAddress ?? 'Enter an address'"
+          :options="{
+            fields: ['address_components', 'formatted_address', 'place_id', 'url'],
+          }"
+          class="p-inputtext p-component w-full"
+          data-cy="input-address"
+          @place_changed="setAddress"
+        >
+        </GMapAutocomplete>
+      </div>
+    </div>
+    <div v-if="orgType === 'districts'" class="form-field">
+      <label :class="{ 'font-light uppercase text-sm': !editMode }">NCES ID</label>
+      <div v-if="!editMode" :class="{ 'text-xl': !editMode }">{{ serverOrgData.ncesId ?? 'None' }}</div>
+      <PvInputText v-else v-model="localOrgData.ncesId" />
+    </div>
+    <div class="form-field">
+      <label :class="{ 'font-light uppercase text-sm': !editMode }">Tags</label>
+      <PvChips v-model="localOrgData.tags" />
+    </div>
+    <div>
+      <PvCheckbox v-model="localOrgData.testData" binary />
+      <label class="ml-1">Test Data</label>
+    </div>
+    <div>
+      <PvCheckbox v-model="localOrgData.demoData" binary />
+      <label class="ml-1">Demo Data</label>
+    </div>
+  </div>
+</template>
+<script setup>
+import { ref, defineProps, onMounted, watch } from 'vue';
+import { useAuthStore } from '@/store/auth';
+import { storeToRefs } from 'pinia';
+import { fetchDocById } from '@/helpers/query/utils';
+import { useQuery } from '@tanstack/vue-query';
+import _isEmpty from 'lodash/isEmpty';
+const props = defineProps({
+  orgType: { type: String, required: true },
+  orgId: { type: String, required: true },
+  editMode: { type: Boolean, default: true },
+});
+
+// +------------+
+// | Initialize |
+// +------------+
+const initialized = ref(false);
+const authStore = useAuthStore();
+const { roarfirekit } = storeToRefs(authStore);
+
+const emit = defineEmits(['modalClosed', 'update:orgData']);
+
+// +----------------------------+
+// | Query for existing orgData |
+// +----------------------------+
+const { data: serverOrgData } = useQuery({
+  queryKey: ['org', props.orgType, props.orgId],
+  queryFn: () => fetchDocById(props.orgType, props.orgId),
+  keepPreviousData: true,
+  enabled: initialized,
+  staleTime: 5 * 60 * 1000, // 5 minutes
+});
+
+// +-------------+
+// | Local State |
+// +-------------+
+const localOrgData = ref({
+  name: null,
+  abbreviation: null,
+  address: null,
+  ncesId: null,
+  tags: [],
+  testData: false,
+  demoData: false,
+});
+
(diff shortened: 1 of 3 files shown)


src/components/EditOrgsForm.vue
```diff
@@ -0,0 +1,207 @@
...
+// +----------------------------+
+// | Query for existing orgData |
+// +----------------------------+
+const { data: serverOrgData } = useQuery({
+  queryKey: ['org', props.orgType, props.orgId],
+  queryFn: () => fetchDocById(props.orgType, props.orgId),
+  keepPreviousData: true,
+  enabled: initialized,
+  staleTime: 5 * 60 * 1000, // 5 minutes
+});
```
maximilianoertel, line 83: What do you think of moving this query out of the component and encapsulate the query logic in a separate file as a reusable composable? This is usually the preferred approach when using something like Tanstack Query as it lets us re-use the queries, write some unit tests for it and overall make the code more maintainable. 

For example, something along the lines of: 

```js
// src/queries/useOrganizationQuery.vue
import { fetchDocById } from '@/helpers/query/utils';
import { useQuery } from '@tanstack/vue-query';
import { ORGANIZATION_QUERY_KEY } from 'constants/QueryKeys'

const useOrganizationQuery = (orgType, orgId, queryOptions = undefined) => {
  return useQuery({
    queryKey: [ORGANIZATION_QUERY_KEY, orgType, orgId],
    queryFn: () => fetchDocById(orgType, orgId),
    keepPreviousData: true,
    ...queryOptions
  });
}

export default useOrganizationQuery
```

```js
// src/constants/queryKeys.js
export const ORGANIZATION_QUERY_KEY = 'org'
```

Then, in any component where we want to use this query we can: 

```vue
<script setup>
import useOrganizationQuery from '@/queries/useOrganizationQuery'

// […]

const { data: orgData, isLoading: loadingOrgData } =
    useOrganizationQuery(
      orgType,
      orgId,
      {
        enabled: initialized
      }
    )

// […]
</script>
```

Additionally, the `staleTime` should ideally be configured globally as it's rarely changed on a case-by-case basis. Configuring a default globally in `main.js` is a good practice and then just override if it's necessary for a single query. 

Was planing on writing a ticket for this but noticed it in this PR, so figured I'd also raise this here.

What do you think?

src/components/EditOrgsForm.vue
```diff
@@ -0,0 +1,207 @@
...
+        <label :class="{ 'font-light uppercase text-sm': !editMode }">Name</label>
+        <div v-if="!editMode" :class="{ 'text-xl': !editMode }">{{ serverOrgData.name ?? 'None' }}</div>
+        <PvInputText v-else v-model="localOrgData.name" />
+      </div>
+      <div class="form-field w-full">
+        <label :class="{ 'font-light uppercase text-sm': !editMode }">Abbreviation</label>
+        <div v-if="!editMode" :class="{ 'text-xl': !editMode }">{{ serverOrgData.abbreviation ?? 'None' }}</div>
+        <PvInputText v-else v-model="localOrgData.abbreviation" />
+      </div>
+    </div>
```
maximilianoertel, line 14: To avoid repetitions and design inconsistencies, what do you think of creating a general `TextInput` component that can be used here and in all other components that include a form? Ideally we would expand this to have `<FormContainer>`, `<TextInput>`, `<Checkbox>`, `<FormSubmit>` (which leverages a `<Button>`), etc.

This would primarily facilitate creating new and updating existing forms as well as future design changes. At the moment, in the existing approach, we would have to update the tailwind classes in all the different components that use a text input instead of updating them in a single place

src/components/EditOrgsForm.vue
```diff
@@ -0,0 +1,207 @@
...
+  if (!_isEmpty(serverOrgData.value)) setupOrgData(serverOrgData.value);
+});
+
+// +---------------------+
+// | Handle update event |
+// +---------------------+
+watch(
+  () => localOrgData,
+  (orgData) => {
+    console.log('emitting new orgData', orgData);
```
maximilianoertel, line 155: Are the various console logs intentional or accidental left overs?

src/components/ListOrgs.vue
```diff
@@ -337,10 +381,52 @@ const showCode = async (selectedOrg) => {
...
+      localOrgData.value,
+      _get(localOrgData.value, 'testData', false),
+      _get(localOrgData.value, 'demoData', false),
+      currentEditOrgId.value,
+    )
+    .then(() => {
+      isSubmitting.value = false;
+      closeEditModal();
+      toast.add({
+        severity: 'success',
```
maximilianoertel, line 413: I know this wasn't introduced by your PR, but taking the opportunity here: ideally we would use constants for these. 

Something like: 

```js
// src/constants/toastConfig.js`

export const TOAST_SEVERITIES = Object.freeze({ 
  SUCCESS: 'success', 
  ERROR: 'error',
  // etc.
})

export const TOAST_DEFAULT_LIFE_DURATION = 3000

File: src/main.js [hunk 5]
```diff
@@ -136,6 +137,7 @@ app.component('PvCard', PvCard);
 app.component('PvChart', PvChart);
 app.component('PvCheckbox', PvCheckbox);
 app.component('PvChip', PvChip);
+app.component('PvChips', PvChips);
```
maximilianoertel, line 140: Curious why you're preferring to load the component globally instead of locally? Same as above, I'm intending to write a ticket for this for generally speaking, but since this is introducing a new global component, I wanted to flag this here as well.

As I'm new to the project I might be not fully familiar with some of the reasonings behind certain approaches but generally speaking, global component registration is pretty bad [for a few reasons](https://vuejs.org/guide/components/registration#local-registration), one of them being tree-shaking.

maximilianoertel (commented):
Had a bit of time and took the liberty to do an initial review – hope you don't mind! 

In addition to the comments below, I would recommend moving the `EditOrgsForm` component to a separate `src/containers/` directory as this is a "smart" component. Creating a separation between "dumb" and "smart" components using the `/components` and `/containers` directory is usually good practice for a growing application.

src/components/ListOrgs.vue
richford: 👍🏻 to @maximilianoertel's comment here.

src/main.js
richford: I agree with @maximilianoertel here. Thanks for bringing it up. I've added a task to the next sprint to prefer local registration over global registration and assigned it to you @maximilianoertel. I think it's a really good point that you bring up but I don't think it should block this PR.

src/components/EditOrgsForm.vue
richford: Agreed. We should remove the console logs after development.

src/components/EditOrgsForm.vue
richford: I really like this suggestion and I've added it as a task to the next sprint (assigned to @maximilianoertel). I think we should definitely do this but I don't think it should be a blocker for this current PR.

src/components/EditOrgsForm.vue
richford: I like this approach. I'm inclined to add this as a separate task rather than hold up this current PR. What do you think @kellyel and @maximilianoertel?

richford (changes requested):
Looking good. I'm requesting changes that mostly just co-sign on @maximilianoertel's comments. I agree with his comments. Of his five comments, I think two are worth holding up this PR, two should be converted into new tasks on the next sprint. And I've asked for your input on whether the remaining comment should be a blocker for this PR or should be a separate task after this PR is merged.

src/components/EditOrgsForm.vue
maximilianoertel: Definitely something that could (and really should) be done as a separate PR, happy to write a ticket for it as I was planing on to do so any way. Let me know!

src/components/EditOrgsForm.vue
maximilianoertel: Agree that this isn't something that shouldn't hold up the PR, that's why I left these comments as "comments" not "change requests" – apologies, should have made that clearer in the comments themselves. But thank you, happy to work on this on the next sprint!

Emily-ejag (approved):
LGTM :)

src/components/ListOrgs.vue
```diff
@@ -431,10 +476,51 @@ const showCode = async (selectedOrg) => {
...
+    })
+    .catch((error) => {
+      toast.add({
+        severity: TOAST_SEVERITIES.ERROR,
+        summary: 'Unexpexcted error',
+        detail: `Unexpected error occurred: ${error.message}`,
+        life: TOAST_DEFAULT_LIFE_DURATION,
+      });
+      isSubmitting.value = false;
+    });
```
maximilianoertel, line 521: This is swallowing the error, or better preventing the error from being logged/reported to Sentry as you're catching the error and then only displaying it to the user. 

Instead, it would be better to report the error back to Sentry as it was done here by @Emily-ejag: 
https://github.com/yeatmanlab/roar-dashboard/blob/65953f10e6aa7189e9c145008616d763bd2a9cb6/src/components/ListOrgs.vue#L350-L358

src/components/ListOrgs.vue
```diff
@@ -431,10 +476,51 @@ const showCode = async (selectedOrg) => {
...
+        severity: TOAST_SEVERITIES.SUCCESS,
+        summary: 'Updated',
+        detail: 'Organization data updated successfully!',
+        life: TOAST_DEFAULT_LIFE_DURATION,
+      });
+    })
+    .catch((error) => {
+      toast.add({
+        severity: TOAST_SEVERITIES.ERROR,
+        summary: 'Unexpexcted error',
```
maximilianoertel, line 516:
```suggestion
        summary: 'Unexpected error',
```

src/components/ListOrgs.vue
```diff
@@ -431,10 +476,51 @@ const showCode = async (selectedOrg) => {
...
+        detail: `Unexpected error occurred: ${error.message}`,
+        life: TOAST_DEFAULT_LIFE_DURATION,
+      });
+      isSubmitting.value = false;
```
maximilianoertel, line 520: Just a detail: to avoid the repetition `isSubmitting.value = false`, this could be moved to a final `.finally()` block

maximilianoertel (changes requested):
Thanks for the updates @kellyel ! 

In addition to the couple of comments below, one more thing: in the current implementation after updating the organisation, the data is not updated for the user despite seeing the success toast. 

To fix this, the `updateOrgData` would need to be moved to a mutation where we can invalidate the queried data – or better, apply an optimistic update without requiring a data refetch (to save on Firestore queries). 

Uploading query-invalidation-compressed.mov…

Is this something you would like to address as part of this PR?

richford (approved):

richford:
Approved but component tests are failing. I'll defer to you on whether you'd like to respond to @maximilianoertel's comment in this PR or open an issue and have one of the devs address it in another.

maximilianoertel (approved):
🚀

PR #743, opened 2024-08-14
Add Firebase App Check

ksmontville:
## Proposed changes
Companion PR: ~~https://github.com/yeatmanlab/roar-firekit/pull/135~~ ~~https://github.com/yeatmanlab/roar-firekit/pull/146~~

Docs PR: https://github.com/yeatmanlab/roar-docs/pull/15

**MAJOR CHANGES**:

- This PR sets up Firebase App Check for use within Firekite and the Dashboard. App Check uses reCAPTCHA tokens to verify that all requests are coming from a list of allowed domains.
- The list of pre-approved domains is `roar.education`, `roar-staging.web.app `for production App Check tokens.
- `localhost` is allowed when developing locally using the debug token, which devs will need in their local `.env` file.
- For PR deploy urls, a new job within the `Build and Deploy` workflow dynamically adds the PR link to the list of approved domains, ensuring that any duplicates are removed. This is achieved with the `gcloud cli` using Firebase credentials stored as repo secrets.

**HOW TO TEST ON PROD**:
1. Open PR link
3. Login as any account type
4. Open the dev console and inspect the network tab for an outgoing request to `https://firestore.googleapis.com/**/*`
5. Check that the headers of the request includes a field for `X-Firebase-AppCheck:` with a token; every request should have this token.
6. Monitor requests here: https://console.firebase.google.com/u/0/project/gse-roar-admin/appcheck/products, https://console.firebase.google.com/u/0/project/gse-roar-assessment/appcheck/products
7. Once ENFORCED mode is on, all requests must have this app check token or they will fail.

**HOW TO TEST ON DEV**:
1. Checkout PR code
2. `npm run dev`
3. Open developer console
4. If you have the debug token saved to your local `.env` file, you should see a console.log() including the debug token. You are successfully bypassing app check in development.
5. If you do not have the debug token saved to your local `.env` file, all requests should fail (?)

## Types of changes

- [x] Documentation Update
- [ ] Repository Maintenance

## Checklist

### DIFF
--- a/.github/workflows/firebase-hosting-preview.yml
+++ b/.github/workflows/firebase-hosting-preview.yml
@@ -44,8 +44,9 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
-      - name: Create .env file with App Check Debug Token
-        run: echo "VITE_APPCHECK_DEBUG_TOKEN=${{ secrets.VITE_APPCHECK_DEBUG_TOKEN }}" >> .env
+      #      # Allows Cypress tests to bypass the app check verification
+      #      - name: Create .env file with App Check Debug Token
+      #        run: echo "VITE_APPCHECK_DEBUG_TOKEN=${{ secrets.VITE_APPCHECK_DEBUG_TOKEN }}" >> .env
 
       - name: Build project
         run: npm run build
@@ -92,7 +93,7 @@ jobs:
         run: |
           CURRENT_DOMAINS=$(gcloud recaptcha keys describe ${{ secrets.FIREBASE_APPCHECK_SITE_KEY_GSE_ROAR_ADMIN }} --format="value(webSettings.allowedDomains)")
           CURRENT_DOMAINS=$(echo $CURRENT_DOMAINS | tr ';' ',')
-          NEW_DOMAIN=${{ needs.build-and-preview.outputs.deployUrl }}
+          NEW_DOMAIN=$(echo ${{ needs.build-and-preview.outputs.deployUrl }} | sed 's|https://||')
           UPDATED_DOMAINS="${CURRENT_DOMAINS},${NEW_DOMAIN}"
           gcloud recaptcha keys update ${{ secrets.FIREBASE_APPCHECK_SITE_KEY_GSE_ROAR_ADMIN }} --domains="$UPDATED_DOMAINS" --web
 
@@ -122,6 +123,6 @@ jobs:
         run: |
           CURRENT_DOMAINS=$(gcloud recaptcha keys describe ${{ secrets.FIREBASE_APPCHECK_SITE_KEY_GSE_ROAR_ASSESSMENT }} --project=gse-roar-assessment --format="value(webSettings.allowedDomains)")
           CURRENT_DOMAINS=$(echo $CURRENT_DOMAINS | tr ';' ',')
-          NEW_DOMAIN=${{ needs.build-and-preview.outputs.deployUrl }}
+          NEW_DOMAIN=$(echo ${{ needs.build-and-preview.outputs.deployUrl }} | sed 's|https://||')
           UPDATED_DOMAINS="${CURRENT_DOMAINS},${NEW_DOMAIN}"
           gcloud recaptcha keys update ${{ secrets.FIREBASE_APPCHECK_SITE_KEY_GSE_ROAR_ASSESSMENT }} --domains="$UPDATED_DOMAINS" --web
--- a/.github/workflows/firebase-hosting-preview.yml
+++ b/.github/workflows/firebase-hosting-preview.yml
@@ -44,9 +44,9 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
-      #      # Inject the App Check Debug Token into the .env file at build time
-      #      - name: Create .env file with App Check Debug Token
-      #        run: echo "VITE_APPCHECK_DEBUG_TOKEN=${{ secrets.VITE_APPCHECK_DEBUG_TOKEN }}" >> .env
+      # Inject the App Check Debug Token into the .env file at build time
+      - name: Create .env file with App Check Debug Token
+        run: echo "VITE_APPCHECK_DEBUG_TOKEN=${{ secrets.VITE_APPCHECK_DEBUG_TOKEN }}" >> .env
 
       - name: Build project
         run: npm run build
--- a/.github/workflows/firebase-hosting-preview.yml
+++ b/.github/workflows/firebase-hosting-preview.yml
@@ -41,13 +41,14 @@ jobs:
         with:
           node-version: 'lts/*'
 
-      - name: Install dependencies and build
-        run: |
-          npm ci
-          npm run build
+      - name: Install dependencies
+        run: npm ci
 
       - name: Create .env file with App Check Debug Token
         run: echo "VITE_APPCHECK_DEBUG_TOKEN=${{ secrets.VITE_APPCHECK_DEBUG_TOKEN }}" >> .env
+
+      - name: Build project
+        run: npm run build
 
       - name: Deploy to Firebase Hosting Channel
         id: firebase-deploy
--- a/.github/workflows/firebase-hosting-preview.yml
+++ b/.github/workflows/firebase-hosting-preview.yml
@@ -44,6 +44,10 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      # Inject the App Check Debug Token into the .env file at build time
+      - name: Create .env file with App Check Debug Token
+        run: echo "VITE_APPCHECK_DEBUG_TOKEN=${{ secrets.VITE_APPCHECK_DEBUG_TOKEN }}" >> .env
+
       - name: Build project
         run: npm run build
 
@@ -84,10 +88,10 @@ jobs:
 
       - name: Update App Check Domains for gse-roar-admin
         run: |
-          CURRENT_DOMAINS=$(gcloud recaptcha keys describe ${{ secrets.FIREBASE_APPCHECK_SITE_KEY_GSE_ROAR_ADMIN }} --format="value(webSettings.allowedDomains)")
-          CURRENT_DOMAINS=$(echo $CURRENT_DOMAINS | tr ';' ',')
+          CURRENT_DOMAINS=$(gcloud recaptcha keys describe ${{ secrets.FIREBASE_APPCHECK_SITE_KEY_GSE_ROAR_ADMIN }} --format="value(webSettings.allowedDomains)" | tr ';' '\n')
           NEW_DOMAIN=$(echo ${{ needs.build-and-preview.outputs.deployUrl }} | sed 's|https://||')
-          UPDATED_DOMAINS="${CURRENT_DOMAINS},${NEW_DOMAIN}"
+          ALL_DOMAINS=$(echo -e "${CURRENT_DOMAINS}\n${NEW_DOMAIN}")
+          UPDATED_DOMAINS=$(echo "$ALL_DOMAINS" | sort | uniq | paste -sd ',' -)
           gcloud recaptcha keys update ${{ secrets.FIREBASE_APPCHECK_SITE_KEY_GSE_ROAR_ADMIN }} --domains="$UPDATED_DOMAINS" --web
 
   whitelist-pr-domains-roar-assessment:
@@ -111,8 +115,8 @@ jobs:
 
       - name: Update App Check Domains for gse-roar-assessment
         run: |
-          CURRENT_DOMAINS=$(gcloud recaptcha keys describe ${{ secrets.FIREBASE_APPCHECK_SITE_KEY_GSE_ROAR_ASSESSMENT }} --project=gse-roar-assessment --format="value(webSettings.allowedDomains)")
-          CURRENT_DOMAINS=$(echo $CURRENT_DOMAINS | tr ';' ',')
+          CURRENT_DOMAINS=$(gcloud recaptcha keys describe ${{ secrets.FIREBASE_APPCHECK_SITE_KEY_GSE_ROAR_ASSESSMENT }} --format="value(webSettings.allowedDomains)" | tr ';' '\n')
           NEW_DOMAIN=$(echo ${{ needs.build-and-preview.outputs.deployUrl }} | sed 's|https://||')
-          UPDATED_DOMAINS="${CURRENT_DOMAINS},${NEW_DOMAIN}"
+          ALL_DOMAINS=$(echo -e "${CURRENT_DOMAINS}\n${NEW_DOMAIN}")
+          UPDATED_DOMAINS=$(echo "$ALL_DOMAINS" | sort | uniq | paste -sd ',' -)
           gcloud recaptcha keys update ${{ secrets.FIREBASE_APPCHECK_SITE_KEY_GSE_ROAR_ASSESSMENT }} --domains="$UPDATED_DOMAINS" --web
--- a/.github/workflows/firebase-hosting-preview.yml
+++ b/.github/workflows/firebase-hosting-preview.yml
@@ -105,33 +105,33 @@ jobs:
         UPDATED_DOMAINS="${CURRENT_DOMAINS},${NEW_DOMAIN}"
         gcloud recaptcha keys update ${{ secrets.FIREBASE_APPCHECK_SITE_KEY_GSE_ROAR_ADMIN }} --domains="$UPDATED_DOMAINS" --web
 
-  whitelist-pr-domains-roar-assessment:
-    - name: Whitelist PR Domains on Firebase App Check for gse-roar-assessment
-      runs-on: ubuntu-latest
-      needs: build-and-preview
-
-      steps:
-        - name: Checkout code
-        - uses: actions/checkout@v4
-
-        - name: Authenticate with Google Cloud SDK
-        - uses: google-github-actions/auth@v2
-        - with:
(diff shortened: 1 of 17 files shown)


lucasxsong (approved):

src/config/firebaseRoar.js
maximilianoertel, line 1: What would you think of adding those new env variables to a `.env.example` file, similar to how it's [being introduced here](https://github.com/yeatmanlab/roar-dashboard/blob/477343a970d9039a95ee889e753b3b4594745443/.env.example)? This is usually pretty helpful because it gives everyone a clear template of the necessary environment variables, making setup easier for new developers and more consistent for existing ones when new vars are introduced. 

Something along the lines of: 
```js
// Firebase App Check
// Set per-environment site keys or the debug token for bypassing the checks during local development. 
VITE_GSE_ROAR_ASSESSMENT_DEV_APPCHECK_SITE_KEY=your-appcheck-key
VITE_GSE_ROAR_ADMIN_DEV_APPCHECK_SITE_KEY=your-appcheck-key
VITE_GSE_ROAR_ASSESSMENT_APPCHECK_SITE_KEY=your-appcheck-key
VITE_GSE_ROAR_ASSESSMENT_APPCHECK_SITE_KEY=your-appcheck-key 
VITE_APPCHECK_DEBUG_TOKEN=your-appcheck-debug-token 
```

ksmontville: This is a good idea, thanks!

richford:
> #### ⚠️ GitGuardian has uncovered 4 secrets following the scan of your pull request.
> Please consider investigating the findings and remediating the incidents. Failure to do so may lead to compromising the associated services or software components.
> 
> 🔎 Detected hardcoded secrets in your pull request
> 🛠 Guidelines to remediate hardcoded secrets
> 🦉 [GitGuardian](https://dashboard.gitguardian.com/auth/login/?utm_medium=checkruns&utm_source=github&utm_campaign=cr1) detects secrets in your source code to help developers and security teams secure the modern development process. You are seeing this because you or someone else with access to this repository has authorized GitGuardian to scan your pull request.

False positive: These are not secrets and I've skipped the check.

src/config/firebaseRoar.js
```diff
@@ -30,23 +29,26 @@ if (isEmulated) {
     },
   };
 } else if (useSandbox) {
-  console.log('Using sandbox data from the dev Firebase projects');
+  console.log('Using sandbox data from the dev Firebase projects.');
```
richford, line 32: Could we get rid of this entirely?
```suggestion
```

src/helpers/query/utils.js
```diff
@@ -74,6 +74,17 @@ export const getAxiosInstance = (db = 'admin', unauthenticated = false) => {
...
+  // Add appCheckToken to the headers if it exists in the firekit config
+  const appCheckToken = roarfirekit.value[db]?.appCheckToken;
+
+  if (appCheckToken) {
+    axiosOptions.headers = {
+      ...axiosOptions.headers,
+      'X-Firebase-AppCheck': appCheckToken,
+    };
+  }
+
```
richford, line 87: Should we conditionally add this to the headers? Is there any harm in adding this if `appCheckToken` is `null` or `undefined`?

richford (changes requested):

src/helpers/query/utils.js
ksmontville: Do you mean should we always attach the appCheckToken, even if it's null or undefined?

Otherwise, won't the `if (appCheckToken)` clause take care of this situation? Since that will be falsy if `appCheckToken `is null or undefined?

richford: OMG. Sorry. Please disregard. Apparently, I just can't read.

richford (changes requested):
I don't know why this is happening because you seem to have followed [these instructions](https://firebase.google.com/docs/app-check/web/debug-provider#ci) to the letter, but for some reason the debug token is being console logged. Can you please revoke the debug token until we figure out how to keep it from being console logged?

ksmontville:
> I don't know why this is happening because you seem to have followed [these instructions](https://firebase.google.com/docs/app-check/web/debug-provider#ci) to the letter, but for some reason the debug token is being console logged. Can you please revoke the debug token until we figure out how to keep it from being console logged?

I looked into this because I noticed the token being console logged, as well.

This is a function of the SDK itself and I don't think that it can be turned off. Perhaps we could suppress all console.log() statements when in production?

maximilianoertel:
@ksmontville @richford curious, what is the concern with the App Check debug token being logged to the console by the Firebase SDK? 

As far as I understand the situation, the token is not a secret as it's designed to be used in the frontend and even if it was not logged to console, you could always extract it from the source code or network request headers.

This also shouldn't be an issue in production, since the debug token is designed to be used in non-production environments?

maximilianoertel:
Wasn't thinking straight yesterday, sorry – and thanks for the explanation @ksmontville! 

However, one point remains: the debug token is only supposed to be used locally and in the CI environments, so we don't need to worry about it being logged in staging or production? Of course, it's not ideal that the token is logged in the CI environment as the logs are stored/viewable in Cypress. However, the people with access to Cypress will also have the token in their local environment for development, so the exposure here is similar to developers being able to add their own debug token to the Firebase console. 

To improve security a little bit, we could have each developer generate their own debug token and add it to the Firebase console, keeping the CI debug token a separate token. Also making it simpler in case the CI or an individual token needs to be cycled.

src/main.js
```diff
@@ -108,7 +109,14 @@ app.use(VueRecaptchaPlugin, {
...
 
-initSentry(app);
+if (process.env.NODE_ENV === 'production') {
+  // Configure logging, and redirect console.log, console.warn, and console.error to the logger
+  const logger = log({ level: 'error' });
+  console.log = logger.info;
+
+  // Initialize Sentry if in a production environment
+  initSentry(app);
+}
```
maximilianoertel, line 119: Even though using a logging library to suppress logging levels is a great idea, I would recommend against that approach as it can make debugging in remote environments so much more complicated. Theoretically the most important logs would be at the error log level, but not always. 

Since this would only be a concern in the CI environment (right?), can we not use `cy.on()` to intercept the console messages and then filter the messages for the debug token before they are being printed to the console? 

Alternatively, if you really want to control console logging, I'd recommend using the [debug](https://www.npmjs.com/package/debug) package. It's a great little tool that would let us use debug keys to control logging output. In other words: only those with the correct debug key in localStorage could see the console outputs. Wanted to write a ticket to recommend this package as it would also help bring some structure in the existing console logs.

src/config/firebaseRoar.js
```diff
@@ -9,7 +9,6 @@ const useSandbox = import.meta.env.VITE_FIREBASE_DATA_SOURCE === 'sandbox';
 const isStaging = import.meta.env.VITE_STAGING_BUILD === 'true';
```
maximilianoertel, line 10: What do you think of adding the following and use `firebaseAppCheckDebugToken` in the `appConfig` declaration? 

```js
const firebaseAppCheckDebugToken = import.meta.env.VITE_APPCHECK_DEBUG_TOKEN || (import.meta.env.VITE_APPCHECK_DEBUG_TOKEN === 'true')
```

That way, we can set `VITE_APPCHECK_DEBUG_TOKEN=true` locally in order to generate our own debug token locally. Without this, the env var gets always interpreted as string and the Firebase SDK treats `true` as a debug token. 

With the expression above, we can control whether we want to generate a new token or set an existing one.

ksmontville:
> Wasn't thinking straight yesterday, sorry – and thanks for the explanation @ksmontville!
> 
> However, one point remains: the debug token would only logged locally and in the CI environments, so we don't need to worry about it being logged in staging or production? Of course, it's not ideal that the token is logged in the CI environment as the logs are stored/viewable in Cypress. However, the people with access to Cypress will also have the token in their local environment for development, so the exposure here is similar to developers being able to add their own debug token to the Firebase console.
> 
> To improve security a little bit, we could have each developer generate their own debug token and add it to the Firebase console, keeping the CI debug token a separate token. Also making it simpler in case the CI or an individual token needs to be cycled.

I think the main issue is with our PR preview links (like this one https://roar-staging--pr743-enh-app-check-xiyhrhmi.web.app/) are logging the debug token, and we consider these links to be a production environment since we use them to let our partners preview any changes. The roar-dashboard is also a public repo, so anyone could theoretically open up a PR link and inspect the console for our debug token.

So it seems that we would still need a "universal" debug token that can be injected at build time and used in the PR links to bypass the App Check domain verification. Otherwise, I do think that it would be a good idea for each developer to have their own debug token.

This whole situation could all be avoided if Firebase would allow us to use a wildcard in its domain verification, then any URL of the form matching of our PR links could just use the actual App Check token rather than the debug token. But Firebas only allows explicit domains in its whitelist, so we have to use a debug token to bypass and allow our PR links (which are variable) to run.

src/main.js
ksmontville: I would agree that it would be a good idea to use the cy.on() hook to filter out console logs that contain the debug token. But the issue is still that our PR links, which use the debug token, are logging the token in the console. These PR links are publicly accessible, since our dashboard repo is public.

src/config/firebaseRoar.js
ksmontville: I like this idea, thanks!

richford:
> This whole situation could all be avoided if Firebase would allow us to use a wildcard in its domain verification, then any URL of the form matching of our PR links could just use the actual App Check token rather than the debug token. But Firebas only allows explicit domains in its whitelist, so we have to use a debug token to bypass and allow our PR links (which are variable) to run.

I wonder if there is a way to use the gcloud CLI to add the PR preview link to the list of trusted domains for that token. Then we could do it in the GitHub action. While, we're at it, we could add it to the list of trusted domains for Firebase auth, then Google SSO would work by default on the PR preview links without human intervention.

ksmontville:
> > This whole situation could all be avoided if Firebase would allow us to use a wildcard in its domain verification, then any URL of the form matching of our PR links could just use the actual App Check token rather than the debug token. But Firebas only allows explicit domains in its whitelist, so we have to use a debug token to bypass and allow our PR links (which are variable) to run.
> 
> I wonder if there is a way to use the gcloud CLI to add the PR preview link to the list of trusted domains for that token. Then we could do it in the GitHub action. While, we're at it, we could add it to the list of trusted domains for Firebase auth, then Google SSO would work by default on the PR preview links without human intervention.

This would be a great idea, since we already grab the PR preview link in the GitHub action. We'll have to look into what kinds of operations the gcloud cli can do with App Check and/or reCAPTCHA and determine if this is possible. 

I'll add a subtask to the card in ClickUp.

Emily-ejag (approved):

Emily-ejag:
Approved, but not sure if GitGuardian is going to follow us later

.github/workflows/firebase-hosting-preview.yml
```diff
@@ -41,21 +41,82 @@ jobs:
...
-      - name: Install dependencies and build
-        run: |
-          npm ci
-          npm run build
+      - name: Install dependencies
+        run: npm ci
+
+      # Inject the App Check Debug Token into the .env file at build time
+      - name: Create .env file with App Check Debug Token
+        run: echo "VITE_APPCHECK_DEBUG_TOKEN=${{ secrets.VITE_APPCHECK_DEBUG_TOKEN }}" >> .env
```
richford, line 49: Is this part still necessary now that we are adding the preview url to the whitelist?

richford (changes requested):

.env.example
```diff
@@ -1,4 +1,25 @@
...
+SUPER_ADMIN_PASSWORD=your-super-admin-password
+SUPER_ADMIN_ID=your-super-admin-firestore-id
+PARTNER_ADMIN_USERNAME=your-partner-admin-username
+PARTNER_ADMIN_PASSWORD=your-partner-admin-password
+PARTNER_ADMIN_ID=your-partner-admin-firestore-id
+PARTICIPANT_USERNAME=your-participant-username
+PARTICIPANT_PASSWORD=your-participant-password
+PARTICIPANT_EMAIL=your-participant-email
+PARTICIPANT_EMAIL_PASSWORD=your-participant-email-password
+
```
maximilianoertel:
```suggestion
```

.github/workflows/firebase-hosting-preview.yml
```diff
@@ -41,21 +41,82 @@ jobs:
...
+        with:
+          project_id: gse-roar-admin
+
+      - name: Update App Check Domains for gse-roar-admin
+        run: |
+          CURRENT_DOMAINS=$(gcloud recaptcha keys describe ${{ secrets.FIREBASE_APPCHECK_SITE_KEY_GSE_ROAR_ADMIN }} --format="value(webSettings.allowedDomains)" | tr ';' '\n')
+          NEW_DOMAIN=$(echo ${{ needs.build-and-preview.outputs.deployUrl }} | sed 's|https://||')
+          ALL_DOMAINS=$(echo -e "${CURRENT_DOMAINS}\n${NEW_DOMAIN}")
+          UPDATED_DOMAINS=$(echo "$ALL_DOMAINS" | sort | uniq | paste -sd ',' -)
+          gcloud recaptcha keys update ${{ secrets.FIREBASE_APPCHECK_SITE_KEY_GSE_ROAR_ADMIN }} --domains="$UPDATED_DOMAINS" --web
```
maximilianoertel, line 95: Not a requirement per-se but this would be a great candidate for a reusable action to avoid the duplication between `whitelist-pr-domains-roar-admin` and `whitelist-pr-domains-roar-assessment`

src/config/firebaseRoar.js
```diff
@@ -8,8 +8,15 @@ const isEmulated = import.meta.env.VITE_FIREBASE_EMULATOR === 'true';
 const useSandbox = import.meta.env.VITE_FIREBASE_DATA_SOURCE === 'sandbox';
 const isStaging = import.meta.env.VITE_STAGING_BUILD === 'true';
 
+function setDebugToken(config) {
+  const debugToken = import.meta.env.VITE_APPCHECK_DEBUG_TOKEN;
```
maximilianoertel, line 12: Do we want to apply [this suggestion](https://github.com/yeatmanlab/roar-dashboard/pull/743/files#r1723581902) here to allow engineers to generate their own debug token instead of passing a single token around? 
```suggestion
  const debugToken = import.meta.env.VITE_APPCHECK_DEBUG_TOKEN || (import.meta.env.VITE_APPCHECK_DEBUG_TOKEN === 'true')

```

src/firebaseInit.js
```diff
@@ -13,8 +13,12 @@ export async function initNewFirekit() {
       db: false,
       functions: false,
     },
-
     verboseLogging: import.meta.env.MODE === 'LEVANTE' ? false : true,
+
+    // The site key is used for app check token verification
+    // The debug token is used to bypass the app check token requirement for local development and for PR deployments
+    siteKey: roarConfig.siteKey,
+    debugToken: roarConfig.debugToken,
```
maximilianoertel, line 21: Haven't tested, but this might cause problems when the debugToken is not set in the config, right? 

```suggestion
    debugToken: roarConfig?.debugToken,
```

maximilianoertel (changes requested):
Great solution leveraging the gcloud CLI to whitelist the PR URLs! 

To avoid having an ever-growing list of whitelisted domains, should we add a workflow that also removes the previously whitelisted URL upon PR merging/closing? Without having tested this, I believe this can be achieved by leveraging the GitHub Environment API

.env.example
ksmontville: This is a good idea. I will make a note of it on ClickUp and we will add it to the tech-debt pile.

.github/workflows/firebase-hosting-preview.yml
ksmontville: I also thought about making the whitelisting job its own workflow, but I was not sure how to capture the deployment link output from the Build and Deploy job and use it in a different workflow. If that's even possible?

src/firebaseInit.js
ksmontville: Good catch, added.

src/config/firebaseRoar.js
ksmontville: Added.

.env.example
maximilianoertel: Sorry, should have been clearer – those no longer need to be added here / should be removed as they are available in `.env.example.test` that was introduced in PR #767

richford (approved):

src/config/firebaseRoar.js
ksmontville: After some testing, this line seems to cause our Cypress tests to fail. I think that it is generating a new debug token each time, and that token won't be valid until it is added to the Firestore app check console. I am going to leave this line out for now and investigate further.

.github/workflows/firebase-hosting-preview.yml
maximilianoertel: This should be possible by extending the existing workflows to create deployments and register environments. For each PR environment, the workflow would automatically create a new environment with the corresponding PR URL. 

The separate workflow would listen to `deployment` events and then identify the URL from that deployment. 

Of course this would need to be extended further to deactivate environments and subsequently remove the the URL from the whitelist

src/config/firebaseRoar.js
maximilianoertel: Odd – when I tried it, it was only generating new tokens when `VITE_APPCHECK_DEBUG_TOKEN` was set to `true`

maximilianoertel (approved):
🚀

PR #751, opened 2024-08-20
Fix `<HomeSelector>` race condition

maximilianoertel:
## Proposed changes
This PR fixes a race condition in the `<HomeSelector>` that would cause the user type check `isAdmin` to switch back and forth between `true` and `false` during page load, effectively causing unwanted renders and side effects on the homepage. This issue seems to only be reproducible after upgrading to the latest `@tanstack/vue-query` version as done in the base branch.

With the proposed changes, the improved loading of the homepage provides a smoother user experience by preventing the flashing of unwanted components and slightly better performance due the asynchronously import of the homepage components. 

#### Before
https://github.com/user-attachments/assets/85445c3c-23c4-4528-b7ad-2f35f2b5a57a

#### After

https://github.com/user-attachments/assets/76068704-a11c-4243-9d78-67f5bc896ffa

> [!NOTE]
> The empty-state placeholder is, unfortunately, still flashing in the updated version. This is due to the nature of `isLoading` initially evaluating to `false` when a TanStack query is initially disabled. This is currently being addressed and will be fixed separately. Additionally, we should address the navbar loading delay as reported in: yeatmanlab/roar/issues/322

## Types of changes
- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist

n/a

## Further comments
n/a


ksmontville (approved):
This looks great!

Do you think that it makes sense to have `SUPER_ADMIN `as a third userType to return? Based on the presence of the `super_admin` key in the user document.

We do have some additional checks and renders throughout the code base that rely on differentiating between an `ADMIN` and a `SUPER_ADMIN`. Would adding that distinction in this PR reduce complexity down the road?

I also noticed the initial "flashing" of the empty state, when administrations are initialized but not yet loaded. Hopefully we can address this, as well.

This also happens when switching from a student account to an admin account, and vice versa. The navbar seems to store the previous state of the user and displays a student username on the admin account, and the admin username on the student account. I suspect this is related to the navbar PR that you are working on.

image

Just some thoughts, otherwise I approve.

richford (approved):

maximilianoertel:
Thanks for the review @ksmontville! 

100% with you on extending the `useUserType` composable, planing on [using it here](https://github.com/yeatmanlab/roar-dashboard/blob/24aa45f1845c424296512a81584e9665b8c052f9/src/composables/queries/useDistrictSchoolsQuery.js#L24-L27). 

> This also happens when switching from a student account to an admin account, and vice versa. The navbar seems to store the previous state of the user and displays a student username on the admin account, and the admin username on the student account. I suspect this is related to the navbar PR that you are working on.
Regarding the above, you could you please confirm: when you mean "switching between student and admin account", you mean you log out, log in again and the name is incorrect in the navbar, correct? If so that is currently expected as I still need to implement query invalidation on the main PR so that should be addressed/fixed shortly

PR #755, opened 2024-08-21
Export Org Users

Emily-ejag:
## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist

### DIFF
--- a/src/components/ListOrgs.vue
+++ b/src/components/ListOrgs.vue
@@ -118,6 +118,7 @@ import { orgFetcher, orgFetchAll, orgPageFetcher } from '@/helpers/query/orgs';
 import { orderByDefault, exportCsv, fetchDocById } from '@/helpers/query/utils';
 import { fetchUsersByOrg, countUsersByOrg } from '@/helpers/query/users';
 import { ref, computed, onMounted, watch } from 'vue';
+import * as Sentry from '@sentry/vue';
 import { storeToRefs } from 'pinia';
 import { useQuery } from '@tanstack/vue-query';
 import { useAuthStore } from '@/store/auth';
@@ -278,18 +279,18 @@ const exportAll = async () => {
 
 /**
  * Exports users of a given organization type to a CSV file.
- * 
+ *
  * @NOTE In order to avoid overly large exports, the function will allow exports up to a predefined limit (currently
  * 10,000 records). To avoid running a large and potentially unecessary query, we first run an aggregation query to
  * verify that the export is within the limit.
- * 
+ *
  * @TODO Replace this logic with a server driven export, for example a cloud function that generate a download link for
  * the user, effectively allowing complete and large exports.
- * 
+ *
  * @param {Object} orgType - The organization type object.
  * @param {string} orgType.id - The ID of the organization type.
  * @param {string} orgType.name - The name of the organization type.
- * 
+ *
  * @returns {Promise<void>} - A promise that resolves when the export is complete.
  */
 const exportOrgUsers = async (orgType) => {
@@ -353,6 +354,7 @@ const exportOrgUsers = async (orgType) => {
       detail: error.message,
       life: 3000,
     });
+    Sentry.captureException(error);
   }
 };
 
--- a/src/components/ListOrgs.vue
+++ b/src/components/ListOrgs.vue
@@ -116,7 +116,7 @@
 <script setup>
 import { orgFetcher, orgFetchAll, orgPageFetcher } from '@/helpers/query/orgs';
 import { orderByDefault, exportCsv, fetchDocById } from '@/helpers/query/utils';
-import { fetchUsersByOrg } from '@/helpers/query/users';
+import { fetchUsersByOrg, countUsersByOrg } from '@/helpers/query/users';
 import { ref, computed, onMounted, watch } from 'vue';
 import { storeToRefs } from 'pinia';
 import { useQuery } from '@tanstack/vue-query';
@@ -125,6 +125,7 @@ import { useToast } from 'primevue/usetoast';
 import _get from 'lodash/get';
 import _head from 'lodash/head';
 import _kebabCase from 'lodash/kebabCase';
+import { CSV_EXPORT_MAX_RECORD_COUNT } from '@/constants/csvExport';
 
 const initialized = ref(false);
 const orgsQueryKeyIndex = ref(0);
@@ -277,10 +278,10 @@ const exportAll = async () => {
 
 const exportOrgUsers = async (orgType) => {
   try {
-    // Fetch the first page of users to estimate total count
-    const users = await fetchUsersByOrg(activeOrgType.value, orgType.id, ref(10001), ref(0), orderBy);
+    // First, count the users
+    const userCount = await countUsersByOrg(activeOrgType.value, orgType.id, orderBy);
 
-    if (!users || users.length === 0) {
+    if (userCount === 0) {
       toast.add({
         severity: 'error',
         summary: 'Export Failed',
@@ -290,7 +291,7 @@ const exportOrgUsers = async (orgType) => {
       return;
     }
 
-    if (users.length >= 10000) {
+    if (userCount >= CSV_EXPORT_MAX_RECORD_COUNT) {
       toast.add({
         severity: 'error',
         summary: 'Export Failed',
@@ -299,6 +300,9 @@ const exportOrgUsers = async (orgType) => {
       });
       return;
     }
+
+    // Fetch the users if the count is within acceptable limits
+    const users = await fetchUsersByOrg(activeOrgType.value, orgType.id, userCount, ref(0), orderBy);
 
     const computedExportData = users.map((user) => ({
       Username: _get(user, 'username'),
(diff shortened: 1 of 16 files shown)


maximilianoertel (commented):
What kind of data size are talking here for the export? Since this is just for the list of users for a given organisation, I assume it's not a huge dataset? 

I'm asking because the query seems to set a user limit of 1 million records (which is quite a lot to fetch on a single request!), and it might be wiser to implement this server-side. As you're likely aware, this sort of processing client-side can be quite memory intensive and depending on the user, this might become problematic. 

Not familiar enough with the project yet to know whether this here would be an issue (based on the actual data sets we're talking about), but just wanted to flag it in case.

src/components/ListOrgs.vue
```diff
@@ -272,6 +275,48 @@ const exportAll = async () => {
...
+  } catch (error) {
+    toast.add({
+      severity: 'error',
+      summary: 'Export Failed',
+      detail: error.message,
+      life: 3000,
+    });
+  }
+};
+
```
ksmontville, line 319:
```suggestion
const exportOrgUsers = async (orgType) => {
  try {
    // Fetch the first page of users to estimate total count
    const users = await fetchUsersByOrg(activeOrgType.value, orgType.id, ref(10001), ref(0), orderBy);

    if (!users || users.length === 0) {
      toast.add({
        severity: 'error',
        summary: 'Export Failed',
        detail: 'No users found for the organization.',
        life: 3000,
      });
      return;
    }

    if (users.length >= 10000) {
      toast.add({
        severity: 'error',
        summary: 'Export Failed',
        detail: 'Too many users to export. Please filter the users by selecting a smaller org type.',
        life: 3000,
      });
      return;
    }

    const computedExportData = users.map((user) => ({
      "Username": _get(user, 'username', "(No Username Found)"),
      "Email": _get(user, 'email', "(No Email Found)"),
      "First Name": _get(user, 'name.first', "(No First Name Found)"),
      "Last Name": _get(user, 'name.last', "(No Last Name Found)"),
      "Grade": _get(user, 'studentData.grade', "(No Grade Found)"),
      "Gender": _get(user, 'studentData.gender', "(No Gender Found)"),
      "Date Of Birth": _get(user, 'studentData.dob', "(No Date of Birth Found)"),
      "User Type": _get(user, 'userType', "(No User Type Found)"),
    }));

    // ex. cypress-test-district-users-export.csv
    exportCsv(computedExportData, `${_kebabCase(orgType.name)}-users-export.csv`);

    toast.add({
      severity: 'success',
      summary: 'Export Successful',
      detail: 'Users have been exported successfully!',
      life: 3000,
    });
  } catch (error) {
    toast.add({
      severity: 'error',
      summary: 'Export Failed',
      detail: error.message,
      life: 3000,
    });
  }
};
```
This is looking good! I have a couple of suggestions:

1. `orgType` is an object with properties that we can access
2. Use toasts to inform the user about the state of the export, rather than using `throw` as flow-control
3. Using strings as the column headers allows the .csv export to have spaces in the column names; this is not super critical and probably depends on user preference either way
4. Use the `orgType` properties to make a more descriptive .csv export filename
5. In the future, we might even have a checkbox or button that DOES allow the user to bypass the check and export more than 10,000 users if their browser can handle the computation.

ksmontville (changes requested):

src/components/ListOrgs.vue
Emily-ejag: Thank you for your suggestions, let me ask P&R if they would like the exporting to contain a default value, or nothing

ksmontville (approved):

src/components/ListOrgs.vue
```diff
@@ -272,6 +275,60 @@ const exportAll = async () => {
   exportCsv(exportData, `roar-${activeOrgType.value}.csv`);
 };
 
+const exportOrgUsers = async (orgType) => {
+  try {
+    // Fetch the first page of users to estimate total count
+    const users = await fetchUsersByOrg(activeOrgType.value, orgType.id, ref(10001), ref(0), orderBy);
```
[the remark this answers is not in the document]
richford: Agreed, let's use an aggregation query first. `countUsersByOrg` should do the trick. That way we can avoid the query if it turns out there are too many.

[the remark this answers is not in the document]
richford: I agree that it would be good to put our constants in `@/constants/csvExport`. Max's example code there is somewhat obviated by the request to use `countUsersByOrg`. But still, putting the constant in that file and then using the constant for comparing against the count is a good idea.

src/components/ListOrgs.vue
```diff
@@ -272,6 +275,60 @@ const exportAll = async () => {
...
+    const computedExportData = users.map((user) => ({
+      Username: _get(user, 'username'),
+      Email: _get(user, 'email'),
+      FirstName: _get(user, 'name.first'),
+      LastName: _get(user, 'name.last'),
+      Grade: _get(user, 'studentData.grade'),
+      Gender: _get(user, 'studentData.gender'),
+      DateOfBirth: _get(user, 'studentData.dob'),
+      UserType: _get(user, 'userType'),
+    }));
```
richford, line 312: Can we export all of the available data that we might have? Some of it will be null. But we should include it if we can. E.g., `ell_status`, `frl_status`, etc.

richford (changes requested):

src/components/ListOrgs.vue
Emily-ejag: done!

src/components/ListOrgs.vue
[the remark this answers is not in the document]
Emily-ejag: thank you! done!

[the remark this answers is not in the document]
Emily-ejag: Yes, thanks @maximilianoertel ! done!

src/components/ListOrgs.vue
```diff
@@ -272,6 +276,70 @@ const exportAll = async () => {
   exportCsv(exportData, `roar-${activeOrgType.value}.csv`);
 };
 
+const exportOrgUsers = async (orgType) => {
```
maximilianoertel, line 279: Not going to block the PR for this, but two quick comments:

* This is the type of function that would best live outside of the component in form of service. But this would require some additional tweaks. Let me know if this is something you want to explore and I'd be happy to elaborate. 

* Not sure if this is something the team wants to do, but personally I'd add a JSDoc comment here, mostly to document the approach and decision here. Something along the lines of: 

```suggestion
/**
 * Exports users of a given organization type to a CSV file.
 * 
 * @NOTE In order to avoid overly large exports, the function will allow exports up to a predefined limit (currently
 * 10,000 records). To avoid running a large and potentially unecessary query, we first run an aggregation query to
 * verify that the export is within the limit.
 * 
 * @TODO Replace this logic with a server driven export, for example a cloud function that generate a download link for
 * the user, effectively allowing complete and large exports.
 * 
 * @param {Object} orgType - The organization type object.
 * @param {string} orgType.id - The ID of the organization type.
 * @param {string} orgType.name - The name of the organization type.
 * 
 * @returns {Promise<void>} - A promise that resolves when the export is complete.
 */
const exportOrgUsers = async (orgType) => {
```

src/components/ListOrgs.vue
```diff
@@ -272,6 +276,70 @@ const exportAll = async () => {
...
+      life: 3000,
+    });
+  } catch (error) {
+    toast.add({
+      severity: 'error',
+      summary: 'Export Failed',
+      detail: error.message,
+      life: 3000,
+    });
+  }
```
maximilianoertel, line 340: This is a dangerous one as any error that arises from within the `try` block gets effectively swallowed. Whilst the user gets to see a Toast, we very likely won't have any errors in Sentry about this. 

Again, not holding the PR back for this simply because this isn't the only place where this is happening and I'll write a ticket for it. But, generally speaking, it would be good to manually capture any exception here to ensure Sentry logs it. 

```js
import * as Sentry from '@sentry/vue'

try { 
  // …
} catch {
  toast.add({})
  Sentry.captureException(error)
}
```

src/components/ListOrgs.vue
```diff
@@ -272,6 +276,70 @@ const exportAll = async () => {
...
+      toast.add({
+        severity: 'error',
+        summary: 'Export Failed',
+        detail: 'No users found for the organization.',
+        life: 3000,
+      });
+      return;
+    }
+
+    if (userCount >= CSV_EXPORT_MAX_RECORD_COUNT) {
```
maximilianoertel, line 294:
```suggestion
    if (userCount > CSV_EXPORT_MAX_RECORD_COUNT) {
```

A tiny detail, not important, just might be easier to state that we allow exports up to 10.000 users, not up to 9.999 users

maximilianoertel (approved):
Thanks for the changes @Emily-ejag – looks great! 

I left a couple of comments, but approving the PR as I don't want to hold back the PR for this – up to you to decide whether the feedback should be incorporated or not.

src/components/ListOrgs.vue
Emily-ejag: Ohh this is cool, let me add the sentryCapture

maximilianoertel (approved):
🚀

richford (approved):

richford:
Approved. There is a merge conflict in the roar-firebase-function submodule. Can you either rebase or update the submodule to the latest upstream/main?

PR #765, opened 2024-08-27
Export combination of Score and Progress Report

Emily-ejag:
## Proposed changes

This PR export score reports into a single function that handles exporting for selected rows, the entire report, and combined reports.

**Key Updates:**
Unified Export Function: The `export-selected`,` export-whole`, and `export-combined` functionalities are now managed by a single function, simplifying the export logic and reducing code duplication.

**Steps to Review:**
- Navigate to the Score Report of an admin.
- Select "Export Combined Reports" to generate the CSV.
- Verify that the exported report is formatted correctly and includes the expected data for both selected rows and the entire report as applicable.

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist


Emily-ejag:
Things I would like to do: once this is working, I would like to do a function outside the components to avoid all the controlling on the tasks scores

Emily-ejag:
> I noticed that for "CA BRSSD Spring 2024" the combined export did not include progress columns for Morphology or Picture Vocab. Could you take a look and see why it is not catching those two assessments?

this should be fixed now

src/pages/ScoreReport.vue
```diff
@@ -128,8 +134,8 @@
             :lazy-pre-sorting="orderBy"
             data-cy="roar-data-table"
             @reset-filters="resetFilters"
-            @export-all="exportAll"
-            @export-selected="exportSelected"
+            @export-all="exportData({ includeProgress: false })"
```
maximilianoertel, line 137: `includeProgress` defaults to `false` so we can drop this here? 

```suggestion
            @export-all="exportData"
```

src/pages/ScoreReport.vue
```diff
@@ -128,8 +134,8 @@
...
+            @export-selected="exportData({ selectedRows: $event, includeProgress: false })"
```
maximilianoertel, line 138:
```suggestion
            @export-selected="exportData({ selectedRows: $event })"
```

src/pages/ScoreReport.vue
```diff
@@ -883,150 +911,197 @@ const viewOptions = ref([
...
+ * score types (percent correct, raw scores, standard scores, etc.) based on task configuration.
+ * @param {boolean} [includeProgress=false] - Flag indicating whether to include task progress data in the export. If true,
+ * progress data will be fetched and appended for each task per user.
+ *
+ * @returns {Object[]} - The formatted data array, where each object represents a user and their associated scores.
+ * This data is ready for CSV export and optionally includes progress information.
+ *
+ * @NOTE This function also checks for the user's role (e.g., super admin) to determine additional fields (such as PID),
+ * handles task-specific score presentation based on configuration, and validates task reliability using engagement flags.
+ * If scores are found unreliable, the reliability reason is included. If the task is incomplete, it is marked as such.
```
maximilianoertel, line 933: A tiny detail, but I would recommend grouping the function description and keeping the `param` and `returns` to follow the standard JSDoc format. I would keep `@NOTE` for additional highlighted but non-critical information or tips. The existing comments really seem to be rather describing the functionality and hence do fine as regular description.

Also, I believe we might want to change the return type to be `Array<Object>`? 

```js
/**
 * Creates and formats the data for exporting user, score, and optionally, progress information to a CSV file.
 *
 * This function generates a structured dataset based on user and score data, with optional inclusion of progress
 * data. It ensures that the data is organized appropriately for export, including task-specific formatting an
 * reliability checks. If progress data is included, it appends relevant progress information per task.
 *
 * This function also checks for the user's role (e.g., super admin) to determine additional fields (such as PID),
 * handles task-specific score presentation based on configuration, and validates task reliability using engagement
 * flags. If scores are found unreliable, the reliability reason is included. If the task is incomplete, it is marked as
 * such.
 *
 * @param {Object[]} rows - The array of user data and associated scores.
 * @param {Object} rows[].user - The user object containing user details such as username, email, first name, and last
 * name. 
 * @param {Object} rows[].scores - The scores object containing task-related score data for the user. It supports
 * different score types (percent correct, raw scores, standard scores, etc.) based on task configuration. 
 * @param {boolean} [includeProgress=false] - Flag indicating whether to include task progress data in the export. If
 * true, progress data will be fetched and appended for each task per user. 
 *
 * @returns {Array<Object>} - The formatted data array, where each object represents a user and their associated scores.
 * This data is ready for CSV export and optionally includes progress information.
 */

File: src/pages/ScoreReport.vue [hunk 4]
```diff
@@ -883,150 +911,197 @@ const viewOptions = ref([
...
+ */
+
+const createExportData = ({ rows, includeProgress = false }) => {
+  const computedExportData = _map(rows, ({ user, scores }) => {
     let tableRow = {
       Username: _get(user, 'username'),
+      Email: _get(user, 'email'), // This will only be used when exporting all rows
       First: _get(user, 'firstName'),
       Last: _get(user, 'lastName'),
       Grade: _get(user, 'grade'),
```
maximilianoertel, line 943: No need for `lodash`, we can achieve the exact same using optional chaining:  

```suggestion
      Username: user?.username,
      Email: user?.email, // This will only be used when exporting all rows
      First: user?.firstName,
      Last: user?.lastName,
      Grade: user?.grade,
```

src/pages/ScoreReport.vue
```diff
@@ -883,150 +911,197 @@ const viewOptions = ref([
...
     };
+
     if (authStore.isUserSuperAdmin) {
       tableRow['PID'] = _get(user, 'assessmentPid');
```
maximilianoertel, line 947:
```suggestion
      tableRow['PID'] = user?.assessmentPid
```

src/pages/ScoreReport.vue
```diff
@@ -883,150 +911,197 @@ const viewOptions = ref([
...
     }
+
     if (props.orgType === 'district') {
       tableRow['School'] = _get(user, 'schoolName');
```
maximilianoertel, line 951:
```suggestion
      tableRow['School'] = user?.schoolName
```

src/pages/ScoreReport.vue
```diff
@@ -598,6 +604,28 @@ const getScoresAndSupportFromAssessment = ({
...
+  if (!assignmentData.value) return [];
+  return assignmentData.value.map(({ assignment }) => {
+    const progress = assignment.assessments.reduce((acc, assessment) => {
+      const status = assessment.optional
+        ? 'optional'
+        : assessment.completedOn !== undefined
+        ? 'completed'
+        : assessment.startedOn !== undefined
+        ? 'started'
+        : 'assigned';
```
maximilianoertel, line 617: We should be able to simplify this by removing the `!== undefined` checks, as the values themselves will be evaluated as truthy or falsy.

```suggestion
      const status = assessment.optional
        ? 'optional'
        : assessment.completedOn
        ? 'completed'
        : assessment.startedOn
        ? 'started'
        : 'assigned';
```

src/pages/ScoreReport.vue
```diff
@@ -598,6 +604,28 @@ const getScoresAndSupportFromAssessment = ({
...
+
+      acc[assessment.taskId] = { value: status };
+      return acc;
+    }, {});
+    return {
+      userPid: assignment.userData?.assessmentPid, // Assuming user contains a `username` property
+      progress,
+    };
```
maximilianoertel, line 625: Totally optional, but we could consider destructuring the `assignment` object to make the code more readable: 

```suggestion
  return assignmentData.value.map(({ assignment: { assessments, userData } }) => {
    const progress = assessments.reduce((acc, assessment) => {
      const status = assessment.optional
        ? 'optional'
        : assessment.completedOn
        ? 'completed'
        : assessment.startedOn
        ? 'started'
        : 'assigned';

      acc[assessment.taskId] = { value: status };
      return acc;
    }, {});
    return {
      userPid: userData?.assessmentPid
      progress,
    };
```

src/pages/ScoreReport.vue
```diff
@@ -883,150 +911,197 @@ const viewOptions = ref([
...
               includedValidityFlags[taskId].includes(flag),
             );
-            if (filteredFlags.length === 0) {
-              tableRow[`${tasksDictionary.value[taskId]?.publicName ?? taskId} - Reliability`] = 'Unreliable';
-            } else {
-              const engagementFlagString = 'Unreliable: ' + filteredFlags.map((key) => _lowerCase(key)).join(', ');
-              tableRow[`${tasksDictionary.value[taskId]?.publicName ?? taskId} - Reliability`] = engagementFlagString;
-            }
+            tableRow[`${taskName} - Reliability`] =
+              filteredFlags.length === 0 ? 'Unreliable' : `Unreliable: ${filteredFlags.map(_lowerCase).join(', ')}`;
```
maximilianoertel, line 988: No need for `lodash` here either: 

```suggestion
              filteredFlags.length === 0 ? 'Unreliable' : `Unreliable: ${filteredFlags.map(flag => flag.toLowerCase()).join(', ')}`
```

src/pages/ScoreReport.vue
```diff
@@ -883,150 +911,197 @@ const viewOptions = ref([
...
           } else {
-            const engagementFlagString = 'Unreliable: ' + engagementFlags.map((key) => _lowerCase(key)).join(', ');
-            tableRow[`${tasksDictionary.value[taskId]?.publicName ?? taskId} - Reliability`] = engagementFlagString;
+            tableRow[`${taskName} - Reliability`] = `Unreliable: ${engagementFlags.map(_lowerCase).join(', ')}`;
```
maximilianoertel, line 990: Same here: 

```suggestion
            tableRow[`${taskName} - Reliability`] = `Unreliable: ${engagementFlags.map(flag => flag.toLowerCase()).join(', ')}`
```

src/pages/ScoreReport.vue
```diff
@@ -883,150 +911,197 @@ const viewOptions = ref([
...
-            );
-            if (filteredFlags.length === 0) {
-              tableRow[`${tasksDictionary.value[taskId]?.publicName ?? taskId} - Reliability`] = 'Unreliable';
-            } else {
-              const engagementFlagString = 'Unreliable: ' + filteredFlags.map((key) => _lowerCase(key)).join(', ');
-              tableRow[`${tasksDictionary.value[taskId]?.publicName ?? taskId} - Reliability`] = engagementFlagString;
+      // Add progress immediately after reliability if includeProgress is true
+      if (includeProgress) {
+        const progressRow = computedProgressData.value.find(
+          (progress) => progress.userPid === _get(user, 'assessmentPid'),
```
maximilianoertel, line 1002:
```suggestion
          (progress) => progress.userPid === user?.assessmentPid,
```

src/pages/ScoreReport.vue
```diff
@@ -883,150 +911,197 @@ const viewOptions = ref([
...
+          // If no progressRow is found, mark all scores as "not assigned"
+          scoreReportColumns.value.forEach((column) => {
+            const { field, header: taskName } = column; // Use taskName from the column header
+
+            // Ensure field is defined and is a string before calling startsWith
+            if (field && typeof field === 'string' && field.startsWith('scores')) {
+              tableRow[`${taskName} - Progress`] = 'not assigned';
+            }
+          });
         }
```
maximilianoertel, line 1031: For such length `if`/`else` blocks, I would recommend using the early return pattern as it usually makes the code much more readable and maintainable. For example here: 

```js
// If no progressRow is found, mark all scores as "not assigned"
if (!progressRow) {
  scoreReportColumns.value.forEach((column) => {
    const { field, header: taskName } = column; // Use taskName from the column header

    // Ensure field is defined and is a string before calling startsWith
    if (typeof field === 'string' && field.startsWith('scores')) {
      tableRow[`${taskName} - Progress`] = 'not assigned';
    }
  });

  return 
}

scoreReportColumns.value.forEach((column) => {
  const { field, header: taskName } = column; // Use taskName from the column header

  // Ensure field is defined and is a string before calling startsWith
  if (field && typeof field === 'string' && field.startsWith('scores')) {
    // Extract taskId (e.g., "swr", "sre", etc.)
    const scoreKey = field.split('.').slice(-2, -1)[0];

    // Check if taskId exists in progressRow.progress
    tableRow[`${taskName} - Progress`] = progressRow.progress[scoreKey] 
      ? progressRow.progress[scoreKey].value
      : 'not assigned';
  }
});
```

src/pages/ScoreReport.vue
```diff
@@ -883,150 +911,197 @@ const viewOptions = ref([
...
+        const progressRow = computedProgressData.value.find(
+          (progress) => progress.userPid === _get(user, 'assessmentPid'),
+        );
+
+        if (progressRow) {
+          scoreReportColumns.value.forEach((column) => {
+            const { field, header: taskName } = column; // Use taskName from the column header
+
+            // Ensure field is defined and is a string before calling startsWith
+            if (field && typeof field === 'string' && field.startsWith('scores')) {
```
maximilianoertel, line 1010: This can be simplified if I'm not mistaken: 

```suggestion
            if (typeof field === 'string' && field.startsWith('scores')) {
```

maximilianoertel (changes requested):
Great to see this enhancement! Like you noted in the comments yourself, it would be really good to move this logic out of the component/page and into a simple helper for which we can also write a couple of unit tests for. 

Other than that, a couple of comments below, nothing major but just a few areas where I think we can simplify the code a little bit:

src/pages/ScoreReport.vue
Emily-ejag: Yes, that was from the previous implementation, I forgot to delete.

src/pages/ScoreReport.vue
Emily-ejag: Let me double check this first, since I have used `toLowerCase()` before and it has not printed it as I wanted

src/pages/ScoreReport.vue
Emily-ejag: same here

src/pages/ScoreReport.vue
Emily-ejag: I test it and without _lodash it prints like this: `Unreliable: responsetimetoofast` With _lodash it prints like this: `Unreliable: response time too fast`. So I think for now I will keep the _lodash

maximilianoertel: Oh, yes, that's expected as lodash's `toLowerCase` is case method that inserts spaces – apologies for overlooking that, I expected the flags to be a different format.

```js 
const str = 'ResponseTimeTooFast'

console.log(_.lowerCase(str))  // prints: response time too fast
console.log(_.toLower(str))    // prints: responsetimetoofast
console.log(str.toLowerCase()) // prints: responsetimetoofast
```

maximilianoertel (approved):
🚀

richford (approved):

PR #771, opened 2024-09-03
fix/Terms of Service Popup

ksmontville:
Fix an issue where terms of service status was being checked before userData was loaded, which caused the TOS modal to always appear even if TOS was previously signed.


Emily-ejag (approved):

maximilianoertel (approved):
🚀

PR #785, opened 2024-09-10
Migrate Participant Homepage to composable TanStack queries

maximilianoertel:
## Proposed changes
This PR migrates the `ParticipantHomepage` page to use new composable TanStack queries and mutations. 

## Types of changes
- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist

n/a

## Further comments
n/a

Ref https://github.com/yeatmanlab/roar/issues/318

### DIFF
--- a/src/composables/queries/useSurveyResponsesQuery.js
+++ b/src/composables/queries/useSurveyResponsesQuery.js
@@ -1,4 +1,4 @@
-import { computed } from 'vue';
+import { computed, toValue } from 'vue';
 import { useQuery } from '@tanstack/vue-query';
 import { storeToRefs } from 'pinia';
 import { useAuthStore } from '@/store/auth';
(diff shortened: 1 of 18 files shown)


src/composables/queries/useSurveyResponsesQuery.js
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Survey responses query.
+ *
+ * @param {QueryOptions|undefined} queryOptions – Optional TanStack query options.
+ * @returns {UseQueryResult} The TanStack query result.
+ */
+const useSurveyResponsesQuery = (queryOptions = undefined) => {
+  const authStore = useAuthStore();
+  const { uid } = storeToRefs(authStore);
```
maximilianoertel, line 17: @Zio-4 is this supposed to use the `uid` or `roarUid`? @kellyel recently explained to me that we want to fetch by `roarUid` everywhere due to the Clever implementation, and all queries do - except this one. Is this expected since this for Levante, or should this also be using the `roarUid`?

Emily-ejag: I think it needs to be roarUid, but I will let @kellyel confirm

kellyel: RoarUid should work in this case. 

@Zio-4, does Levante use Clever as a possible SSO provider? If so, roarUid will be correct here. New Clever users that get synced to the platform starting (about) last month will have a different roarUid and adminUid, where roarUid is to be used to grab the userDocument. If you're not using clever at all, roarUid should still be the exact same as uid so changing it to roarUid would not break it. 

I would recommend using roarUid, as it doesn't break anything and it's future-proofing in case Levante ever decides to use Clever or edit their usage of uids in the future, as we have.

Emily-ejag (approved):

richford (approved):

PR #832, opened 2024-09-25
Fix consent modal re-rendering 

maximilianoertel:
## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist

### DIFF
--- a/src/constants/consentTypes.js
+++ b/src/constants/consentTypes.js
@@ -1,0 +1,10 @@
+/**
+ * Consent types
+ *
+ * @constant {Object} CONSENT_TYPES – The types of consent.
+ */
+export const CONSENT_TYPES = Object.freeze({
+  TOS: 'tos',
+  ASSENT: 'assent',
+  ASSENT_ES: 'assent-es',
+});
--- a/src/pages/HomeParticipant.vue
+++ b/src/pages/HomeParticipant.vue
@@ -1,10 +1,25 @@
 <template>
   <div>
-    <div v-if="!noGamesAvailable">
-      <div v-if="isFetching" class="loading-container">
-        <AppSpinner style="margin-bottom: 1rem" />
-        <span>{{ $t('homeParticipant.loadingAssignments') }}</span>
+    <div v-if="isFetching" class="loading-container py-8">
+      <AppSpinner style="margin-bottom: 1rem" />
+      <span>{{ $t('homeParticipant.loadingAssignments') }}</span>
+    </div>
+
+    <div v-else>
+      <div v-if="!hasAssignments">
+        <div class="col-full text-center py-8">
+          <h1>{{ $t('homeParticipant.noAssignments') }}</h1>
+          <p class="text-center">{{ $t('homeParticipant.contactAdministrator') }}</p>
+          <router-link :to="{ name: 'SignOut' }">
+            <PvButton
+              :label="$t('navBar.signOut')"
+              class="no-underline bg-primary border-none border-round p-2 text-white hover:bg-red-900"
+              icon="pi pi-sign-out"
+            />
+          </router-link>
+        </div>
       </div>
+
       <div v-else>
         <h2 v-if="userAdministrations?.length == 1" class="p-float-label dropdown-container">
           {{ userAdministrations.at(0).publicName || userAdministrations.at(0).name }}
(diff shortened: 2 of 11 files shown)


maximilianoertel:
@richford can you please take a look at this one and confirm whether the introduced changes in regards to error handling are okay? In other words: is it correct that a user should not be able to use the dashboard until their consent was given?

src/constants/consentTypes.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Consent types
+ *
+ * @constant {Object} CONSENT_TYPES – The types of consent.
+ */
+export const CONSENT_TYPES = Object.freeze({
+  TOS: 'tos',
+  ASSENT: 'assent',
+  ASSENT_ES: 'assent-es',
```
Emily-ejag, line 9: probably needed for later:
```suggestion
  ASSENT_ES: 'assent-es',
  CONSENT: 'consent',
  CONSENT_ES: 'consent-es'
```

We have a bunch more (see legal collection on the Roar Admin database), but those are the ones more typically used, others are : 

- assent-behaivoral-eye-tracking-es
- assent-behavioral-eye-tracking
- assent-clinic-adobe
- assent-video-audio-eye-tracking
- assent-video-audio-eye-tracking-es
- assent-video-audio-recording
- assent-video-audio-recording-es
- consent-behaivoral-eye-tracking-es
- consent-behavioral-eye-tracking
- consent-clinic-adobe
- consent-video-audio-eye-tracking
- consent-video-audio-eye-tracking-es
- consent-video-audio-recording
- consent-video-audio-recording-es

src/pages/HomeParticipant.vue
```diff
@@ -232,38 +237,35 @@ async function checkConsent() {
...
-    const legalDocs = _get(toRaw(consentStatus), consentDoc.version);
+  if (consentStatus?.[consentDoc.version]) {
+    const legalDocs = consentStatus?.[consentDoc.version];
+
     let found = false;
     let signedBeforeAugFirst = false;
-    let signedAfterAugFirst = false;
 
-    _forEach(legalDocs, (document) => {
+    for (const document of legalDocs) {
```
Emily-ejag, line 250: I just want to understand: Why not using _forEach to optimize for performance by stopping the loop early?

src/constants/consentTypes.js
Emily-ejag: I guess we do not this for now, just the Spanish ones and TOS

src/pages/HomeParticipant.vue
```diff
@@ -121,17 +123,18 @@ const consentType = ref('');
 const consentParams = ref({});
 
 const isLevante = import.meta.env.MODE === 'LEVANTE';
+
+const { mutateAsync: updateConsentStatus } = useUpdateConsentMutation();
```
Emily-ejag, line 127: This is interesting

Emily-ejag (approved):

richford (changes requested):
LGTM. You are correct that we do not want to let users use the site until they have either assented or consented. However, I'm requesting that you add the different consent types that @Emily-ejag pointed out in her comment. Other than that, this is good to go. Thanks for catching this.

src/pages/HomeParticipant.vue
maximilianoertel: It really is – very important here that we use `mutateAsync` and not the usual `mutate` as we need it to throw any errors it encounters in order to catch them in the `<ConsentModal />`

src/constants/consentTypes.js
maximilianoertel: Thanks for providing the list! I noticed there were more consent types but since I didn't touch / see another part of the code that explicitly checked for those, I omitted them (for now) from the constant. However, it definitely does not hurt to have them there, so added all of those as requested by you and @richford.

Only question are the names of the following actually correct or just accidental typos? 
* `assent-behaivoral-eye-tracking-es`
* `consent-behaivoral-eye-tracking-es`

src/pages/HomeParticipant.vue
maximilianoertel: I made the change because, just from experience, `lodash` only comes with benefits in specific cases  – the rest of the time it's an unnecessary dependency with (modern) native JS utilities being able to do the exact same. There are however cases where, whilst doable in native JS, using a `lodash` method is cleaner / improves readability. 

In this case though, for such as simple loop, lodash is a bit overkill in my opinion. You are however completely right, since this loop is actually just intended to find a single element, braking the loop as soon as we have found a result is beneficial. I had overlooked this since until now the breaker was only applied if we found a document signed after the limit date. 

Anyway, here's a little JSPerf example to show the performance differences on a slightly simplified example that runs on a collection of 10 documents: https://jsperf.app/rakoqo/3 

jsperf app_rakoqo_3

Emily-ejag: This explanation is so informative, thank you for sharing :)

src/constants/consentTypes.js
maximilianoertel: Can we correct it in the database(s)?

maximilianoertel: @Emily-ejag @richford just following up – do we want to correct in the db or should I introduce them with a typo in the constants?

richford: Yes, let's correct in the DB. @Emily-ejag can you please do that?

richford: Hmm. That would be effort than I originally thought. Because this issue fixes a bug and is more urgent, @maximilianoertel , can you use the incorrect names for now? That is, can you incorporate the typos in the source code so that the dashboard code is consistent with the database?

Then, we will add another task to fix the typos. But this will involve updating all of the collection names and also all of the user docs for users who have assented/consented to those documents.

richford (changes requested):

src/constants/consentTypes.js
maximilianoertel: Done @richford

richford (approved):

PR #835, opened 2024-09-27
Replace existing sign-out method with query mutation

maximilianoertel:
## Proposed changes

This PR replaces the existing sign-out method with a `useSignOutMutation` that handles actual sign-out, redirect and query invalidation.

At the same time, this PR follows-up with a `TODO` found in the code and removes the currently used `/signout` page in favour of signing the user after they click the sign-out button. 

## Types of changes
- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist

n/a

## Further comments
n/a

Ref https://github.com/yeatmanlab/roar/issues/318


src/composables/mutations/useSignOutMutation.js
```diff
@@ -0,0 +1,45 @@
...
+      // remove it manually from sessionStorage to prevent any issues.
+      authStore.$reset();
+      sessionStorage.removeItem('authStore');
+      sessionStorage.removeItem('gameStore');
+
+      // Redirect to sign-in page.
+      await router.push({ path: APP_ROUTES.SIGN_IN });
+
+      // Clear the query client to remove all cached data.
+      queryClient.clear();
```
richford, line 37: Not a change request but just a question: does ordering matter here? Is there a difference between routing to the sign-in page then clearing the query client and clearing the client before routing to sign in?

If not, would it be better to clear before routing so that we further guarantee that we arrive at sign-in with no cached data?

richford (approved):
Approved but I asked a question about the sign out mutation order of operations.

Emily-ejag (approved):

src/composables/mutations/useSignOutMutation.js
maximilianoertel: Excellent question! In this case, order does usually matter. 

If we were to execute `queryClient.clear()` before routing, the user should see data flash/disappear in front of their eyes for a brief moment before being redirected to the sign-in page. That's why I usually opt to `await` the routing result and only remove cached data from the query client then.  

However, I tested this again and it doesn't really be a concern here – especially since we're already resetting auth store state before the redirect. So I moved the cache removal to happen before the routing.

Should we observe some ugly UX during sign-out, I'll happily go back and improve it.

PR #846, opened 2024-10-03
E2E: Export All Users from an Org

Emily-ejag:
## Proposed changes
This tests tries to:
- Log in,
- Navigate to the organization list,
- Verify the existence of a specific district,
- Export user data for that district into a CSV file.

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist


cypress/e2e/partner-admin/default-tests/exportAllUsersFromOrg.cy.js
```diff
@@ -0,0 +1,30 @@
...
+  beforeEach(() => {
+    cy.login(Cypress.env('partnerAdminUsername'), Cypress.env('partnerAdminPassword'));
+    cy.wait(0.2 * timeout);
+    cy.navigateTo('/');
+    cy.wait(0.2 * timeout);
+    cy.navigateTo(listOrgsUrl, { timeout: timeout });
+  });
+
+  context(`when navigating to the ${Cypress.env('testPartnerDistrictName')} tab`, () => {
+    it(`should see the organization ${Cypress.env('testPartnerDistrictName')}`, () => {
```
maximilianoertel, line 24: The `it` should be descriptive of the actual test, in this case something along the following the lines would be more appropriate: 

```suggestion
    it(`should export ${Cypress.env('testPartnerDistrictName')} organisation users as CSV`, () => {
```

cypress/e2e/partner-admin/default-tests/exportAllUsersFromOrg.cy.js
```diff
@@ -0,0 +1,30 @@
...
+const listOrgsUrl = '/list-orgs';
+
+function checkOrgExists() {
+  cy.get('ul > li', { timeout: timeout }).contains('Districts', { timeout: timeout }).click();
+
+  cy.get('div', { timeout: timeout }).should('contain.text', Cypress.env('testPartnerDistrictName'), {
+    timeout: timeout,
+  });
+  cy.log(`${Cypress.env('testPartnerDistrictName')} exists.`);
+}
```
maximilianoertel, line 12: A couple of things here: 

- Technically no need to make this a separate function as we only use this in one test it would be easier/better to have this directly inside the test.

- Similar to other comments: way too many timeouts. We should remove all the timeouts that were defined here as Cypress already waits 4s by default and if these actions take longer then that's not a problem with Cypress but with the application.

- Instead of using generic selectors like `ul > li` or `div`, we should use `data-testId` attributes (preferred over `data-cy` as we can reuse the `testId` ones in Vitest): https://docs.cypress.io/guides/references/best-practices#Selecting-Elements

cypress/e2e/partner-admin/default-tests/exportAllUsersFromOrg.cy.js
```diff
@@ -0,0 +1,30 @@
...
+    cy.navigateTo('/');
+    cy.wait(0.2 * timeout);
+    cy.navigateTo(listOrgsUrl, { timeout: timeout });
+  });
+
+  context(`when navigating to the ${Cypress.env('testPartnerDistrictName')} tab`, () => {
+    it(`should see the organization ${Cypress.env('testPartnerDistrictName')}`, () => {
+      checkOrgExists();
+      cy.get('button').contains('Export Users').click();
+      cy.readFile(`${Cypress.env('cypressDownloads')}/cypress-test-district-users-export.csv`);
```
maximilianoertel, line 27: We should extend this test to also check the contents of the CSV as this test could still pass but result in an empty or malformed CSV. Here's an example: https://github.com/cypress-io/cypress-example-recipes/blob/master/examples/testing-dom__csv-table/cypress/e2e/spec.cy.js

maximilianoertel (changes requested):

cypress/e2e/partner-admin/default-tests/exportAllUsersFromOrg.cy.js
Emily-ejag: I agree we should test this, but I think it is out of the scope for now, it should be a different unit test

cypress/e2e/partner-admin/default-tests/exportAllUsersFromOrg.cy.js
Emily-ejag: This is true, we try to use data-cy as much as we can, but sometimes we don't get the right DOM element, since PrimeVue injects some html to the component

maximilianoertel: @Emily-ejag in that case you should still use a `data-testId` at the outer level of your selector. If you do `cy.get('ul > li')` for example, Cypress will search for a `ul > li` on the page and attempt to validate against it. Same and even worse with the `cy.get('div')` as there will likely be many more of those in the DOM

cypress/e2e/partner-admin/default-tests/exportAllUsersFromOrg.cy.js
maximilianoertel: Up to you and the team to decide if you want to tackle this another time, but just saying that this E2E test will still pass after empty or malformed CSVs were generated. That's why, IMO, this should be part of this test

maximilianoertel (changes requested):
Re-opened https://github.com/yeatmanlab/roar-dashboard/pull/846#discussion_r1792589864

cypress/e2e/partner-admin/default-tests/exportAllUsersFromOrg.cy.js
Emily-ejag: Let me double check, this is a function @ksmontville or @lucasxsong wrote, I am re-using. but I will double check

cypress/e2e/partner-admin/default-tests/exportAllUsersFromOrg.cy.js
Emily-ejag: Sure I am not sure how I will use that link, since we don't have a table to compare this since we don't have a table of the users we are exporting from. I will have to after export go to the users page and compare. Asking @richford  and @ksmontville to check if I should implement this.

Emily-ejag: If the answer is yes, we should do this for all of our test that use exporting, like export score and progress reports

cypress/e2e/partner-admin/default-tests/exportAllUsersFromOrg.cy.js
Emily-ejag: Also I moved this function to commands :)

cypress/e2e/partner-admin/default-tests/exportAllUsersFromOrg.cy.js
richford: I think @maximilianoertel is right here. And so are you @Emily-ejag that this should apply to the export score and progress report tests as well. Let's strategize in the next sprint about how to do that. I think it might involve separating out the CSV download logic such that it can be called with mock data.

richford (changes requested):
This is looking good. I concur with @maximilianoertel's testing recommendations. Perhaps that means we will separate out the CSV download behavior so that we can test it independently.

Emily-ejag:
> This is looking good. I concur with @maximilianoertel's testing recommendations. Perhaps that means we will separate out the CSV download behavior so that we can test it independently.

Sounds good, I will move this task to next sprint

Emily-ejag:
Closed, since we have BERF now, and migrating apps to monorepo

PR #848, opened 2024-10-04
Remove hard-coded configs in favour of env variables

maximilianoertel:
## Proposed changes

This PR removes hardcoded Firebase configuration values from the codebase and replaces them with environment variables. This change provides improved security by preventing accidental exposure of sensitive data and allows for easier configuration management across different deployment environments.

Additionally, this PR removes the `.firebaserc` configuration file, following Google's recommendation to exclude it from public repositories to prevent accidental exposure of sensitive information.

**Key highlights**
- Hardcoded config values are replaced with environment variables using the [dotenvx](https://dotenvx.com/) library for encrypted  configurations.
- Environment configurations are stored in a separate private repository and injected as a submodule in the `env-configs` directory.
- Forks of the project can easily set their own configurations by using the traditional `.env` file format supported by Vite at the root of the project. 
- A `.env.example` file is provided at the root, serving as a source of truth for supported environment variables.
- A pre-commit hook is included to ensure no unencrypted dotenv files are accidentally committed to the repository.

Given the changes in how environment configurations are handled, this PR also restructures the npm commands available on the repo. More specifically, it removes all `dev` and `build` command variants in favor of two simple `npm run dev` and `npm run build` commands. This is done to avoid accidentally running the command with the wrong environment configuration and better support forks and alternative environments such as ROAR Clinic or Levante. 

Should you wish to execute a dev or production build with an alternative env configuration, it is recommended to use the `--mode` flag, for example `npm run build -- --mode production` which would use the `.env.production` config.

This approach promotes better separation of concerns, improves maintainability, and enhances the overall security posture of the application.

## Types of changes
- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist

n/a

## Further comments
n/a


src/firekit.js
```diff
@@ -0,0 +1,132 @@
...
+ */
+export async function initializeFirekit() {
+  const firekit = new RoarFirekit({
+    roarConfig: firekitConfig,
+    authPersistence: 'session',
+    markRawConfig: {
+      // @TODO: Check what this is for?
+      auth: false,
+      db: false,
+      functions: false,
```
richford, line 126: There was a time when we though we had to use `markRaw` for all of the Firebase service class instances. So we build this into firekit:

- See [here](https://github.com/yeatmanlab/roar-firekit/blob/21292c4782f1f15696f94cb09539e5cc78f2c113/src/firekit.ts#L209)
- And [here](https://github.com/yeatmanlab/roar-firekit/blob/21292c4782f1f15696f94cb09539e5cc78f2c113/src/firestore/util.ts#L157)

But as you can see, we don't use this. I think we could remove this config here since it is the default behavior. Could you also submit a firekit PR removing this config entirely. I don't believe we use it anywhere and perhaps we shouldn't maintain it if we never intend to use it.

richford (commented):
Looks good to me so far. I left a comment about the `markRaw` config. I'm not approving since this is in draft and it seems like you still want to do more with it. Can you also be sure to copy any env files that you are using to make this work so that you can share privately with the dev team after this is merged?

src/firekit.js
maximilianoertel: Thanks for the explanation! I'll open separate PRs to remove this from the dashboard and firekit repos shortly

richford (approved):

PR #864, opened 2024-10-10
Fix 'view by school' distribution graph

kellyel:
## Proposed changes
The issue: unenrolled students will not have a schoolId in their schools.current array. Therefore, when viewing the distribution graph by school, it will attempt to organize by a field in which some students have undefined as a value. 

~The solution here is to cross reference a student's schools.all and the assignment's assigningOrgs.schools to find the school that the user was a member of when it was assigned. If that fails, we fallback to '0 Unknown School' to prevent the graph from failing catastrophically.~

Rather than doing the cross-referencing on the front end, we will utilize the assignment doc's `assigningOrgs` object which is already an intersection of the administration's assigning orgs and the user's orgs at the time of assignment. This is much more efficient and eliminates the possibility of a user being listed under the wrong school.

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist


src/pages/ScoreReport.vue
```diff
@@ -577,11 +577,17 @@ const computeAssignmentAndRunData = computed(() => {
...
+      const userSchools = user.schools?.all;
+      const assignmentOrgs = assignment.readOrgs.schools;
+      const matchingSchools = userSchools?.filter((schoolId) => assignmentOrgs.includes(schoolId));
+
       // for each row, compute: username, firstName, lastName, assessmentPID, grade, school, all the scores, and routeParams for report link
       const grade = user.studentData?.grade;
       // compute schoolName
       let schoolName = '';
-      const schoolId = user?.schools?.current[0];
+      const schoolId = matchingSchools[0];
```
ksmontville, line 590: Is it possible that this could return the wrong school?

If a student has an old school in their `schools.all` array, and an assignment is also assigned to that old school?

kellyel: Yes, possibly, in the case that a user has switched schools within a district, and that the report is district level. 
This PR will resolve the issue for a majority of cases. A long term solution I've thought about is denormalizing the user's `minimalOrgs` into the assignment document, similar to how we do with the user data. However I feel as if I want to discuss that in a 16th minute when we're not doing sprint review.

richford: Could we see if `user?.schools?.current[0]` is in `assignment.readOrgs.schools` first? And if so, use that as the `schoolId`? And only if it isn't use the fallback that you have defined above where we compute matching schools?

richford (changes requested):

Emily-ejag (approved):

richford (approved):

PR #967, opened 2024-11-22
Update ROAV-Crowding version to 1.1.30

richford:
This PR updates the version of `@bdelab/roav-crowding` to 1.1.30.


Emily-ejag:
there is a 1.1.31 version now

PR #976, opened 2024-12-02
Refactor `<ManageTasks />` component

maximilianoertel:
## Proposed changes
This PR refactors the `<ManageTasks />` component to improve user experience and maintainability by:

- Breaking down the single component into smaller, focused, and reusable components, promoting better code organisation and maintainability.
- Applying UI improvements in regards to overall styling, alignment, loading states, and introducing a more intuitive interface for configuring task parameters.
- Enhancing accessibility by structuring the forms using fieldset elements, adding descriptive legends, and ensuring proper label associations.
- Improving validation by applying validation rules to task parameter configuration, including unique name and blacklist validation rules to task parameter configuration, ensuring data integrity.
- Improving error handling and providing better user feedback by implementing consistent toast notifications for task update/creation errors, enhancing the overall UX.

#### Before
Screenshot 2024-12-03 at 16 31 35

#### After
Screenshot 2024-12-11 at 17 22 18

#### Before
Screenshot 2024-12-11 at 17 24 33

#### After
Screenshot 2024-12-11 at 17 32 46

## Types of changes

What types of changes does this pull request introduce?

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Refactoring (non-breaking change that does not add functionality but makes code cleaner or more efficient)
- [ ] Documentation Update
- [ ] Tests (new or updated tests)
- [ ] Style (changes to code styling)
- [ ] CI (continuous integration changes)
- [ ] Repository Maintenance
- [ ] Other (please describe below)

## Checklist

- [x] I have read the [guidelines for contributing](https://github.com/yeatmanlab/roar-dashboard/blob/main/.github/CONTRIBUTING.md).
- [x] The changes in this PR are as small as they can be. They represent one and only one fix or enhancement.
- [x] Linting checks pass with my changes.
- [x] Any existing unit tests pass with my changes.
- [x] Any existing end-to-end tests pass with my changes.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [ ] If this PR fixes an existing issue, I have added a unit or end-to-end test that will detect if this issue reoccurs.
- [x] I have added JSDoc comments as appropriate.
- [ ] I have added the necessary documentation to the [roar-docs repository](https://github.com/yeatmanlab/roar-docs).
- [x] I have shared this PR on the roar-pr-reviews channel (if I have access)
[...]

PR #1035, opened 2025-02-10
Role Based Access Control

kellyel:
## Proposed changes
This PR, in conjunction with the parallel PR in ROAR-Firekit, implements role-based access control for the dashboard. 

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist


maximilianoertel (changes requested):
Only doing a high-level review until feedback in the `firekit` repo is discussed, but generally speaking: I would recommend using a composable to create an abstraction layer and make the codebase more maintainable. 

Here's an example: 

```js
import { PermissionsService, Permissions } from '@stanford-roar/firekit'
import { useAuthStore } from '@/stores/auth'
import { storeToRefs } from 'pinia'

export function usePermissions() {
  const authStore = useAuthStore()
  const { accessToken } = storeToRefs(authStore)
  
  const userCan = (permission: Permission) => {
    return PermissionsService.canUser(accessToken.value, permission)
  }

  return {
    userCan,
    Permissions
  }
}
```

Then, in the various components we can use that composable in the following manner: 

Only doing a high-level review until feedback in the `firekit` repo is discussed, but generally speaking: I would recommend using a composable to create an abstraction layer and make the codebase more maintainable. 

Here's an example: 

```js
import { usePermissions } from '@/composables/usePermissions'
const { can, Permissions } = usePermissions()

// Before
- if (canUser(accessToken, 'dashboard.users.edit_administrator'))

// After
+ if (userCan(Permissions.Users.LIST))
```

Emily-ejag (dismissed):

src/components/ListUsers.vue
```diff
@@ -226,18 +237,39 @@ const columns = ref([
...
     sort: false,
   },
-  {
+]);
+
+if (
+  userCan(Permissions.Dashboard.Users.EDIT) ||
+  userCan(Permissions.Dashbaord.Users.CHANGE_PASSWORD) ||
+  userCan(Permissions.Dashboard.Administrators.EDIT)
+) {
```
maximilianoertel, line 246: This is a bit tricky as it mixes different permission checks, suggesting that either the component is doing too much or there is a flaw with the permissions structure. 

For instance, this edit profile button should technically be only be shown for `userCan(Permissions.Users.UPDATE)`

src/components/ListUsers.vue
```diff
@@ -226,18 +237,39 @@ const columns = ref([
...
+// | Permissions Handling |
+// +----------------------+
+const canUserEdit = computed(() => {
+  const userType = currentEditUser.value?.userType;
+  if (userType === 'admin') {
+    return userCan(Permissions.Dashboard.Administrators.EDIT);
+  } else {
+    return userCan(Permissions.Dashboard.Users.EDIT);
+  }
+});
```
maximilianoertel, line 271: Could you please elaborate what use-case this is supposed to cover?

src/components/tasks/ManageTasks.vue
maximilianoertel, line 1: This file was removed in https://github.com/yeatmanlab/roar-dashboard/pull/976/files#diff-f377fe742229cc65ad9bc3c06d9a28fa777fa8e50e4b751d94c353dd44431c4a and seems to have been accidentally been re-introduced in the PrimeVue update PR. Any changes to permissions should be applied in the new containers and components introduced in PR #976

src/components/tasks/ManageVariants.vue
maximilianoertel, line 1: Same applies here, this file should no longer exist / be used

src/composables/usePermissions.js
```diff
@@ -0,0 +1,16 @@
+import { PermissionsService, Permissions } from '@bdelab/roar-firekit';
+import { useAuthStore } from '@/store/auth';
+import { storeToRefs } from 'pinia';
+
+export function usePermissions() {
+  const userCan = (permission) => {
+    const authStore = useAuthStore();
+    const { accessToken } = storeToRefs(authStore);
```
maximilianoertel, line 8: If I'm not mistaken, the `authStore` and `accessToken` instances are recreated on every `userCan` call, which isn't optimal. I think we can move these outside the function?

It would also be good to add some basic unit tests to cover the composable logic.

src/pages/RegisterStudents.vue
```diff
@@ -101,7 +101,7 @@
             label="Start Registration"
             class="bg-primary text-white border-none border-round p-2 hover:bg-red-900"
             :icon="activeSubmit ? 'pi pi-spin pi-spinner' : ''"
-            :disabled="activeSubmit"
+            :disabled="activeSubmit || !userCan(Permissions.Dashboard.Users.CREATE)"
```
maximilianoertel, line 104: In general, we should avoid disabling buttons in case of missing permissions, as it can make it unclear why an action isn’t available. Instead, we might want to consider removing the button, providing additional descriptive messaging, or offering an alternative UI.

src/router/index.js
```diff
@@ -453,42 +488,16 @@ router.beforeEach(async (to, from, next) => {
...
-
-  if (isUserSuperAdmin) {
-    next();
-    return;
-  } else if (isUserAdmin) {
-    // LEVANTE dashboard has opened some pages to administrators before the ROAR dashboard
-    // So if isLevante, then allow regular admins to access any route with requireAdmin = true.
-    // and if ROAR, then prohibit regular admins from accessing any route with requireSuperAdmin = true.
-    if (isLevante && requiresAdmin) {
+  const routePermission = _get(to, 'meta.permission', null);
```
maximilianoertel, line 491: No need to import `lodash` for this: 

```suggestion
  const routePermission = to?.meta?.permission ?? null
```

A good ressource for these sort of things: https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore

maximilianoertel (changes requested):

src/components/EditUsersForm.vue
```diff
@@ -202,6 +205,12 @@
       </div>
     </div>
   </div>
+  <div v-else class="form-container">
```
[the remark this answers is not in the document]
kellyel: This if/else chain accounts for admins. There is a case on line 129 for that userType. This case will only be triggered if it's a completely different user type (which we do not support editing for right now), or a user does not have permissions. I will add a message about unsupported userTypes if that is the reason we end up in this case.

src/components/ListUsers.vue
kellyel: This function is used to decide whether to render the button to save changes to a user. `currentEditUser` is the user being edited. Since editing students and admins are different permissions, this function is used to check the user being edited's userType, and if the calling user is able to complete the action. It is used on line 100.

src/components/ListUsers.vue
kellyel: The component is doing to much. In this case, it's a modal that displays either an edit students form (if the user is a student) or an edit admin form (if the user is an admin). I see the value in removing the password change permission, as I check for that permission before rendering the edit password button. However I think this component will still need to check for both users.Update and administrators.Update until this component can be refactored.

src/components/ListUsers.vue
```diff
@@ -41,6 +41,11 @@
       </div>
       <AppSpinner v-else />
       <RoarModal
+        v-if="
+          userCan(Permissions.Dashboard.Users.EDIT) ||
+          userCan(Permissions.Dashboard.Administrators.EDIT) ||
+          userCan(Permissions.Dashboard.Users.CHANGE_PASSWORD)
+        "
```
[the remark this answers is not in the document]
kellyel: There is already a permission check when adding this column to the table, therefore not rendering the buttons at all. This check was just additional, so the browser doesn't have to spend time creating the modal that will never be triggered. However I realize this is superfluous and can be removed.

Emily-ejag (approved):

package.json
```diff
@@ -32,7 +32,7 @@
   },
   "dependencies": {
     "@bdelab/roam-apps": "1.0.11",
-    "@bdelab/roar-firekit": "^9.2.0",
+    "@bdelab/roar-firekit": "^9.6.0-alpha.0",
```
maximilianoertel, line 35: Let's bump this to the actual `firekit` version under which we're publishing the changes

```suggestion
    "@bdelab/roar-firekit": "^9.3.0",
```

maximilianoertel (changes requested):

maximilianoertel (approved):

Emily-ejag (approved):

PR #1042, opened 2025-02-20
Adding task launcher components

lucasxsong:
## Proposed changes

This PR adds the Task Launching feature, which intends to allow administrators (such as teachers or parents) to administer their students/children directly from their authentication session. The feature is implemented by wrapping participant components (such as `HomeParticipant` and `TaskSWR`) with a `/launch/:launchId` route parameter where `:launchId` is the adminUid of the student being launched. This `:launchId` is then passed into the various functions and queries that are called during task instantiation and assessment to store the data under the target student (`:launchId`) from the parent's admin account. 

Super-administrators can launch any user in the store. Partner administrators are limited to user's that belong to an organizations in which they administrate. 

There is a custom `canLaunch` user attribute that toggles the display of the launch button in the score table. Launching components in the **ScoreReport Datatable** are conditionally rendered if `userData.canLaunch == true`. This is just an interim solution to conditionally render these external launch components, may be configured in the future to more closely align with the role-based access control that is established in the dashboard. 

Changes can be tested on the dashboard with the cypress partner admin credentials, as the `canLaunch` attribute is already set in that account. 

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist


firebase/assessment/firestore.rules
```diff
@@ -310,22 +324,22 @@ service cloud.firestore {
...
+            || isAdminForReadOrg('families', adminOrgs));
         }
 
         function canReadRun() {
           return loggedIn() && (myData() || isAdminForAnyAssigningOrg() || isAdminForAnyReadOrg());
         }
 
         function canCreateRun() {
-          return myData();
+          return loggedIn() && (myData() || isAdminForAnyUserOrg())
```
kellyel, line 338: Do we know if this will affect guest/anonymous users running tasks in standalone mode?

src/components/RoarDataTable/RoarDataTable.vue
```diff
@@ -174,6 +174,22 @@
                   />
                 </router-link>
               </div>
+              <div v-else-if="col.launcher">
```
kellyel, line 177: Can this column type be generalized to `externalLink`? If it needs to be specific for a reason that's fine, I just want to minimize feature-specific code in general-use components as much as possible.

src/composables/queries/useUserStudentDataQuery.test.js
```diff
@@ -56,95 +56,95 @@ describe('useUserStudentDataQuery', () => {
     expect(fetchDocById).toHaveBeenCalledWith('users', mockUserRoarId, ['studentData']);
   });
 
-  it('should correctly control the enabled state of the query', async () => {
-    const mockUserRoarId = ref(nanoid());
+  // it('should correctly control the enabled state of the query', async () => {
```
kellyel, line 59: Is this test deprecated?

src/containers/Navigation/Navigation.vue
```diff
@@ -78,11 +78,33 @@ const navbarBlacklist = [
...
+  'Launch Fluency-Alpaca-ES',
+  'Launch RAN',
+  'Launch Crowding',
+  'Launch MEP',
   'ReadAloud',
 ];
 
 const showAccountSettingsLink = computed(() => {
-  return userCan(Permissions.Profile.READ);
+  return userCan(Permissions?.Profile?.READ);
```
kellyel, line 107: I have updated the permissions package so that conditional chaining is no longer required 🙂

src/router/index.js
```diff
@@ -239,7 +489,7 @@ const routes = [
     component: () => import('../pages/RegisterStudents.vue'),
     meta: {
       pageTitle: 'Register Students',
-      permission: Permissions.Users.CREATE,
+      permission: Permissions?.Users?.CREATE,
```
kellyel, line 492: See above about not requiring conditional chaining 🙂

src/router/index.js
```diff
@@ -197,14 +197,264 @@ const routes = [
...
+    component: () => import('../components/tasks/TaskSWR.vue'),
+    props: (route) => ({
+      taskId: 'swr',
+      language: 'en',
+      launchId: route.params.launchId, // Accessing the launchId prop
+    }),
+    meta: { pageTitle: 'SWR' },
+  },
+  {
+    path: '/game/swr-es',
```
kellyel, line 219: Is there a difference between launch routes with `/launch/:launchId/game/{taskId}` and launch routes with `/game/{taskId}`?

src/composables/queries/useUserStudentDataQuery.test.js
lucasxsong: Not deprecated – I'm working on getting these tests to pass with the updated changes to the query. These are just commented out in the interim to make it easier to debug.

src/containers/Navigation/Navigation.vue
lucasxsong: Ah great, thanks for the heads up!

src/router/index.js
lucasxsong: Ah, thanks for pointing this out. I'd missed out on adding the `/launch/:launchId` to some of the routes. Just updated the pr accordingly.

src/components/RoarDataTable/RoarDataTable.vue
lucasxsong: I had tried to generalize it under one component but had run into some issues with recursion (max recursive updates exceeded) I wasn't able to find the root of the problem so my workaround so decided to just create a new column type

cypress/e2e/partner-admin/default-tests/launchExternalStudent.cy.js
```diff
@@ -0,0 +1,33 @@
...
+function returnToAdmin() {
+  cy.get('button').contains('Return to administrator account').click();
+  cy.get('body').should('contain', 'View Administrations');
+}
+
+describe('The partner admin can launch an external student.', () => {
+  it('Selects an administration and launches a student into their tasks', () => {
+    checkUrl();
+    launchExternalStudent();
+    returnToAdmin();
```
maximilianoertel, line 31: Since these are not re-used anywhere here, more generally speaking I don't think it's necessary to split these out to different functions

cypress/e2e/partner-admin/default-tests/launchExternalStudent.cy.js
```diff
@@ -0,0 +1,33 @@
...
+
+function launchExternalStudent() {
+  cy.navigateTo('/launch/yXuZ8S0En1UsOE4C0uh6wUlQ5Wt1');
+  cy.wait(3 * timeout);
+  cy.get('body', { timeout: 10 * timeout }).should('contain', 'external launch mode');
+}
+
+function returnToAdmin() {
+  cy.get('button').contains('Return to administrator account').click();
+  cy.get('body').should('contain', 'View Administrations');
```
maximilianoertel, line 24: Let's use testIds to target elements like button for example, otherwise Cypress needs to traverse the entire DOM looking at every single `<button>`

cypress/e2e/partner-admin/default-tests/launchExternalStudent.cy.js
```diff
@@ -0,0 +1,33 @@
...
+const testUserList = Cypress.env('testUserList');
+
+function checkUrl() {
+  cy.login(testPartnerAdminUsername, testPartnerAdminPassword);
+  cy.navigateTo('/');
+  cy.url({ timeout: timeout }).should('eq', `${baseUrl}/`);
+}
+
+function launchExternalStudent() {
+  cy.navigateTo('/launch/yXuZ8S0En1UsOE4C0uh6wUlQ5Wt1');
```
maximilianoertel, line 17: Is it possible to navigate without hard-coding the path?

src/composables/queries/useUserAssignmentsQuery.js
```diff
@@ -4,23 +4,28 @@ import { useAuthStore } from '@/store/auth';
...
-  const queryConditions = [() => !!roarUid.value];
+  // We need to have the orgId and orgType for a non-superadmin call of an external fetch
+  const queryConditions = [() => !!uid.value && (isExternalCallWithoutSuperAdmin ? orgType && orgIds : true)];
   const { isQueryEnabled, options } = computeQueryOverrides(queryConditions, queryOptions);
 
   return useQuery({
-    queryKey: [USER_ASSIGNMENTS_QUERY_KEY, roarUid],
-    queryFn: () => getUserAssignments(roarUid),
+    queryKey: [USER_ASSIGNMENTS_QUERY_KEY, uid, orgType, orgIds],
+    queryFn: () => getUserAssignments(uid.value, orgType?.value, orgIds?.value),
```
maximilianoertel, line 28: Could you please help me understand why the user assignments query needs `orgType` and `orgIds` as parameters? And why that is related to the current user's super_admin role? My understanding was that in the dashboard, the user will always see a list of all their assignments?

src/constants/queryKeys.js
```diff
@@ -19,6 +19,7 @@ export const SURVEY_RESPONSES_QUERY_KEY = 'survey-responses';
 export const TASKS_QUERY_KEY = 'tasks';
 export const TASK_VARIANTS_QUERY_KEY = 'task-variants';
 export const USER_DATA_QUERY_KEY = 'user';
+export const USERS_DATA_QUERY_KEY = 'users';
```
maximilianoertel, line 22: This constant doesn't seem to be used anywhere? 

```suggestion
```

src/pages/HomeParticipant.vue
```diff
@@ -147,17 +165,56 @@ const {
...
+const isOrgIntersectionReady = ref(false);
+
+const userAssignmentsQueryEnabled = computed(() => {
+  return isOrgIntersectionReady.value && initialized.value;
+});
+
+const isSuperAdmin = computed(() => {
+  if (authStore?.userClaims.claims?.super_admin) return true;
+  return false;
+});
```
maximilianoertel, line 194: We have a `useUserType` composable for this that exposes `isSuperAdmin`

src/pages/HomeParticipant.vue
```diff
@@ -147,17 +165,56 @@ const {
...
+  return orgIntersection;
+});
+
+watch(adminOrgIntersection, (newOrgIntersection) => {
+  orgType.value = newOrgIntersection?.orgType;
+  orgIds.value = newOrgIntersection?.orgIds;
+});
+
+const orgType = ref('');
+const orgIds = ref('');
```
maximilianoertel, line 183: Just wondering if these should default to `null`? 
Down in line 215-216, we could potentially pass empty strings which could create issues downstream

src/pages/ScoreReport.vue
```diff
@@ -1220,6 +1222,19 @@ const scoreReportColumns = computed(() => {
     });
   }
 
+  // For now, only add the launch component if the user has a custom `canLaunch` boolean set in their userData
+  if (authStore.userData.canLaunch) {
```
maximilianoertel, line 1226: Should this not be controlled via the permissions system introduced by @kellyel ?

src/router/index.js
```diff
@@ -204,7 +204,257 @@ const routes = [
...
+    path: '/launch/:launchId/game/roav-mep',
+    name: 'Launch MEP',
+    component: () => import('../components/tasks/TaskMEP.vue'),
+    props: (route) => ({
+      taskId: 'roav-mep',
+      language: 'en',
+      launchId: route.params.launchId, // Accessing the launchId prop
+    }),
+    meta: { pageTitle: 'MEP' },
+  },
```
maximilianoertel, line 457: For all these new routes it would be great to use/extend the `APP_ROUTES` constant. 

Also, there would be the possibility to define the above as [nested routes](https://router.vuejs.org/guide/essentials/nested-routes.html#Nested-Routes) to make it more obvious that these are launcher related routes

maximilianoertel (changes requested):

src/composables/queries/useUserStudentDataQuery.js
```diff
@@ -14,17 +15,18 @@ import { FIRESTORE_COLLECTIONS } from '@/constants/firebase';
  * @param {QueryOptions|undefined} queryOptions – Optional TanStack query options.
  * @returns {UseQueryResult} The TanStack query result.
  */
-const useUserStudentDataQuery = (queryOptions = undefined) => {
+const useUserStudentDataQuery = (queryOptions = undefined, launchId = null) => {
```
maximilianoertel, line 18: All other existing query composable have the `queryOptions` as last parameter, and I think it would be good to keep that consistency?

src/constants/queryKeys.js
lucasxsong: Ah, good find. I'd added this for a query that is now removed. Thanks for the suggestion, I'll remove it now.

src/composables/queries/useUserStudentDataQuery.js
lucasxsong: Good point, I'll make that changes right now.

src/pages/HomeParticipant.vue
lucasxsong: I hadn't realized, thanks! I've updated to use the composable.

src/composables/queries/useUserAssignmentsQuery.js
lucasxsong: Apologies for the confusion, let me try to clear it up:

There are three cases in the use of this query:

1. Call of the query from a student: The query doesn't need to contain any orgIds or orgTypes so the query should be enabled even in their absence 
2. Call of the query from a partner admin: In this case an orgId and orgType needs be passed in so that the query that is generated from this call (see `getUserAssignments` and further `getAssignmentsRequestBody`) has that necessary attributes to allow for validation against the firestore rules to determine whether the administrator has permission to read from said student's assignments 
3. Call of the query from a super admin: In this case, an orgId and orgType need not be passed in as super-admins are able to read data from anywhere in the database. So we still allow the query to be enabled even if orgId and orgType is undefined.

src/composables/queries/useUserStudentDataQuery.js
```diff
@@ -14,17 +15,18 @@ import { FIRESTORE_COLLECTIONS } from '@/constants/firebase';
...
   const { isQueryEnabled, options } = computeQueryOverrides(queryConditions, queryOptions);
 
   return useQuery({
-    queryKey: [USER_STUDENT_DATA_QUERY_KEY, roarUid],
-    queryFn: () => fetchDocById(FIRESTORE_COLLECTIONS.USERS, roarUid, ['studentData']),
+    queryKey: [USER_STUDENT_DATA_QUERY_KEY, uid],
+    queryFn: () => fetchDocById(FIRESTORE_COLLECTIONS.USERS, uid, ['studentData']),
     enabled: isQueryEnabled,
     ...options,
   });
```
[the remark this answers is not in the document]
lucasxsong: Good point, I've changed the query to take in the param as `userId`, and also updated the JSDoc comments in the updated queries to reflect the new params.

src/router/index.js
lucasxsong: Thanks for the suggestion, I've just updated the game routes to utilize the newly created `GAME_ROUTES` route constants and updated the launch routes to use the `GAME_ROUTES` as children routes under the `/launch` root.

src/pages/ScoreReport.vue
lucasxsong: I've just updated to use the permission `Permissions.Tasks.Launch` to control the rendering of this column. Thanks!

src/composables/queries/useUserAssignmentsQuery.js
```diff
@@ -4,23 +4,32 @@ import { useAuthStore } from '@/store/auth';
...
+ * @param {String|undefined} orgType– The orgtype that is shared between the admin user and the target studnet user
+ * @param {String|undefined} orgId – The orgId that is shared between the admin user and the target studnet user
+ * OrgType and OrgId are passed in in order to construct a query that will validate the administrator's permissions to view target user data
  * @returns {UseQueryResult} The TanStack query result.
  */
-const useUserAssignmentsQuery = (queryOptions = undefined) => {
+const useUserAssignmentsQuery = (queryOptions = undefined, userId = null, orgType = null, orgIds = null) => {
   const authStore = useAuthStore();
   const { roarUid } = storeToRefs(authStore);
+  const uid = computed(() => userId || roarUid.value);
```
richford, line 22: Should this be the nullish coalescing operator instead of or?
```suggestion
  const uid = computed(() => userId ?? roarUid.value);
```

richford (dismissed):

src/composables/queries/useUserAssignmentsQuery.js
lucasxsong: That's correct, thanks for the suggestion, I've committed your change.

richford (approved):

kellyel (approved):

richford (approved):

richford (approved):

kellyel (dismissed):

richford (approved):

maximilianoertel (approved):

PR #1045, opened 2025-02-27
CollectionGroup Query Rules for /variants/

kellyel:
## Proposed changes
This PR updates the firestore rules for the variants collection. A separate permission is required to allow CollectionGroup queries. Until now, the collection was being accessed exclusively by Super Admins. Now with the Platform Administrator role, a rule was required to allow them access to this collection.

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist

### DIFF
--- a/firebase/assessment/firestore.rules
+++ b/firebase/assessment/firestore.rules
@@ -482,5 +482,12 @@ service cloud.firestore {
         allow update: if canUpdateTask();
       }
     }
+
+    // Security rules for the collection **GROUP** variants
+    // We only need to specify read permissions, as write and update operations will
+    // require a specific taskId, therefore subject to the rules above.
+    match /{path=**}/variants/{variantId} {
+      allow read: if loggedIn();
+    }
   }
 }


lucasxsong (approved):

maximilianoertel (changes requested):
According to the PR description, the `/variants/*` should only be accessed by super admins and partner admins, correct? If so, should we not use the new role claim that you've introduced? 

Haven't tested it, but as very rough example: 

```
function hasRole(expectedRole) {
  return request.auth.token.role == expectedRole
}

match /{path=**}/variants/{variantId} {
  allow read: if hasRole('super_admin') || hasRole('partner_admin);
}
```

kellyel:
We have always considered this collection available to read to all. However until this point, nobody but super admins have had any reason to do so. Therefore we did not realize that the rule allowing people to read this collection did not also cover collection group queries. I don't think further restriction is necessary. Happy to hear opinions in favor of restricting it, though.

maximilianoertel (approved):
Thanks for the explanation @kellyel. I'll go ahead and approve the change as it's in line with the existing rules. 

Generally speaking though, I would argue that if a collection is designed to only be read by a subset of users (for example super admins and platform admins as described in the PR description), then only those should be able to access it.

PR #1079, opened 2025-03-26
Add query for ActivationCodes by orgId

lucasxsong:
## Proposed changes

Adds a query composable and corresponding helpers to query for activationCodes by orgId. 

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)
- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist

### DIFF
--- a/src/composables/queries/useActivationCodeQuery.js
+++ b/src/composables/queries/useActivationCodeQuery.js
@@ -1,0 +1,28 @@
+import { useQuery } from '@tanstack/vue-query';
+import { computeQueryOverrides } from '@/helpers/computeQueryOverrides';
+import { fetchDocumentsById } from '@/helpers/query/utils';
+import { hasArrayEntries } from '@/helpers/hasArrayEntries';
+import { FIRESTORE_COLLECTIONS } from '@/constants/firebase';
+import { ACTIVATION_CODE_QUERY_KEY } from '../../constants/queryKeys';
+
+/**
+ * Classes query.
+ *
+ * @param {Array} orgId – The orgId of activation codes to fetch.
+ * @param {QueryOptions|undefined} queryOptions – Optional TanStack query options.
+ * @returns {UseQueryResult} The TanStack query result.
+ */
+const useActivationCodeQuery = (orgId, queryOptions = undefined) => {
+  // Ensure all necessary data is loaded before enabling the query.
+  const conditions = [() => orgId];
+  const { isQueryEnabled, options } = computeQueryOverrides(conditions, queryOptions);
+
+  return useQuery({
+    queryKey: [ACTIVATION_CODE_QUERY_KEY, orgId],
+    queryFn: () => activationCodeFetcher(orgId),
+    enabled: isQueryEnabled,
+    ...options,
+  });
+};
+
+export default useActivationCodeQuery;
--- a/src/composables/queries/useActivationCodeQuery.js
+++ b/src/composables/queries/useActivationCodeQuery.js
@@ -1,9 +1,7 @@
 import { useQuery } from '@tanstack/vue-query';
 import { computeQueryOverrides } from '@/helpers/computeQueryOverrides';
-import { fetchDocumentsById } from '@/helpers/query/utils';
-import { hasArrayEntries } from '@/helpers/hasArrayEntries';
-import { FIRESTORE_COLLECTIONS } from '@/constants/firebase';
-import { ACTIVATION_CODE_QUERY_KEY } from '../../constants/queryKeys';
+import { ACTIVATION_CODE_QUERY_KEY } from '@/constants/queryKeys';
+import { activationCodeFetcher } from '@/helpers/query/activationCodes';
 
 /**
  * Classes query.
@@ -14,7 +12,7 @@ import { ACTIVATION_CODE_QUERY_KEY } from '../../constants/queryKeys';
  */
 const useActivationCodeQuery = (orgId, queryOptions = undefined) => {
   // Ensure all necessary data is loaded before enabling the query.
-  const conditions = [() => orgId];
+  const conditions = [() => true];
   const { isQueryEnabled, options } = computeQueryOverrides(conditions, queryOptions);
 
   return useQuery({
--- a/src/helpers/query/activationCodes.js
+++ b/src/helpers/query/activationCodes.js
@@ -1,0 +1,36 @@
+/**
+ * Fetches Activation Codes for an orgId.
+ * @param orgId The target orgId for which activation codes we want to return
+ * @returns {Promise<Array<Object>>} A promise that resolves to an array of legal document objects.
+ */
+export const getActivationCodesRequestBody = ({ orgId }) => {
+  const requestBody = {
+    structuredQuery: {
+      from: [
+        {
+          collectionId: 'activationCodes',
+          allDescendants: false,
+        },
+      ],
+    },
+  };
+
+  if (orgId) {
+    requestBody.structuredQuery.where = {
+      fieldFilter: {
+        field: { fieldPath: 'orgId' },
+        op: 'EQUAL',
+        value: { stringValue: orgId },
+      },
+    };
+  }
+
+  return requestBody;
+};
+
+export const activationCodeFetcher = async (orgId) => {
+  const axiosInstance = getAxiosInstance('admin');
+  const requestBody = getActivationCodesRequestBody({ orgId });
+  const response = await axiosInstance.post('/:runQuery', requestBody);
+  return response.data;
+};
--- a/src/helpers/query/activationCodes.js
+++ b/src/helpers/query/activationCodes.js
@@ -1,3 +1,4 @@
+import { getAxiosInstance } from './utils';
 /**
  * Fetches Activation Codes for an orgId.
  * @param orgId The target orgId for which activation codes we want to return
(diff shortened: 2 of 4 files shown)


src/composables/queries/useActivationCodeQuery.js
```diff
@@ -0,0 +1,22 @@
+import { useQuery } from '@tanstack/vue-query';
+import { computeQueryOverrides } from '@/helpers/computeQueryOverrides';
```
Emily-ejag, line 2:
```suggestion

```

Emily-ejag (changes requested):

src/composables/queries/useActivationCodeQuery.js
Emily-ejag: lint issue

richford: Or better yet, could we use `computeQueryOverrides` as we do in the other query composables. I think it would be something like:

```
  const conditions = [() => _isString(orgId)];
  const { isQueryEnabled, options } = computeQueryOverrides(conditions, queryOptions);
```

And then in the `return useQuery` block, add the parameters
```
    enabled: isQueryEnabled,
    ...options,
```
instead of
```
  ...queryOptions
```

You'll have to import
```
import _isString from "lodash/isString"
```

This would also allow you to adopt the additional tests for the query composable, like the ones in `useAdministrationsQuery.test.js` for example.

richford (changes requested):

richford (commented):
Not a blocking comment. Just curious. See my inline question.

src/composables/queries/useActivationCodeQuery.js
```diff
@@ -12,10 +13,15 @@ import { activationCodeFetcher } from '@/helpers/query/activationCodes';
...
 const useActivationCodeQuery = (orgId, queryOptions = undefined) => {
   // Ensure all necessary data is loaded before enabling the query.
+  const queryConditions = [() => _isString(orgId)];
+  const { isQueryEnabled, options } = computeQueryOverrides(queryConditions, queryOptions);
   return useQuery({
     queryKey: [ACTIVATION_CODE_QUERY_KEY, orgId],
     queryFn: () => activationCodeFetcher(orgId),
-    ...queryOptions,
+    enabled: isQueryEnabled,
+    staleTime: 5 * 60 * 1000, // 5 minutes
```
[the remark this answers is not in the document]
richford: Actually, this might be causing the test failures since the default in `src/plugins.js` is Cypress dependent.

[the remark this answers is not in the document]
lucasxsong: I believe the majority of the test failures were due to an out of date Firekit version that was causing permissions errors. I've since removed the stale time as well.

Emily-ejag (dismissed):

src/composables/queries/useActivationCodeQuery.js
```diff
@@ -0,0 +1,27 @@
+import { useQuery } from '@tanstack/vue-query';
+import { computeQueryOverrides } from '@/helpers/computeQueryOverrides';
+import { ACTIVATION_CODE_QUERY_KEY } from '@/constants/queryKeys';
+import { activationCodeFetcher } from '@/helpers/query/activationCodes';
+import _isString from 'lodash/isString';
+
+/**
+ * Classes query.
+ *
+ * @param {Array} orgId – The orgId of activation codes to fetch.
```
maximilianoertel, line 10:
```suggestion
 * @param {String} orgId – The orgId to fetch an activation code for.
```

src/composables/queries/useActivationCodeQuery.js
```diff
@@ -0,0 +1,27 @@
...
+ * @param {QueryOptions|undefined} queryOptions – Optional TanStack query options.
+ * @returns {UseQueryResult} The TanStack query result.
+ */
+const useActivationCodeQuery = (orgId, queryOptions = undefined) => {
```
maximilianoertel, line 14: The `orgId` is, I believe a `ref` to ensure reactivity in the components that use the composable? If so, the `activationCodeFetcher` (aka `getActivationCodeByOrgId`) should use the `toValue` helper (see examples on other query composables)

src/helpers/query/activationCodes.js
```diff
@@ -0,0 +1,37 @@
...
+
+  if (orgId) {
+    requestBody.structuredQuery.where = {
+      fieldFilter: {
+        field: { fieldPath: 'orgId' },
+        op: 'EQUAL',
+        value: { stringValue: orgId },
+      },
+    };
+  }
```
maximilianoertel, line 27: Can you please elaborate why the `orgId` is only added conditionally? From the rest of the code, it looks like we always need to pass the `orgId`?

maximilianoertel (changes requested):

src/composables/queries/useActivationCodeQuery.js
```diff
@@ -0,0 +1,27 @@
...
+  // Ensure all necessary data is loaded before enabling the query.
+  const queryConditions = [() => _isString(orgId)];
+  const { isQueryEnabled, options } = computeQueryOverrides(queryConditions, queryOptions);
+  return useQuery({
+    queryKey: [ACTIVATION_CODE_QUERY_KEY, orgId],
+    queryFn: () => activationCodeFetcher(orgId),
+    enabled: isQueryEnabled,
+    refetchOnWindowFocus: 'always',
```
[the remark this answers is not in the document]
lucasxsong: This was added by windsurf – I think the default is probably the better options so I'll remove this now.

src/composables/queries/useActivationCodeQuery.js
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Classes query.
+ *
+ * @param {Array} orgId – The orgId of activation codes to fetch.
+ * @param {QueryOptions|undefined} queryOptions – Optional TanStack query options.
+ * @returns {UseQueryResult} The TanStack query result.
+ */
+const useActivationCodeQuery = (orgId, queryOptions = undefined) => {
+  // Ensure all necessary data is loaded before enabling the query.
+  const queryConditions = [() => _isString(orgId)];
```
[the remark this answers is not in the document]
lucasxsong: Good point, I'll change the query condition to use `_isString` instead.

src/composables/queries/useActivationCodeQuery.js
lucasxsong: Good point, I've added the `toValue` wrapping `orgId` here

src/helpers/query/activationCodes.js
lucasxsong: I had previously intended the query to return all `activationCodes` if the orgId was not passed in, but had redesigned after realizing that the query would likely only be used to query codes fro an orgId. I can remove the conditional statement here.

richford (dismissed):

src/composables/queries/useActivationCodeQuery.test.js
maximilianoertel, line 1: It would be good to extend the tests to cover the following: 
- `orgId` parameter value type (ref vs regular value)
- queryOptions should not override internally computed values
- enable query only once parameter is available (incl. reactivity)

Some examples can be found in `useAdministrationsQuery.test.js`

src/composables/queries/useActivationCodeQuery.js
```diff
@@ -0,0 +1,26 @@
...
+import { toValue } from 'vue';
+
+/**
+ * Classes query.
+ *
+ * @param {String} orgId – The orgId to fetch activation codes for.
+ * @param {QueryOptions|undefined} queryOptions – Optional TanStack query options.
+ * @returns {UseQueryResult} The TanStack query result.
+ */
+const useActivationCodeQuery = (orgId, queryOptions = undefined) => {
```
maximilianoertel, line 14: After clarification, we should pluralize the composable name as it's designed to fetch a collection of activation codes, not a single one:
```diff
- src/composables/queries/useActivationCodeQuery.js
+ src/composables/queries/useActivationCodesQuery.js
```

src/helpers/query/activationCodes.js
```diff
@@ -0,0 +1,41 @@
...
+
+  return requestBody;
+};
+
+/**
+ *  Fetches activation code for an orgId
+ *  @param orgId The target orgId for which activation codes we want to return
+ *  @returns {Promise<Array<Object>>} A promise that resolves to an array of activation codes.
+ * */
+export const getActivationCodeByOrgId = async (orgId) => {
```
maximilianoertel, line 36: Similarly, this method should be pluralized as well: 

```suggestion
export const getActivationCodesByOrgId = async (orgId) => {
```

maximilianoertel (changes requested):

src/composables/queries/useActivationCodeQuery.js
lucasxsong: Good point, I'll update the function names to clarify that a collection of activation codes is being returned.

maximilianoertel (dismissed):
🚀

richford (dismissed):

richford:
@lucasxsong , LGTM but now there are merge conflicts.

Emily-ejag (approved):

richford (approved):

maximilianoertel (approved):

PR #1215, opened 2025-05-19
Fix Edit Administrations

kellyel:
## Proposed changes
This PR fixes a few issues that prevented the edit administrations form from working properly. 
- Fixed issue where tasks would fail to show up in task picker on page load
- Fixed issue where assigning orgs are not pre-selected in org picker, and could be selected multiple times
- Fixed issue where administration data would populate fields late / not at all in some cases

These were the issues I was able to reproduce on my computer. If there are others to fix I'd be happy to look into them. 

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist

### DIFF
--- a/src/components/CreateAdministration.vue
+++ b/src/components/CreateAdministration.vue
@@ -18,7 +18,7 @@
             <PvFloatLabel>
               <PvInputText
                 id="administration-name"
-                v-formTypel="state.administrationName"
+                v-model="state.administrationName"
                 class="w-full"
                 data-cy="input-administration-name"
               />
@@ -35,7 +35,7 @@
             <PvFloatLabel>
               <PvInputText
                 id="administration-public-name"
-                v-formTypel="state.administrationPublicName"
+                v-model="state.administrationPublicName"
                 class="w-full"
                 data-cy="input-administration-name-public"
               />
--- a/src/components/CreateAdministration.vue
+++ b/src/components/CreateAdministration.vue
@@ -753,8 +753,4 @@ watch(state, (newState) => {
   border-radius: 50%;
   transform: translate(-50%, -50%);
 }
-.loading-container {
-  width: 100%;
-  text-align: center;
-}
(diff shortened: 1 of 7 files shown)


richford (approved):

src/components/CreateAdministration.vue
```diff
@@ -600,6 +605,12 @@ watch([existingAdministrationData, allVariants], ([adminInfo, allVariantInfo]) =
     }
   }
 });
+
+watch(state, (newState) => {
+  if (newState.administrationName && newState.administrationPublicName && newState.dateStarted && newState.dateClosed) {
+    statePopulated.value = true;
+  }
+});
```
ksmontville, line 613: I think that there may be a minor bug related to this watcher.

If you select an administration to edit, everything works as expected the first time in terms of populating the form, etc.

If you navigate back to the home page, and then select the same administration to edit again, the page will hang at the "Fetching Administration Data" loader.

This suggests to me that the value of `statePopulated` is being set to `false,` and not updating to `true` because `state` hasn't changed when returning to the form the second time.

This is resolved by refreshing the page before attemping to edit the administration the second time.

Perhaps this could be addressed with an `unMount()` hook which could reset the state before navigating back to the home page?

ksmontville (changes requested):

src/components/CreateAdministration.vue
kellyel: Good catch. The issue here was that the watcher populating state was looking for a change in the object returned by the query. In the event of the query being cached, no changes were required, as you deduced. Adding `{ immediate: true }` allows the watcher to fire on page load, thus handling the case in which the query does not need to refetch. If the query is still fetching, there are no side effects for the immediate call, and it will be updated as normal. Thanks for catching that.

richford (approved):

ksmontville (approved):

PR #1229, opened 2025-06-03
Introduce presets for Administration Dates

kellyel:
## Proposed changes
This PR
- Creates an AdministrationDatePicker component that allows the user to use presets or manually entered dates
- Creates Date Input component
- Creates tests for both new components

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist


richford (dismissed):
LGTM. A big improvement over the status quo in terms of modularity and testing.

@maximilianoertel, please review as well.

Also @kellyel, I recall there was an issue with time zones and dates input in the administration form being shifted by 7 hours (the difference between pacific time and UTC). Can you confirm that the current edits resolve this issue?

maximilianoertel (changes requested):
Happy to see this new implementation break down existing components into smaller dumb/presentational ones! 

In addition to some minor change requests, I do see some issues though about timezone handling. More on that below in the comments. 

Additionally, I believe we can improve the UI/UX by making the full date picker type selector full-width as shared on the screenshot below:

Screenshot 2025-06-05 at 15 00 12



Finally, when switching between "Help me choose" and "I know what to select", there is a reflow as the content heights differ. Not a must, but it would be good to address that to prevent that reflow as this should be a relatively quick fix.

src/components/AdministrationDatePicker/AdministrationDatePicker.vue
```diff
@@ -0,0 +1,144 @@
...
+            { label: 'I know what to select', value: 'custom' },
+          ]"
+          :allow-empty="false"
+          option-label="label"
+          option-value="value"
+          :pt="{ root: { class: 'w-full' }, pcToggleButton: { root: { class: 'w-full' } } }"
+          @change="currentModeChange"
+        />
+      </div>
+      <div v-if="currentMode === DATEPICKER_MODE.PRESETS" class="flex flex-row w-full my-2">
```
maximilianoertel, line 18: Let's make this responsive! 

Unfortunately, as you know, Tailwind isn't fully working on this repo so we need to do a few things: 

```suggestion
      <div v-if="decision === 'presets'" class="grid grid-cols-2 gap-2 xl:grid-cols-4">
```

Then, inside the scoped script we need to override the `.grid` class: 

```css
.grid {
  display: grid !important;
  margin: 0 !important;
}
```

And finally, in `style.css` add the following: 
```scss
@media (min-width: 1280px) {
  .xl\:grid-cols-4 {
    grid-template-columns: repeat(4, minmax(0, 1fr));
  }
}
```

src/components/AdministrationDatePicker/AdministrationDatePicker.vue
```diff
@@ -0,0 +1,144 @@
...
+        <div
+          v-for="(preset, key) in datePresets"
+          :key="key"
+          class="card"
+          :class="{ 'selected-card': selectedPreset === key }"
+          @click="presetChange(key)"
+        >
```
maximilianoertel, line 25: These should ideally be radio inputs as they are technically be part of a form, or at least be something like `button` elements for accessibility purposes (with corresponding `aria` attributes), as these `div` elements have no real semantic meaning for screen readers. If you tab through the page, you'll see that the buttons are completely skipped

src/components/AdministrationDatePicker/presets.js
maximilianoertel, line 1: Would be great to see some unit tests to ensure the date calculations remain correct, especially around year transitions and catch potential issues if someone modifies the presets

maximilianoertel (changes requested):

richford (approved):

PR #1240, opened 2025-06-13
Fix Next Steps PDF link 

amyyeung17:
## Proposed changes

This PR adds the updated Next Steps PDF to the public folder and removes the old version from assets. I included video showing some changes but wasn't sure how to test PDF link without signing in. 

https://github.com/user-attachments/assets/13ce6603-e7a3-4a92-a691-cf54cb929d4a

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist

Video in place of test.


src/pages/ScoreReport.vue
```diff
@@ -1501,7 +1502,6 @@ unsubscribe = authStore.$subscribe(async (mutation, state) => {
 onMounted(async () => {
   TaskReport = (await import('@/components/reports/tasks/TaskReport.vue')).default;
   DistributionChartOverview = (await import('@/components/reports/DistributionChartOverview.vue')).default;
-  NextSteps = (await import('@/assets/NextSteps.pdf')).default;
```
kellyel, line 1504: We use the pattern of `(await import(...)).default` here to reduce chunk size when building the project. I believe this pattern would still work with the path updated to point to the new location of the next steps file. Ideally we would keep this consistent, would you mind trying this?

kellyel (changes requested):
I've confirmed this fix works! Excellent job. Just a few nitpicks and I will be ready to approve

src/pages/ScoreReport.vue
maximilianoertel: With the changes that @amyyeung17 is pushing, the PDF no longer needs to be imported. The PDF now resides in the public directory and is considered a static asset that is automatically copied into the dist directory.

maximilianoertel (approved):
Looks/works great! 

Only minor suggestion would be to move the PDF into a subdirectory inside the `public`, something like `public/docs/` or `public/assets/` so that the `public` directory is a little more structured if we ever add more static assets. But like I said, not _needed_

kellyel (approved):

richford (approved):

PR #1272, opened 2025-07-30
Convert repository to monorepo

maximilianoertel:
## Proposed changes
In preparation for the upcoming backend refactor, this PR converts the existing dashboard repository to a monorepo, powered by [Turborepo](https://turborepo.com/). 

For the time being, this monorepo only includes the existing dashboard app. Configuration for ESLint and Prettier were extracted into separate packages that can be re-used by the different apps in this monorepo. 

## Types of changes
- [ ] Documentation Update
- [x] Repository Maintenance

Resolves https://github.com/yeatmanlab/roar-project-management/issues/1115
Resolves https://github.com/yeatmanlab/roar-project-management/issues/1117


apps/dashboard/src/helpers/consent.js
richford, line 1: Wasn't this renamed to `checkConsentRenewalDate.js` in #1289?

richford (commented):
LGTM, but I have two questions:

1. Didn't we change the consent helper filename in #1289? That change doesn't appear to be reflected here.
2. Should the env-configs be at the top level of the monorepo so that other packages can use them?

I don't think either of these rise to the level of "rejection" but I wanted to ask here first before I approve.

maximilianoertel:
> 1. Didn't we change the consent helper filename in #1289? That change doesn't appear to be reflected here.

Good catch, the merge did introduce `checkConsentRenewalDate.js` but didn't delete `consent.js`. That should be fixed now! 

> 2. Should the env-configs be at the top level of the monorepo so that other packages can use them?

Actually, `env-configs` was only introduced because the team wanted to retain an easy way to switch between dev, staging and prod configs when we removed the hard-coded Firebase configs. Ideally, this submodule wouldn't even exist and each app (dashboard, backend) would have their respective `.env` files. For the time being, the `env-configs` submodule will remain, but only in the dashboard app. We can discuss a way forward next week or so?

maximilianoertel:
Great catch, thanks @ksmontville – issue is now resolved and the PR is ready for a re-review!

richford (approved):

ksmontville (approved):

PR #1286, opened 2025-08-13
Automate CSP injection

richford:
### Why is this needed

This PR makes the Content Security Policy header in the firebase/admin/firebase.{env}.json files easier to maintain. The current header stored in that JSON file is a very long line that is difficult to diff and maintain.

### What else did I try

Unfortunately, we cannot duplicate headers as I once thought. For example,

```json
{
  "headers": [
    { "key": "Content-Security-Policy-Report-Only", "value": "default-src 'self'" },
    { "key": "Content-Security-Policy-Report-Only", "value": "script-src  'self' https://cdn.example.com" },
  ]
}
```

will not work. It seems that under the hood, Firebase only takes the last header for each key rather than concatenating values. Also unfortunately, the "whitespace trick" of writing won't work either.

```
{
  "headers": [
    { "key": "Content-Security-Policy-Report-Only", "value": "
        default-src 'self';
        script-src  'self' https://cdn.example.com;
        "}
  ]
}
```

This only works when you can insert real line-feed characters. But the firebase.json file must be valid JSON, which does not allow literal new-lines inside of a string. If we try to escape them (`\n`), that kind of defeats the whole purpose. (I also think Firebase will reject those headers anyway for security reasons.)

### What did I do

So instead, my solution was to write a predeploy script that allows us to keep a human-readable, multiline CSP template file and then turn it into a single line to inject into the various firebase.json files. I also saw that we have firebase project IDs in the CSP file and decided to insert tokens into the template that can be replaced by our dotenvx environment variables.

- `firebase/admin/csp.template.json` (new)
  - Human-readable CSP directives as arrays with tokens
- `scripts/build_firebase_config.js` (new)
  - Loads encrypted env files via `loadDotenvFiles`
  - Replaces `__ENV__`, `__ADMIN_PROJECT_ID__`, `__ASSESSMENT_PROJECT_ID__` tokens in `csp.template.json`
  - Produces a single-line `Content-Security-Policy-Report-Only` header and injects it into `firebase.<env>.json`
- `firebase/admin/firebase.{dev,staging,production}.json`
  - Add predeploy hook: `node scripts/build_firebase_config.js <env>`
  - Keep placeholder CSP header—overwritten at deploy time
- `scripts/load_dot_env_files.js`
  - Central helper that locates and decrypts env files (env-configs/)
  - This logic was located in `vite.config.js`. I simply moved it's location so that it could be used by both Vite and Node.
- `vite.config.js`
  - Replaced dotenv logic with shared helper

### How to test

- Test existing dotenvx injection:
  - `npm run dev` should work as usual and inject the dev dotenvx files.
- Test the new predeploy script:
  - Make sure your working directory is clean and you have no unstaged changes.
  - Execute
    - `node scripts/build_firebase_config.js dev`
    - `node scripts/build_firebase_config.js staging`
    - `node scripts/build_firebase_config.js production`
  - Use `git diff` to inspect the changes to `firebase.dev.json`, `firebase.staging.json`, and `firebase.production.json`. Compare these CSP headers to the CSP headers on main. This will be the hardest part because, as mentioned above, these headers are long and very hard to diff. If this all works, it will be the last time you have to do it.
  - When you are satisfied with the changes, you can revert them using `git checkout -- firebase/admin`.
  - Additionally, check the headers for the PR preview link hosting deployment.
    - Open up the PR preview link site.
    - Open up the dev tools and go to the network tab.
    - Click the top-level request and choose Headers --> Response Headers.
    - Confirm that the value matches what you expect.

### Remaining work

- We need to figure out how to make sure the changed headers don't get accidentally committed in CI. I tried generating a different output file rather than editing in place. But then we can't use the built in `predeploy` from the firebase.json file. If that hook edits a different file, then the CLI will still deploy the original untouched one. We could generate a different file as part of a separate script but then we have to remember to regenerate and reference the generated file every time we deploy. But we now have to make sure that there is no commit logic after deployment in our CI.

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist


kellyel (dismissed):

maximilianoertel (changes requested):
Love where this is going as Firebase Hosting really does not make this easy to maintain. However, I've got some concerns with the current implementation: 

- In its current form, either a dev or a misconfigured CI workflow could end up accidentally committing the mutated firebase config file. Whilst not a critical issue, this would cause issues. Instead, I think we should tracking a `firebase.template.json`, gitignore the generated `firebase.json`, and always build the latter for dev and CI. This should eliminate accidental commits? 

- The Vite dev server is currently configured to load the CSP headers for local development (see https://github.com/yeatmanlab/roar-dashboard/blob/8a291aaeaa2c74725290f57c7b610e9b6b23424d/vite.config.js#L13-L50). Now, a simple `npm run dev` would fail if the script wasn't executed beforehand. Hence, it would be great if we could adjust the Vite config to auto-generate the `firebase.json` for local dev. 

- Initially, the repo contains three distinct `firebase.{env}.json` files because of some of the different values. If we not use a script to inject those dynamically, we should only track a single `firebase.template.json`. Potentially, the `csp.template.json` would need to become `headers.template.json` to cater for all header configs.

richford:
Thanks @maximilianoertel for your comments. I believe I've addressed all of them and this is ready for re-review.

maximilianoertel (dismissed):

kellyel (approved):

maximilianoertel (changes requested):
> [!WARNING]  
> This repository was recently converted into a monorepo (see #1272).  
> Please rebase or update your branch against the latest `main` to ensure compatibility with the new structure.

PR #1317, opened 2025-09-01
Implement backwards-compatible score field access

richford:
### Problem

ROAR assessments are standardizing field names used to store scores, but score reports need to maintain backwards compatibility with existing legacy field names during the transition period.

### Solution

This PR replaces the `getScoreKeys` function with a new `getScoreValue` function that provides automatic fallback from new standardized field names to legacy field names.

- Add SCORE_FIELD_MAPPINGS configuration for all task types
- Create getScoreValue() with automatic new -> legacy fallback
- Update all components to use new score access pattern
- Remove deprecated getScoreKeys function
- Add comprehensive test coverage for all taskIds

I've also added a `SCORE_FIELD_MIGRATION_GUIDE.md` that lives next to `src/helpers/reports.js`. This provides implementation details and a reminder for how to update score fields. Please read through this guide as well as you review this PR.

### Future score migration

My vision is that as each of the assessment versions are bumped on the dashboard, in the same PR, we modify the new scoring fields to align with the changed made by the assessment. For example, each of the following PR introduce new scoring fields for their assessments.

- `swr`: https://github.com/yeatmanlab/roar-swr/pull/254
- `swr-es`: https://github.com/yeatmanlab/roar-swr/pull/255
- `pa`: https://github.com/yeatmanlab/roar-pa/pull/133
- `sre`: https://github.com/yeatmanlab/roar-sre/pull/140

And there will be more for `sre-es` and `letter`. My hope is that this streamlines the score field updating process so that bumping each of those assessment versions is as simple as updating package.json, package-lock.json and one line in src/helpers/reports.js.

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist


apps/dashboard/src/helpers/reports.js
```diff
@@ -698,71 +699,259 @@ export function getTagColor(supportLevel) {
...
   }
-  if (taskId === 'phonics') {
-    percentileScoreKey = 'totalPercentCorrect';
-    percentileScoreDisplayKey = 'totalPercentCorrect';
-    standardScoreKey = 'totalPercentCorrect';
-    standardScoreDisplayKey = 'totalPercentCorrect';
-    rawScoreKey = 'totalCorrect';
+
+  if (!scoresObject || !taskId || fieldType === undefined) {
+    return undefined;
```
amyyeung17, line 937: This is a super minor thought. Would it make sense to check if the arguments are present before validating their values? I don’t see any edge cases the current order misses, but it just caught my attention.

amyyeung17 (approved):

maximilianoertel (commented):
@richford are the changes introduced here in the score report service sufficient to apply the changes to both the existing and the new score report page? Seems so, but just double checking. 

Also, separately, should the `SCORE_FIELD_MIGRATION_GUIDE` be in the docs instead of the helpers directory?

apps/dashboard/src/helpers/reports.js
richford: Thanks @amyyeung17 , I responded to this in the latest commit.

richford:
> @richford are the changes introduced here in the score report service sufficient to apply the changes to both the existing and the new score report page? Seems so, but just double checking.
> 
> Also, separately, should the `SCORE_FIELD_MIGRATION_GUIDE` be in the docs instead of the helpers directory?

@maximilianoertel, yes, this should be sufficient in both the existing and new score report page. I also added jsdoc comments to see the migration guide.

kellyel (approved):

apps/dashboard/src/helpers/reports.js
```diff
@@ -698,71 +699,262 @@ export function getTagColor(supportLevel) {
...
+ * @returns {*} The score value or undefined if not found
+ */
+export function getScoreValue(scoresObject, taskId, grade, fieldType) {
+  if (!scoresObject || !taskId || fieldType === undefined) {
+    return undefined;
+  }
+
+  if (!ALLOWED_SCORE_FIELD_TYPES.includes(fieldType)) {
+    throw new Error(`Invalid fieldType. Expected one of ${ALLOWED_SCORE_FIELD_TYPES.join(', ')}, but got ${fieldType}`);
   }
```
ksmontville, line 941: Is this check necessary, since it is already handled in `resolveFieldName`? The error handling mechanism in `resolveFieldName` would never get a chance to throw, since this statement would throw first. 

We might be able to instead wrap the fuction logic in a try/catch clause. If `resolveFieldName` throws an error, we would let it bubble up into the catch block and log it, handle it, re-throw, return, or whatever else might be appropriate.

apps/dashboard/src/helpers/reports.js
```diff
@@ -698,71 +699,262 @@ export function getTagColor(supportLevel) {
...
+      new: (grade) => (grade < 6 ? 'tosrecSS' : 'sprStandardScore'),
+      legacy: (grade) => (grade < 6 ? 'tosrecSS' : 'sprStandardScore'),
+    },
+    rawScore: {
+      new: 'sreScore',
+      legacy: 'sreScore',
+    },
+  },
+  letter: {
+    percentile: {
```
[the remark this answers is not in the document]
richford: 💯 I completely agree that this is confusing and inconsistent. In my mind, percent correct is a type of raw score and we should pick between `totalCorrect` and `totalPercentCorrect` for the raw score. However, I'd like to argue that this standardization, while necessary, is beyond the score of this PR. the confusion between percentile and percent correct already exists in the current codebase (see [these lines](https://github.com/yeatmanlab/roar-dashboard/blob/318f8446bc6a786086c0e3216dbcdb1ffbea2232/apps/dashboard/src/helpers/reports.js#L747)). And the intent of this PR is to set up the infrastructure to update score field names in a backwards compatible way as they are updated in the assessments themselves. So I'd prefer to solve the percentile vs. percentCorrect issue in a separate PR.

What do you think @ksmontville?

amyyeung17 (approved):

PR #1408, opened 2025-10-12
Redesign Login Page

Emily-ejag:
## ✅ Objective
Refactor the Sign-In flow to align with the new identifier-first design and platform architecture standards, including composable extraction, UI system updates, and step-based structure.

---

### Design spec
- Implement the login screen according to the Figma flow from @maximilianoertel
- Ensure alignment, padding, spacing, typography, and layout match the design

### Components
- Wrap sign-in UI in shared `<Card>` component
- Replace `PvMessage` with `<Alert>` component (destructive variant for errors)
- Split provider UI into two sections:
  - Generic providers (always visible at email stage)
  - Scoped providers (after identifier check)

### Style guidelines
- Remove inline styles
- Use Tailwind utility classes for spacing, layout, alignment, and state
- Maintain responsive layout

---

### Container / Presentational pattern
- Convert sign-in flow to container + presentational component pattern
- `SignIn.vue` acts only as orchestrator

### Extract business logic into composables
Create the following composables:

| Composable | Responsibility |
|-----------|----------------|
| `useSignInForm` | local form state, validation, flow toggling |
| `useAuth` | API calls, success/error handling, session, redirect |
| `useProviders` | provider discovery & mapping: generic vs scoped |
| `useCapsLock` | caps lock state + handler |

- UI components should only receive state & events from composables

### Folder restructure (identifier-first pattern)
Adopt step-based folder structure matching auth flow:

```

src/
└─ containers/
   └─ SignIn/
      ├─ SignIn.vue
      ├─ composables/
      │  ├─ useSignInForm.js
      │  ├─ useProviders.js
      │  ├─ useAuth.js
      │  └─ useCapsLock.js
      └─ components/
         ├─ SignInCard/
         │  ├─ SignInCard.vue
         │  └─ index.js
         ├─ SignInHeader/
         │  ├─ SignInHeader.vue
         │  └─ index.js
         ├─ SignInError/
         │  ├─ SignInError.vue
         │  └─ index.js
         ├─ identifier-step/
         │  ├─ IdentifierInput/
         │  │  ├─ IdentifierInput.vue
         │  │  └─ index.js
         │  └─ GenericProviders/
         │     ├─ GenericProviders.vue
         │     └─ index.js
         ├─ authentication-step/
         │  ├─ PasswordInput/
         │  │  ├─ PasswordInput.vue
         │  │  └─ index.js
         │  ├─ ScopedProviders/
         │  │  ├─ ScopedProviders.vue
         │  │  └─ index.js
         │  ├─ MagicLinkSuccess/
         │  │  ├─ MagicLinkSuccess.vue
         │  │  └─ index.js
         │  └─ MagicLinkBackButton/
         │     ├─ MagicLinkBackButton.vue
         │     └─ index.js
         └─ shared/
            ├─ SignedInEmailChip/
            │  ├─ SignedInEmailChip.vue
            │  └─ index.js
            └─ ProviderButton/
               ├─ ProviderButton.vue
               └─ index.js
```
---

## 🔐 Auth Flow Requirements
- Identifier-first login:
  1. User enters email/username
  2. System checks providers
  3. Show correct path:
     - Password
     - Scoped provider options
     - Magic link flow

- Include caps-lock detection indicator

---

### Unit tests
- Composables fully covered (`useAuth`, `useSignInForm`, `useProviders`, `useCapsLock`)
- Validation, state transitions, and branch logic tests

### E2E tests
Cover flows:
- Email → Password → Success
- Email → Scoped provider → Success
- Generic provider login
- Magic-link path
- Caps-lock indicator behavior
- Invalid credentials error

---

## 🧪 QA Requirements
- All login paths manually validated
- Session storage & redirect verified
- Autofill & Enter key submission behavior confirmed
- Regression check for student/teacher/parent roles

---

## 📌 Acceptance Criteria

- Login screen matches Figma
- One consolidated PR containing:
  - Composable extraction
  - Component refactor
  - Card + Alert adoption
  - Folder and component restructure
- All auth flows function
- Tailwind only, no inline CSS
- Unit + E2E coverage added
- QA completed before merge

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist


richford (changes requested):
Hi @Emily-ejag . This is looking REALLY good. I noticed a few things:

- When I hit entire instead of pressing continue after putting in my email, it gives me "Incorrect username/email or password"
- I can't see the cute-lion image next to my email on the second screen.
- My personal email is not activated for email link (I think). When I clicked on "Request email link" the button didn't do anything and the console displayed this error:
```
Uncaught (in promise) FirebaseError: Firebase: Error (auth/operation-not-allowed).
    at Rm (index-5ca0aadd.js:499:41)
    at Dn (index-5ca0aadd.js:470:11)
    at mI (index-5ca0aadd.js:913:17)
    at async iU (index-5ca0aadd.js:5577:9)
```
I think we want to show the "Email sent" page for all email users when they click on "Request email link"

maximilianoertel (changes requested):
This is a great start and excited to see the auth flow getting redesigned! 

Before going into a more detailed review of the actual underlying logic, I have a few notes that I'm hoping we can address: 

- Just like with any refactors on the dashboard, this PR is the right opportunity to start refactoring the existing page(s)/components into a Container/Presentational pattern. Happy to walk you through how that pattern works if helpful.

- The layout doesn’t quite match the provided Figma designs in regards to things such as alignment, spacing, padding, etc. Happy to show you how to inspect these details in Figma to make sure we’re consistent with the design spec.

- More generally, we want to avoid using inline style blocks and instead rely on TailwindCSS classes for styling as that’s the direction we’re standardizing across the app.

- There is room for greater test coverage, both on for unit and E2E tests.


Let me know what you think, but once we address these points, happy to dive deeper into the review!

richford (commented):
@Emily-ejag, this is a great start to the container pattern adoption.

I like the way you decomposed the sign-in UI into small presentational components (email input, password block, provider list, chip, header, magic-link back). Next steps should be to push logic and side effects into composables so the components stay purely presentational and easier to test.

### What stays in parent SignIn.vue

Keep the parent SignIn.vue container responsible for

- orchestration and state: current email, password, showPasswordField, multipleProviders, availableProviders, emailLinkSent, invalid, isUsername, hideProviders, etc.
- wiring to auth services: calling check-providers, submit, magic link, forgot-password, third-party SSO flows.
- business-level side effects: navigation on success, analytics events, global notifications, and storing tokens/session.

### What should move into composables:

- useSignInForm (core form state + validation):
  - Exposes: email, password, invalid, isUsername, showPasswordField, emailLinkSent, multipleProviders, hideProviders, availableProviders
  - Methods: checkProviders(email), submit(), sendMagicLink(), forgotPassword(), clearEmail(), backToPassword()
  - Handles: form validation rules (email format vs username), toggling flows (email -> password, email link state), and local error state
- useAuth (auth API wrapper)
  - Methods: signInWithPassword(email, password), sendMagicLink(email), checkProvidersApi(email), signInWithProvider(providerId)
  - Handles API calls, success/error normalization, token storage, redirect handling
- useProviders (provider availability)
  - Exposes: availableProviders (reactive), showGenericProviders, showScopedProviders, reloadProviders()
  - Handles: deciding which provider buttons to show (scoping rules, feature flags, remote config)
- useCapsLock (UI helper)
  - Exposes: capsLockEnabled, onKeydown handler (or checkForCapsLock util)
- useSignInI18n (optional thin wrapper)
  - Exposes keys or computed text for title/subtitle depending on state (or simply leave this logic in SignInHeaderCopy, but if the mapping gets more complex move it)
- useUiFeedback (optional)
  - Exposes: showError(messageKey), showSuccess(...), pushPvMessage or toasts

### Mapping from current code to composables/components

- SignedInEmailChip.vue: stays as presentational.
- SignInEmailInput.vue: move checkForCapsLock to useCapsLock composable and replace emitted event name to consistent naming: 'enter'.
- SignInPasswordBlock.vue: move checkForCapsLock here to useCapsLock; emit submit/forgot/magic-link unchanged.
- SignInProvidersList.vue: provider visibility props come from useProviders.
- SignInHeaderCopy.vue: can accept title/subtitle as props computed by the parent via useSignInForm.
- SignIn.vue (container): should be mostly props wiring + useSignInForm + useAuth/useProviders hooks. The container will implement the emit handlers:
  - onCheckProviders -> useSignInForm.checkProviders(email) -> update showPasswordField / multipleProviders / availableProviders
  - onSubmit -> useSignInForm.submit() which calls useAuth.signInWithPassword
  - onMagicLink -> useSignInForm.sendMagicLink()
  - onAuthX -> useAuth.signInWithProvider('clever'|'classlink'|'nycps'|'google')
  - onForgotPassword -> useSignInForm.forgotPassword()

### Testing

Unit-test the composables (pure logic) and test presentational components separately. Windsurf or other AI tools are quite good at generating these tests given the examples already in the repo.

maximilianoertel (changes requested):
Great work on the initial extraction!  From here on, again, before diving into actual code / details, a couple of comments:  



#### Card component

The sign-in form should be wrapped in a reusable Card component, not just a <section> with utility classes. This helpers us keep consistency for things like border, background, padding, rounded corners, etc. 

```
  <Card>
    <SignInForm ... />
  </Card>
```

---

#### Alert component
We recently introduced a new Alert component which would be the perfect fit here instead of the `PvMessage` component used in some of the sign-in components. 

```diff
-  <PvMessage icon="pi pi-exclamation-circle" severity="error">
-    {{ $t('authSignIn.incorrectEmailOrPassword') }}
-  </PvMessage>
+  <Alert variant="destructive">
+    <AlertTitle>{{ $t('authSignIn.error') }}</AlertTitle>
+    <AlertDescription>{{ $t('authSignIn.incorrectEmailOrPassword')
+   }}</AlertDescription>
+  </Alert>
```

--- 

#### Folder structure

The current flat structure in `/containers/SignIn/components/` works but doesn't communicate the sign-in flow's architecture. This is an identifier-first authentication pattern with distinct steps:

- Step 1: Identifier Entry → User enters email, sees generic SSO options
- Step 2: Authentication Method → Based on identifier, shows: password, scoped SSO providers, or magic link

The current structure is flat:

```
  containers/SignIn/components/
    ├── SignInEmailInput.vue
    ├── SignInPasswordBlock.vue
    ├── SignInProvidersList.vue  ← handles both generic AND scoped
    ├── SignedInEmailChip.vue
    ├── SignInMagicLinkBackButton.vue
    └── SignInHeaderCopy.vue
```
The problem here is that: 
- `SignInEmailInput` and `SignInPasswordBlock` are siblings, but they're shown in different steps
- `SignInProvidersList `handles two distinct UI states (generic vs scoped)
- No visual indication of the authentication flow in the folder structure

Instead, I'd recommend a hierarchical structure which mirrors the user flow: 
- Folder structure matches UX journey
- Clearer ownership: Easy to see which components belong to which step
- Easier maintenance: Adding a new auth method means adding to the correct step folder
- Self-documenting: New devs immediately understand this is identifier-first auth

```
src/
└─ containers/
   └─ SignIn/
      ├─ SignIn.vue
      ├─ composables/
      │  ├─ useSignInForm.js
      │  ├─ useProviders.js
      │  ├─ useAuth.js
      │  └─ useCapsLock.js
      └─ components/
         ├─ SignInCard/
         │  ├─ SignInCard.vue
         │  └─ index.js
         ├─ SignInHeader/
         │  ├─ SignInHeader.vue
         │  └─ index.js
         ├─ SignInError/
         │  ├─ SignInError.vue
         │  └─ index.js
         ├─ identifier-step/
         │  ├─ IdentifierInput/
         │  │  ├─ IdentifierInput.vue
         │  │  └─ index.js
         │  └─ GenericProviders/
         │     ├─ GenericProviders.vue
         │     └─ index.js
         ├─ authentication-step/
         │  ├─ PasswordInput/
         │  │  ├─ PasswordInput.vue
         │  │  └─ index.js
         │  ├─ ScopedProviders/
         │  │  ├─ ScopedProviders.vue
         │  │  └─ index.js
         │  ├─ MagicLinkSuccess/
         │  │  ├─ MagicLinkSuccess.vue
         │  │  └─ index.js
         │  └─ MagicLinkBackButton/
         │     ├─ MagicLinkBackButton.vue
         │     └─ index.js
         └─ shared/
            ├─ SignedInEmailChip/
            │  ├─ SignedInEmailChip.vue
            │  └─ index.js
            └─ ProviderButton/
               ├─ ProviderButton.vue
               └─ index.js
```


#### Composables

As pointed out by Adam already, the `/pages/SignIn.vue`  container is ~337 lines and contains significant business logic that should be extracted into composables. 

- Form state (email, password, invalid, showPasswordField, etc.)→ `useSignInForm`
- Provider discovery & normalization (checkAvailableProviders, normalizeProviders) → `useProviders`
- All auth flows (Clever, ClassLink, NYCPS, Google, email/password, magic link) → `useAuth`
- User claims fetching → `useAuth`
- Navigation/routing → `useAuth`
- Error handling → `useSignInForm` + `useAuth`

In other words, the page component should be mostly handling orchestration, not business logic.

apps/dashboard/src/containers/SignIn/components/identifier-step/IdentifierInput/IdentifierInput.vue
```diff
@@ -0,0 +1,32 @@
...
+import PvInputText from 'primevue/inputtext';
+import PvMessage from 'primevue/message';
+import { useCapsLock } from '../../../composables/useCapsLock';
+
+const props = defineProps({ modelValue: { type: String, default: '' }, invalid: { type: Boolean, default: false } });
+defineEmits(['update:modelValue', 'enter']);
+
+const { checkForCapsLock } = useCapsLock();
+const modelValue = props.modelValue; // expose for template readability
+const invalid = props.invalid;
```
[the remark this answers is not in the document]
Emily-ejag: 100%

apps/dashboard/src/pages/SignIn.vue
```diff
@@ -1,455 +1,341 @@
...
-        incorrect.value = true;
-        if (['auth/user-not-found', 'auth/wrong-password'].includes(e.code)) {
-          return;
+        const errorCode = e.code;
+        if (errorCode === 'auth/email-already-in-use') {
+          spinner.value = false;
         } else {
-          throw e;
+          spinner.value = false;
         }
```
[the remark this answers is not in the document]
Emily-ejag: sure

apps/dashboard/src/containers/SignIn/components/SignInCard/SignInCard.vue
```diff
@@ -0,0 +1,5 @@
+<template>
+  <div class="bg-white border border-200 rounded-2xl p-4 md:p-5 shadow-1" :class="rootClass" v-bind="restAttrs">
```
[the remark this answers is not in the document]
Emily-ejag: Sure

apps/dashboard/src/containers/SignIn/components/authentication-step/PasswordInput/PasswordInput.vue
```diff
@@ -0,0 +1,50 @@
...
+    <div class="mt-2 flex w-full align-items-center justify-content-between">
+      <small class="text-sm text-400 cursor-pointer font-semibold hover:text-primary" @click="$emit('forgot-password')">
+        {{ $t('authSignIn.forgotPassword') }}
+      </small>
+      <small
+        v-if="!isUsername"
+        class="text-sm text-400 cursor-pointer font-semibold hover:text-primary"
+        @click="$emit('magic-link')"
+      >
+        {{ $t('authSignIn.magicLink') || 'Magic Link' }}
```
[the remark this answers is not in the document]
Emily-ejag: true

apps/dashboard/src/containers/SignIn/components/authentication-step/ScopedProviders/ScopedProviders.vue
```diff
@@ -0,0 +1,72 @@
...
+import CLEVER_LOGO from '@/assets/provider-clever-logo.svg';
+import CLASSLINK_LOGO from '@/assets/provider-classlink-logo.png';
+import NYCPS_LOGO from '@/assets/provider-nycps-logo.jpg';
+
+const props = defineProps({
+  availableProviders: { type: Array, default: () => [] },
+});
+
+const availableProviders = props.availableProviders;
+
```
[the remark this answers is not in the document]
richford: Again, relatively minor but I agree with Copilot here.

richford (changes requested):
@Emily-ejag this is fantastic! I have one question about the auth flow, one change request about an unused component, and three minor change requests that are merely me agreeing with Copilot suggestions. I'll take those in reverse order:

## 1. Copilot requests

The change requests are inline and if you agree, you can just add Copilot's suggestions to a batch and commit all three together.

## 2. Unused component

The `MagicLinkSuccess` component appears unimported and unused. Can we get rid of it?

## 3. Auth flow consistency question

The larger auth flow question has to do with what the user sees when they click forgot password vs what they see when they request an email link. There seems to be an inconsistent UI between these two flows. That isn't necessarily a bad thing if it's intentional. So I just want to ask about it and understand the reasoning.

When they click "Forgot password?" they see this

Screenshot 2025-11-21 at 5 59 33 AM

It uses the `SuccessAlert` component, displaying new information in an alert below the email chip. Then it waits five seconds before resetting the sign-in UI.

But for the email link, they see this

Screenshot 2025-11-21 at 5 59 12 AM

The new information text is presented above the email chip rather than below and not in an alert component. And the UI is not reset.

I realize that for the password reset flow, the user will likely return to the same tab and attempt to log in again with their new password, while for the email link flow, they will click on a link in their email that opens a new tab, so the current tab is basically abandoned. That is a good justification for the difference in UI reset behavior. What about the difference in the use of an alert component and the location of new information? This last part is not a change request but more of a comment/question. Can you comment on the difference in flows?

apps/dashboard/src/containers/SignIn/components/authentication-step/MagicLinkSuccess/MagicLinkSuccess.vue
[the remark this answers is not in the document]
Emily-ejag: I don’t think we need MagicLinkSuccess.vue anymore — the success state is now managed directly in the SignIn view using emailLinkSent. Since the component isn’t referenced anywhere, I’ll remove it to avoid dead/unused code.

apps/dashboard/src/containers/SignIn/components/authentication-step/ScopedProviders/ScopedProviders.vue
```diff
@@ -15,7 +15,7 @@
 
     <!-- Clever -->
     <PvButton
-      v-if="availableProviders.includes('clever')"
+      v-if="props.availableProviders.includes('clever')"
```
richford: Same here

```suggestion
      v-if="availableProviders.includes('clever')"
```

richford (changes requested):

apps/dashboard/src/containers/SignIn/components/authentication-step/MagicLinkBackButton/MagicLinkBackButton.test.js
maximilianoertel: Great to see tests added! However, most of these assertions check internal structure (Tailwind classes, specific elements, component names). That makes the tests brittle and doesn’t give us much behavioural value.

For a UI component like this one, it’s better to focus on behaviour (e.g., emitting `back-to-password` and showing the correct label) and avoid testing implementation details. If we want to ensure the overall markup stays consistent, a single snapshot test is more appropriate than many granular class checks.


Something along those lines would be more appropriate: 
```
it('renders correctly', () => {
  const wrapper = mount(MagicLinkBackButton);
  expect(wrapper.html()).toMatchSnapshot();
});

it('emits "back-to-password" when clicked', async () => {
  const wrapper = mount(MagicLinkBackButton);

  await wrapper.find('button').trigger('click');

  expect(wrapper.emitted('back-to-password')).toBeTruthy();
});
```

apps/dashboard/src/containers/SignIn/components/authentication-step/PasswordInput/PasswordInput.test.js
```diff
@@ -0,0 +1,242 @@
...
+        show: true,
+        isUsername: false,
+        invalid: false,
+        password: '',
+      },
+    });
+
+    const passwordInput = wrapper.findComponent({ name: 'Password' });
+    expect(passwordInput.attributes('data-cy')).toBe('sign-in__password');
+  });
```
maximilianoertel: I'd remove these tests as we don’t need a dedicated test case just to check that the `data-cy` attribute exists. Since this attribute is primarily for testing and doesn’t affect behaviour, it’s not worth maintaining a separate assertion for it.

```suggestion
```

apps/dashboard/src/containers/SignIn/components/authentication-step/ScopedProviders/ScopedProviders.test.js
```diff
@@ -0,0 +1,258 @@
...
+      },
+    });
+
+    const images = wrapper.findAll('img');
+    images.forEach((img) => {
+      expect(img.classes()).toContain('flex');
+      expect(img.classes()).toContain('p-1');
+      expect(img.classes()).toContain('w-1');
+    });
+  });
```
maximilianoertel: Similar to above, these assertions are directly testing implementation details instead of behaviour. 

```suggestion
```

apps/dashboard/src/containers/SignIn/components/authentication-step/PasswordInput/PasswordInput.test.js
```diff
@@ -0,0 +1,242 @@
...
+      props: {
+        show: true,
+        isUsername: false,
+        invalid: false,
+        password: '',
+      },
+    });
+
+    const links = wrapper.findAll('small');
+    expect(links.length).toBeGreaterThan(0);
```
maximilianoertel: Could you please clarify why we're asserting the existing of the string "small"?

apps/dashboard/src/containers/SignIn/components/authentication-step/ScopedProviders/ScopedProviders.test.js
```diff
@@ -0,0 +1,258 @@
...
+    const wrapper = mount(ScopedProviders, {
+      props: {
+        availableProviders: ['google'],
+      },
+    });
+
+    const buttons = wrapper.findAllComponents({ name: 'Button' });
+    const googleButton = buttons.find((btn) => btn.attributes('data-cy') === 'sign-in__google-sso');
+    expect(googleButton).toBeDefined();
+  });
```
maximilianoertel: We generally try to avoid using data-cy in unit tests because it’s meant for Cypress E2E, and coupling both layers to the same selectors can get messy long-term. Ideally, we would use neutral data-testid attributes (via PrimeVue pass-through) so unit tests and E2E tests can share stable selectors without being tied to each other’s conventions.

Given the current state of the codebase, I'd say we should use a dedicated `data-testid` so we can change E2E selectors independently of unit tests when needed.

Also, this test only checks that the Google button exists, but the behaviour we actually care about is that when availableProviders = ['google'], the Google button is rendered _and other provider buttons are not_.


```suggestion
  it('renders only Google when availableProviders contains google', () => {
  const wrapper = mount(ScopedProviders, {
    props: {
      availableProviders: ['google'],
    },
  });

  const googleButton = wrapper.find('[data-testid="signin__provider-google"]');
  expect(googleButton.exists()).toBe(true);

  // Assert that other providers are not rendered
  expect(wrapper.find('[data-testid="signin__provider-github"]').exists()).toBe(false);
  expect(wrapper.find('[data-testid="signin__provider-microsoft"]').exists()).toBe(false);
});
  });
```

The same applies to the test cases below and other files that apply the same pattern.

apps/dashboard/src/containers/SignIn/components/authentication-step/ScopedProviders/ScopedProviders.vue
maximilianoertel: There is quite a bit of code duplication going on here that we could simplify. Here's an example which uses the `ProviderButton` component you've introduced in this PR: 

```vue
<template>
    <div class="flex flex-column w-full align-content-center justify-content-center">
      <ProviderButton
        v-for="provider in visibleProviders"
        :key="provider.id"
        :label="$t('authSignIn.signInWith') + ' ' + provider.name"
        :img-src="provider.logo"
        :data-cy="`sign-in__${provider.id}-sso-scoped`"
        btn-class="provider-button m-1 surface-0 border-200 border-1 border-round-md hover:border-primary
  hover:surface-ground"
        @click="$emit(`auth-${provider.id}`)"
      />
    </div>
  </template>

  <script setup>
  import ProviderButton from '@/containers/SignIn/components/shared/ProviderButton/ProviderButton.vue';
  import GOOGLE_LOGO from '@/assets/provider-google-logo.svg';
  import CLEVER_LOGO from '@/assets/provider-clever-logo.svg';
  import CLASSLINK_LOGO from '@/assets/provider-classlink-logo.png';
  import NYCPS_LOGO from '@/assets/provider-nycps-logo.jpg';
  import { computed } from 'vue';

  const props = defineProps({
    availableProviders: { type: Array, default: () => [] },
  });

  defineEmits(['auth-google', 'auth-clever', 'auth-classlink', 'auth-nycps']);

  const PROVIDERS = [
    { id: 'google', name: 'Google', logo: GOOGLE_LOGO },
    { id: 'clever', name: 'Clever', logo: CLEVER_LOGO },
    { id: 'classlink', name: 'ClassLink', logo: CLASSLINK_LOGO },
    { id: 'nycps', name: 'NYCPS', logo: NYCPS_LOGO },
  ];

  const visibleProviders = computed(() =>
    PROVIDERS.filter((p) => props.availableProviders.includes(p.id))
  );
  </script>
```

Even better would be to extract the hard-coded provider config into a constant in `src/constants/sso-providers.js`

```js
import GOOGLE_LOGO from '@/assets/provider-google-logo.svg';
import CLEVER_LOGO from '@/assets/provider-clever-logo.svg';
import CLASSLINK_LOGO from '@/assets/provider-classlink-logo.png';
import NYCPS_LOGO from '@/assets/provider-nycps-logo.jpg';

export const SSO_PROVIDERS = [
  { id: 'google', name: 'Google', logo: GOOGLE_LOGO },
  { id: 'clever', name: 'Clever', logo: CLEVER_LOGO },
  { id: 'classlink', name: 'ClassLink', logo: CLASSLINK_LOGO },
  { id: 'nycps', name: 'NYCPS', logo: NYCPS_LOGO },
];
```

apps/dashboard/src/containers/SignIn/components/identifier-step/GenericProviders/GenericProviders.test.js
```diff
@@ -0,0 +1,148 @@
...
+  it('should have correct container CSS classes', () => {
+    const wrapper = mount(GenericProviders);
+
+    const container = wrapper.find('div');
+    expect(container.classes()).toContain('flex');
+    expect(container.classes()).toContain('flex-column');
+    expect(container.classes()).toContain('w-full');
+    expect(container.classes()).toContain('align-content-center');
+    expect(container.classes()).toContain('justify-content-center');
+  });
```
maximilianoertel:
```suggestion
```

apps/dashboard/src/containers/SignIn/components/identifier-step/IdentifierInput/IdentifierInput.test.js
```diff
@@ -0,0 +1,197 @@
...
+    const wrapper = mount(IdentifierInput, {
+      props: {
+        modelValue: '',
+        invalid: false,
+      },
+    });
+
+    const input = wrapper.findComponent({ name: 'InputText' });
+    expect(input.attributes('data-cy')).toBe('sign-in__username');
+  });
```
maximilianoertel:
```suggestion
```

apps/dashboard/src/containers/SignIn/components/identifier-step/IdentifierInput/IdentifierInput.vue
```diff
@@ -0,0 +1,30 @@
+<template>
+  <PvFloatLabel class="mt-4">
+    <PvInputText
+      id="email"
+      :class="['w-full', 'border-200', { 'p-invalid': invalid }]"
+      :model-value="modelValue"
+      aria-describedby="email-error"
```
maximilianoertel: Where is the `email-error` container you're referencing here?

apps/dashboard/src/containers/SignIn/components/shared/SignInEmailChip/SignInEmailChip.test.js
```diff
@@ -0,0 +1,129 @@
...
+describe('SignInEmailChip.vue', () => {
+  it('should render the component', () => {
+    const wrapper = mount(SignInEmailChip, {
+      props: {
+        email: 'test@example.com',
+      },
+    });
+
+    expect(wrapper.exists()).toBe(true);
+  });
```
maximilianoertel: As above, this should likely be a snapshot test

apps/dashboard/src/containers/SignIn/components/shared/SignInEmailChip/SignInEmailChip.test.js
```diff
@@ -0,0 +1,129 @@
...
+  it('should have removable chip', () => {
+    const wrapper = mount(SignInEmailChip, {
+      props: {
+        email: 'test@example.com',
+      },
+    });
+
+    const chip = wrapper.findComponent({ name: 'Chip' });
+    expect(chip.props('removable')).toBe(true);
+  });
```
maximilianoertel:
```suggestion
```

apps/dashboard/src/containers/SignIn/components/shared/SignInEmailChip/SignInEmailChip.test.js
```diff
@@ -0,0 +1,129 @@
...
+  it('should handle email with special characters', () => {
+    const wrapper = mount(SignInEmailChip, {
+      props: {
+        email: 'test+tag@example.co.uk',
+      },
+    });
+
+    const chip = wrapper.findComponent({ name: 'Chip' });
+    expect(chip.props('label')).toBe('test+tag@example.co.uk');
+  });
```
maximilianoertel:
```suggestion
```

apps/dashboard/src/containers/SignIn/components/SignInCard/SignInCard.test.js
```diff
@@ -0,0 +1,51 @@
+import { describe, it, expect } from 'vitest';
+import { mount } from '@vue/test-utils';
+import SignInCard from './SignInCard.vue';
+
+describe('SignInCard.vue', () => {
+  it('should render the component', () => {
+    const wrapper = mount(SignInCard);
+
+    expect(wrapper.exists()).toBe(true);
+  });
```
maximilianoertel: As above, should likely be a snapshot test

apps/dashboard/src/containers/SignIn/components/SignInCard/SignInCard.test.js
```diff
@@ -0,0 +1,51 @@
...
+    const wrapper = mount(SignInCard);
+
+    const card = wrapper.findComponent({ name: 'Card' });
+    expect(card.classes()).toContain('bg-white');
+    expect(card.classes()).toContain('border');
+    expect(card.classes()).toContain('border-round-xs');
+    expect(card.classes()).toContain('border-200');
+    expect(card.classes()).toContain('p-3');
+    expect(card.classes()).toContain('shadow-1');
+  });
```
maximilianoertel:
```suggestion
```

apps/dashboard/src/containers/SignIn/components/SignInCard/SignInCard.test.js
```diff
@@ -0,0 +1,51 @@
...
+
+  it('should render slot content', () => {
+    const wrapper = mount(SignInCard, {
+      slots: {
+        default: '<p>Test content</p>',
+      },
+    });
+
+    expect(wrapper.html()).toContain('Test content');
+  });
```
maximilianoertel: Fine as it is, but could also be a snapshot test to align with the above

apps/dashboard/src/containers/SignIn/components/SignInCard/SignInCard.test.js
```diff
@@ -0,0 +1,51 @@
...
+  it('should render multiple slot elements', () => {
+    const wrapper = mount(SignInCard, {
+      slots: {
+        default: '<p>First</p><p>Second</p>',
+      },
+    });
+
+    expect(wrapper.html()).toContain('First');
+    expect(wrapper.html()).toContain('Second');
+  });
```
maximilianoertel:
```suggestion
```

apps/dashboard/src/components/LanguageSelector.vue
```diff
@@ -1,41 +1,96 @@
...
+        href="#"
+        class="text-400 w-full inline-block text-left text-sm pt-2 rounded-md hover:text-primary"
+        :aria-label="$t('navBar.changeLanguage')"
+        aria-haspopup="menu"
+        :aria-expanded="menuVisible ? 'true' : 'false'"
+        @click.prevent="toggleMenu"
+      >
+        <span>{{ currentLanguageLabel }}</span>
+        <i class="pi pi-chevron-down text-sm pl-2" aria-hidden="true" />
+      </a>
```
maximilianoertel: This should be a `<button>` instead of an `<a>` because the element doesn’t perform navigation — it triggers an action (toggleMenu). Using an anchor (`<a href="#">`) for something that isn’t a link is incorrect semantically and can, in certain circumstances, cause accessibility issues

apps/dashboard/src/containers/SignIn/components/identifier-step/IdentifierInput/IdentifierInput.vue
```diff
@@ -0,0 +1,30 @@
...
+<script setup>
+import PvFloatLabel from 'primevue/floatlabel';
+import PvInputText from 'primevue/inputtext';
+import PvMessage from 'primevue/message';
+import { useCapsLock } from '../../../composables/useCapsLock';
+
+defineProps({ modelValue: { type: String, default: '' }, invalid: { type: Boolean, default: false } });
+defineEmits(['update:modelValue', 'enter']);
+
+const { checkForCapsLock } = useCapsLock();
```
maximilianoertel: No need to use the `useCapsLock` composable here as identifiers such as emails and usernames should be case insensitive

apps/dashboard/src/containers/SignIn/components/shared/ProviderButton/ProviderButton.test.js
```diff
@@ -0,0 +1,245 @@
...
+        imgSrc: '',
+        icon: '',
+        btnClass: '',
+        dataCy: '',
+      },
+    });
+
+    const span = wrapper.find('span');
+    expect(span.text()).toBe('Sign in with Google');
+  });
```
maximilianoertel: Those are replaced by the snapshot test 

```suggestion
```

apps/dashboard/src/containers/SignIn/components/shared/ProviderButton/ProviderButton.vue
```diff
@@ -0,0 +1,20 @@
...
+</template>
+
+<script setup>
+import PvButton from 'primevue/button';
+
+defineProps({
+  label: { type: String, default: '' },
+  imgSrc: { type: String, default: '' }, // provider logos
+  icon: { type: String, default: '' },
+  btnClass: { type: String, default: '' },
```
maximilianoertel: I'd like to remove this prop, as similar to a previous comment, style overrides via classes shouldn't be allowed in these component as this opens the door to design inconsistencies. In other words, if a single component needs to adapt to different style variants, those should be declared within the component – a good example of that is the `Alert` component.

apps/dashboard/src/containers/SignIn/components/shared/ProviderButton/ProviderButton.vue
```diff
@@ -0,0 +1,20 @@
+<template>
+  <PvButton :class="['w-full p-2', btnClass]" type="button" :data-cy="dataCy" @click="$emit('click')">
+    <img v-if="imgSrc" :src="imgSrc" alt="" class="mr-2" style="width: 18px; height: 18px; object-fit: contain" />
+    <i v-else-if="icon" :class="['mr-2', icon]" aria-hidden="true"></i>
```
maximilianoertel: Why does this component support icons? I believe the provider buttons only need to support the provider logos?

apps/dashboard/src/containers/SignIn/SignIn.test.js
```diff
@@ -0,0 +1,181 @@
...
+    successEmail: '',
+  };
+
+  it('should render the component', () => {
+    const wrapper = mount(SignIn, {
+      props: defaultProps,
+    });
+
+    expect(wrapper.exists()).toBe(true);
+  });
```
maximilianoertel: As above, likely best to be a snapshot test for consistency

apps/dashboard/src/containers/SignIn/components/SignInError/SignInError.test.js
```diff
@@ -0,0 +1,189 @@
...
+      props: {
+        show: true,
+        title: 'Error',
+        description: 'Error description text',
+      },
+    });
+
+    const alertDescription = wrapper.findComponent({ name: 'AlertDescription' });
+    expect(alertDescription.text()).toContain('Error description text');
+  });
```
maximilianoertel:
```suggestion
```

apps/dashboard/src/containers/SignIn/components/SignInError/SignInError.test.js
```diff
@@ -0,0 +1,189 @@
...
+    });
+
+    let alertDescription = wrapper.findComponent({ name: 'AlertDescription' });
+    expect(alertDescription.text()).toContain('Original description');
+
+    await wrapper.setProps({ description: 'Updated description' });
+
+    alertDescription = wrapper.findComponent({ name: 'AlertDescription' });
+    expect(alertDescription.text()).toContain('Updated description');
+  });
```
maximilianoertel: Similar to previous comments, these are testing Vue internals as we're not making any modifications/transformations to the prop values, hence should be removed


```suggestion
```

apps/dashboard/src/assets/styles/routes/signin.scss
```diff
@@ -4,6 +4,69 @@ body.page-signin {
...
+  display: flex;
+  justify-content: center;
+  margin-bottom: 0.25rem;
+}
+.signin-footer {
+  width: 100%;
+  max-width: 480px;
+  margin: 0.5rem auto 0;
+  padding: 0 0.5rem;
+}
```
maximilianoertel: Together with [#]
```suggestion
```

apps/dashboard/src/pages/SignIn.vue
```diff
@@ -1,455 +1,169 @@
...
-            <PvButton
-              class="flex p-1 mr-2 ml-2 w-3 text-black surface-0 border-black-alpha-10 justify-content-center hover:border-primary hover:surface-ground"
-              style="border-radius: 3rem; height: 3rem; color: black"
-              data-cy="sign-in__classlink-sso"
-              @click="authWithSSO(AUTH_SSO_PROVIDERS.CLASSLINK)"
+      </SignInCard>
+
+      <!-- Footer (Language, Privacy, Terms) -->
+      <footer class="signin-footer">
+        <a href="#trouble" class="hidden">{{ $t('pageSignIn.havingTrouble') }}</a>
```
maximilianoertel:
```suggestion
```

apps/dashboard/src/pages/SignIn.vue
```diff
@@ -1,455 +1,169 @@
...
+        <div class="w-full flex">
+          <div class="flex-1">
+            <LanguageSelector />
+          </div>
+          <div class="flex gap-2">
+            <!-- hidding privacy for now until we have a document -->
+            <a
+              v-if="false"
+              :href="TERMS_OF_SERVICE_DOCUMENT_PATH"
+              class="text-400 inline-block text-sm hover:text-primary pt-2"
```
maximilianoertel: For links and buttons to be easily clickable, you usually want to add padding all around. Same applies below

```suggestion
              class="text-400 inline-block text-sm hover:text-primary p-1"
```

apps/dashboard/src/containers/SignIn/components/authentication-step/PasswordInput/PasswordInput.vue
```diff
@@ -0,0 +1,46 @@
...
+
+defineProps({
+  show: { type: Boolean, default: false },
+  isUsername: { type: Boolean, default: false },
+  invalid: { type: Boolean, default: false },
+  password: { type: String, default: '' },
+});
+defineEmits(['update:password', 'forgot-password', 'magic-link', 'submit']);
+
+const { checkForCapsLock } = useCapsLock();
```
maximilianoertel: For the password input, caps lock detection is indeed important. However, while the component imports and ties the `checkForCapsLock` to the keydown event nothing is actually done here to handle? The password input should indicate whenever caps lock is on, ideally through an icon displayed on the right next to the toggle text icon

apps/dashboard/src/containers/SignIn/composables/useSignInForm.test.js
maximilianoertel: The container/presentational pattern is inverted here. Currently:
  - Page has all the logic (3 composables, ~30 reactive values)
  - Container just passes props/events through with no logic

This makes the container a "middle-man" that adds complexity without value. Every new prop or event requires changes in 3 places.

Instead, we want to achieve this in our codebase:
```
pages/SignIn.vue              ← Minimal: just mounts container + page-level concerns (title, layout)
    └── containers/SignIn.vue   ← All composables, state management, business logic
          └── components/*      ← Pure presentational, props in, events out
```

  | Layer      | Should Do                          | Currently Does     |
  |------------|------------------------------------|--------------------|
  | Page       | Routing, page layout, spinner      | All business logic |
  | Container  | Business logic, state, composables | Prop drilling      |
  | Components | Pure UI                            | ✅ Correct          |

  Benefits:
  - No more prop drilling (remove 12 props, 11 events)
  - Container becomes reusable (could work in a modal)
  - Page becomes simple (~20 lines)
  - Each layer has clear responsibility

apps/dashboard/src/containers/SignIn/composables/useAuth.js
```diff
@@ -0,0 +1,215 @@
...
+ *   route: import('vue-router').RouteLocationNormalizedLoaded,
+ *
+ *   // refs from useSignInForm
+ *   email: import('vue').Ref<string>,
+ *   password: import('vue').Ref<string>,
+ *   invalid: import('vue').Ref<boolean>,
+ *   emailLinkSent: import('vue').Ref<boolean>,
+ *   showPasswordField: import('vue').Ref<boolean>,
+ * }} context
+ */
```
maximilianoertel: I'm not sure this JSDoc comment brings much value? Let's either remove it or extend it to contain meaningful information?

apps/dashboard/src/containers/SignIn/composables/useAuth.js
maximilianoertel: Let's please add unit tests for this composable as it's likely the most important one in this container

apps/dashboard/src/containers/SignIn/composables/useAuth.js
```diff
@@ -0,0 +1,215 @@
...
+      });
+  }
+
+  /* ---------- Email/password ---------- */
+  function authWithEmailPassword() {
+    invalid.value = false;
+    const creds = toRaw({
+      email: email.value.includes('@') ? email.value : `${email.value}@roar-auth.com`,
+      password: password.value,
+    });
```
maximilianoertel: The use of `toRaw` seems redundant here?

apps/dashboard/src/containers/SignIn/components/shared/ProviderButton/ProviderButton.vue
Emily-ejag: same as above

apps/dashboard/src/containers/SignIn/components/SuccessAlert/SuccessAlert.vue
```diff
@@ -0,0 +1,18 @@
+<template>
+  <Alert v-if="show" variant="success" class="w-full mb-2">
```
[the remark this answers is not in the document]
Emily-ejag: 100%

apps/dashboard/src/translations/en/us/en-us-componentTranslations.json
```diff
@@ -90,12 +101,20 @@
   },
   "pageSignIn": {
     "havingTrouble": "Having trouble?",
-    "login": "Log in to access your dashboard",
+    "login": "Log in to ROAR to start ROARing",
```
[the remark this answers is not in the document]
Emily-ejag: Yeah I think your suggestion sounds better

Emily-ejag:
Thank you all 😃

PR #1419, opened 2025-10-21
[ENH] Rework AdministrationCard Tree Table Orgs

kellyel:
## Proposed changes
This PR reworks how the dashboard populates the tree table orgs on each AdministrationCard component. Previously, the HomeAdministrator page would retrieve all orgIds for each admin, then when a user clicked 'Show Details', it would query for all of the orgs and assemble the node tree. However this is inefficient as we do not need to store all orgIds for every admin, and we do not need to query for every org once it becomes time to build the node tree. The latter also causes failures for administrations with lots of orgs assigned. 

Instead, we should query for information on an as-needed basis. When the user clicks Show details, we retrieve minimal orgs and build a basic node tree. As the user expands orgs that have children, we query for the children and add them to the node tree. For both of these queries, we use a new cloud function that will provide the correct information. 

Old (recorded via gse-roar-admin-dev)

https://github.com/user-attachments/assets/05db2240-ea8e-48cf-b8c2-7af3ffefe6fe

New (recorded via localhost)

https://github.com/user-attachments/assets/992895c4-ddac-4172-8479-f901bc567814

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist

### DIFF
--- a/apps/dashboard/src/helpers/query/administrations.js
+++ b/apps/dashboard/src/helpers/query/administrations.js
@@ -28,7 +28,7 @@ export function getTitle(item, isSuperAdmin) {
   }
 }
 
-const processBatchStats = async (axiosInstance, statsPaths, batchSize = 5) => {
+const processBatchStats = async (axiosInstance, statsPaths, batchSize = 500) => {
   const batchStatsDocs = [];
   const statsPathChunks = _chunk(statsPaths, batchSize);
   for (const batch of statsPathChunks) {


amyyeung17 (approved):

maximilianoertel (changes requested):
The introduced changes are great! Looking a little closer at the changes made to `onExpand` however raise some concerns as it currently calls the `roarfirekit` method every single time without any sort of caching. Additionally, there seems to some room for improvement when it comes to error handling. 

Ideally, we would use a TanStack query (e.g. `useAdministrationChildOrgsQuery`) here. For that there is two approaches:

- **Option 1:** Single query with a dynamic query key. This is pretty simple to implement and the easiest approach, but means only a single tree.

- **Option 2:** Map approach where we can track multiple query states simultaneously. This is a tad more advanced but required if we want to render multiple trees at the same time. 


In addition, a bonus would to add a hover prefetching for zero perceived latency on expansion.

Let me know if you want to discuss the proposed changes any further!

PR #1508, opened 2026-01-06
Fix seasonal date presets selecting next winter

Emily-ejag:
The Winter date preset incorrectly calculated years when the current date falls in January-March (the portion of winter after the year changes). This caused the date picker to show dates a full year ahead.

### Bug fixes

- **Winter end date**: When in Jan-Mar, winter end was calculated as `currentYear + 1` instead of `currentYear`. On Jan 1, 2025, Winter showed ending Apr 1, 2026 instead of Apr 1, 2025.
- **Future season offsets**: Spring/Summer/Fall were incorrectly pushed to next year when in Jan-Mar winter, since the wrap-around detection didn't account for being past the year boundary.

### Changes

- Added `inWinterJanMar` check to skip year offset for seasons that haven't occurred yet in the current year
- Fixed winter end date calculation to use conditional offset based on current month
- Extracted magic numbers (`WINTER_SEASON_INDEX`, `MARCH_MONTH_INDEX`) as named constants
- Added tests for crossover scenarios (mid-January and March 31)

```javascript
// Before: always added 1 year when wrapping around season indices
if (i > 0 && seasonIndex < currentSeasonIndex) {
  yearOffset = 1;
}

// After: skip offset if in winter Jan-Mar (seasons haven't happened yet this year)
const inWinterJanMar = currentSeasonIndex === WINTER_SEASON_INDEX && currentMonth <= MARCH_MONTH_INDEX;
if (i > 0 && seasonIndex < currentSeasonIndex && !inWinterJanMar) {
  yearOffset = 1;
}
```


maximilianoertel (commented):
Before I review this, could you or someone else on the team help me confirm the intended behaviour? My understanding is that using presets to create administrations should be disallowed when the start date is in the past?

ksmontville (approved):

PR #1658, opened 2026-03-30
Add config-driven scoring classification service

richford:
## Summary

This PR adds a backend scoring classification service that classifies student scores into support levels (`achievedSkill`, `developingSkill`, `needsExtraSupport`). It uses a **JSON config-driven architecture** — all scoring facts (field name mappings, percentile cutoffs, raw score thresholds) live in per-task JSON config files, and the service is a generic interpreter with no task-specific logic.

This is the foundational piece for the score overview endpoint (#1682). The classification logic must exist server-side before the endpoint can aggregate score distributions.

## Architecture

Each task family gets one JSON config file (8 total) that captures three orthogonal concerns:

1. **Score field resolution** — maps 5 field types (`percentile`, `percentileDisplay`, `standardScore`, `standardScoreDisplay`, `rawScore`) to versioned arrays of database column names, with support for grade-conditional resolution
2. **Classification strategy** — discriminated union with 4 types: `percentile-then-rawscore` (swr, sre, pa), `rawscore-only` (phonics), `assessment-computed` (roam-alpaca), `none` (letter)
3. **Value parsing** — which field types need `<>` prefix stripping

All three dimensions use the same **versioned-array pattern**: ordered entries with `minVersion`, first match where `scoringVersion >= minVersion` wins. This supports arbitrarily many scoring version tiers.

```
scoring/
  configs/           # 8 JSON config files (swr, swr-es, sre, sre-es, pa, letter, phonics, roam-alpaca)
  scoring.config-schema.ts    # Zod schema + inferred types
  scoring.config-registry.ts  # Static imports, validation at load time, slug→config Map
  scoring.service.ts           # Generic config interpreter
  scoring.types.ts             # Domain types
```

## What to review

- **JSON configs** (`configs/*.json`): Verify scoring facts match production behavior
  - Percentile cutoffs: legacy `[50, 25]`, updated `[40, 20]` — correct per task/version?
  - Raw score thresholds: swr, sre, pa boundaries per version — still accurate?
  - Score field names: grade-dependent resolution for PA and SRE — match frontend `SCORE_FIELD_MAPPINGS`?
  - Classification types: phonics → rawscore-only, roam-alpaca → assessment-computed, letter → none
- **Zod schema** (`scoring.config-schema.ts`): Grade-conditional field resolution, versioned array pattern
- **Service** (`scoring.service.ts`): Generic interpreter — no task-specific switch/if logic

## Test coverage (81 tests)

**Config validation (17 tests):** All 8 JSON files validate against Zod schema, no duplicate slugs, all expected slugs registered, classification types correct.

**`getSupportLevel` (33 tests):** Raw-score-only tasks, null data, percentile-based classification (legacy and updated norms for swr/sre/swr-es/sre-es), raw score classification (all tasks), grade edge cases, version edge cases, non-classifiable tasks (letter variants).

**`getRawScoreThreshold` (9 tests):** All task/version combinations.

**`resolveScoreFieldNames` (9 tests):** All task slugs, grade-dependent resolution, null grade handling.

**`resolveScoreFieldName` (13 tests):** Version-specific field resolution for all field types across swr (simple), pa (grade-conditional), sre (versioned + grade-conditional), letter (null fields).

## Test plan

- [x] `npm run check-types` passes (no new errors)
- [x] `npm run lint -w apps/backend` passes
- [x] All 81 scoring unit tests pass
- [ ] Engineers confirm scoring thresholds and field mappings match current production behavior

Resolves https://github.com/yeatmanlab/roar-project-management/issues/1682 (partial — scoring classification service only)


amyyeung17 (changes requested):

richford:
@claude please review this PR

richford:
@claude, please re-review this PR.

I fixed most of the findings and will try to provide justification for dismissing the remaining ones below:

## Dismissed Findings

1. **`rawscore-only` naming**: The name describes what data the task has (raw scores only, no percentile), not what the classification does. The discriminated union type and `getSupportLevel` returning `null` make the behavior explicit. Renaming to `display-only` or `unclassified` would lose the meaning.

2. **`classifyByThresholds` boundary at `developingThreshold`**: The exclusive `>` (score exactly at developing = `needsExtraSupport`) matches the frontend's `reports.js` logic.

3. **`GradeConditionalFieldSchema` allows duplicate `gradeLt` conditions**: First-match-wins evaluation is correct and documented. Duplicate entries are harmless (second is unreachable). We author these configs. Adding uniqueness validation is over-engineering for a controlled format.

4. **`roam-alpaca.json` all score fields null**: Correct by design. Roam-alpaca scores come from the assessment itself, not norming tables. `getSupportLevelFieldName('roam-alpaca')` returns `"supportLevel"` for callers to use.

5. **`scoring/index.ts` 0% coverage**: This is a barrel artifact and consistent with the rest of the codebase. If we want to change the coverage config, that should go in a separate PR.

## Data Verification Tasks

1. **`sre-es` minVersion:0 percentile field name**: The config uses `"percentile"` for all versions, but you noted that unversioned `sre-es` displays `correct - incorrect` difference rather than a percentile. Need to check production `run_scores` data to verify whether a `percentile` field actually exists for legacy (pre-v1) `sre-es` runs. If it doesn't, the `minVersion: 0` entry should use `null` instead.

richford:
@claude please re-review

apps/backend/src/services/scoring/configs/sre-es.json
```diff
@@ -0,0 +1,18 @@
+{
+  "taskSlugs": ["sre-es"],
+  "scoreFields": {
+    "percentile": [{ "minVersion": 0, "fieldName": "percentile" }],
+    "percentileDisplay": [{ "minVersion": 0, "fieldName": "percentile" }],
+    "standardScore": [{ "minVersion": 0, "fieldName": "standardScore" }],
+    "standardScoreDisplay": [{ "minVersion": 0, "fieldName": "standardScore" }],
+    "rawScore": [{ "minVersion": 0, "fieldName": "sreScore" }]
```
amyyeung17, line 8: Should `minVersion` be 1 here since `scoringVersion = 0` is unnormed (and shows completely different fields)?

apps/backend/src/services/scoring/configs/swr-es.json
```diff
@@ -0,0 +1,24 @@
+{
+  "taskSlugs": ["swr-es"],
+  "scoreFields": {
+    "percentile": [
+      { "minVersion": 1, "fieldName": "percentile" },
+      { "minVersion": 0, "fieldName": "wjPercentile" }
```
amyyeung17, line 6: Comment in `sre-es.json` applies here too.

apps/backend/src/services/scoring/configs/pa.json
```diff
@@ -0,0 +1,62 @@
...
+  "taskSlugs": ["pa"],
+  "scoreFields": {
+    "percentile": [
+      {
+        "minVersion": 0,
+        "fieldName": {
+          "gradeConditional": true,
+          "conditions": [
+            { "gradeLt": 6, "value": "percentile" },
+            { "default": true, "value": "sprPercentile" }
```
amyyeung17, line 11: I assume `default` always refers to > grade 6 (for now), but it might not be as clear to someone isn't as familiar with the scoring logic, especially if the conditions become more complicated. And would `default` ever be false? I wonder if there’s a clearer way to define this so it's more explicit.

apps/backend/src/services/scoring/scoring.service.ts
```diff
@@ -0,0 +1,310 @@
...
+/**
+ * Resolve a single score field name for a specific task, field type, grade, and scoring version.
+ *
+ * @param taskSlug - The task slug
+ * @param gradeLevel - Numeric grade level, or null
+ * @param fieldType - One of the 5 score field types
+ * @param scoringVersion - The scoring version
+ * @returns The resolved field name, or null if not applicable
+ */
+export function resolveScoreFieldName(
```
amyyeung17, line 173: Just curious, is there a difference between this and `collectVersionSpecificFieldNames` besides their return values? (exact value vs singleton array of value). Perhaps we can consolidate the functions and return `[resolveScoreFieldName]` for lines 226-227? 

In addition, should there be backward compatibility here as well?

apps/backend/src/services/scoring/scoring.service.ts
```diff
@@ -0,0 +1,310 @@
...
+  if (!config) {
+    return { percentileFieldNames: [], rawScoreFieldNames: [] };
+  }
+
+  // When a scoring version is provided, resolve for that specific version only.
+  if (scoringVersion !== undefined) {
+    const version = scoringVersion ?? 0;
+    return {
+      percentileFieldNames: collectVersionSpecificFieldNames(config, 'percentile', gradeLevel, version),
+      rawScoreFieldNames: collectVersionSpecificFieldNames(config, 'rawScore', gradeLevel, version),
```
amyyeung17, line 227: Is there a reason on not including the other field names? (or I guess, what is the intended use of this function?) Both the score report and individual score report require most of the fields listed in the jsons.

apps/backend/src/services/scoring/configs/letter.json
```diff
@@ -0,0 +1,13 @@
...
+  "taskSlugs": ["letter", "letter-es", "letter-en-ca"],
+  "scoreFields": {
+    "percentile": [{ "minVersion": 0, "fieldName": "totalPercentCorrect" }],
+    "percentileDisplay": [{ "minVersion": 0, "fieldName": "totalPercentCorrect" }],
+    "standardScore": [{ "minVersion": 0, "fieldName": null }],
+    "standardScoreDisplay": [{ "minVersion": 0, "fieldName": null }],
+    "rawScore": [{ "minVersion": 0, "fieldName": "totalCorrect" }]
+  },
+  "classification": {
+    "type": "none"
```
amyyeung17, line 11: What does this classification type mean? Phonics also shows totalPercentageCorrect but has type: 'raw score'. Is it because letter is missing `standardScore` & `standardScoreDisplay`?

richford:
@claude, please re-review. Confirm that I have address @amyyeung17's comments without introducing regressions or coverage gaps.

richford:
@claude, please re-review.

apps/backend/src/services/scoring/scoring.service.ts
```diff
@@ -0,0 +1,324 @@
...
+  const gradeLevel = getGradeAsNumber(grade);
+
+  // Try percentile-based classification (config-driven grade threshold).
+  // percentileMaxGrade defaults to 6; null means use percentile for all grades.
+  const maxGrade = classification.percentileMaxGrade;
+  const usePercentile = percentile !== null && (maxGrade === null || (gradeLevel !== null && gradeLevel < maxGrade));
+
+  if (usePercentile) {
+    const cutoffEntry = resolveVersionedEntry(classification.percentileCutoffs, version);
+    if (cutoffEntry) {
```
amyyeung17, line 126: Nitpick: do we want to return early line 13? (alternatively, wrap 137 with `if (thresholdEntry)`)

apps/backend/src/services/scoring/scoring.service.ts
```diff
@@ -0,0 +1,324 @@
...
+    return null;
+  }
+
+  const classification = config.classification;
+  const version = scoringVersion ?? 0;
+  const gradeLevel = getGradeAsNumber(grade);
+
+  // Try percentile-based classification (config-driven grade threshold).
+  // percentileMaxGrade defaults to 6; null means use percentile for all grades.
+  const maxGrade = classification.percentileMaxGrade;
```
amyyeung17: Nitpick: `percentileMaxGrade` suggests that 6 is included. Maybe worth renaming for clarity?

apps/backend/src/services/scoring/scoring.service.ts
```diff
@@ -0,0 +1,324 @@
...
+}
+
+/**
+ * Get raw score thresholds for a task and scoring version.
+ *
+ * @param taskSlug - The task slug (e.g., 'swr', 'sre')
+ * @param scoringVersion - The scoring version, or null for legacy
+ * @returns Raw score thresholds { above, some }, or null if unavailable
+ */
+export function getRawScoreThreshold(taskSlug: string, scoringVersion: number | null): RawScoreThreshold | null {
```
amyyeung17, line 153: I'm a little confused what the intended use for `getRawScoreThresholds` since lines 132-136 in `getSupportLevel` does something similar I believe. It first verifies that the type is 
`percentile-than-rawscore` by eliminating other types, checks if it uses percentile, and finally grabs the thresholds.

apps/backend/src/services/scoring/scoring.service.ts
richford: Good catch. I renamed to `percentileBelowGrade` to make the exclusive boundary explicit. Updated the JSDoc too: "Exclusive upper bound: percentile cutoffs apply for grades strictly below this value."

Fixed in [828335a](https://github.com/yeatmanlab/roar-dashboard/pull/1658/commits/828335a21ef7c0371f15fb63577b0b6abcf3ffe9)

apps/backend/src/services/scoring/scoring.service.ts
richford: `getSupportLevel` uses the thresholds internally to produce a classification label (`above`/`some`/`below`). `getRawScoreThreshold` is a separate accessor for callers that need the threshold values themselves — e.g., to display `"above 25"` in a score report UI.

I added some clarifying text to the JSDoc in [828335a](https://github.com/yeatmanlab/roar-dashboard/pull/1658/commits/828335a21ef7c0371f15fb63577b0b6abcf3ffe9).

apps/backend/src/services/scoring/scoring.service.ts
richford: Yes, you're right. I initially misunderstood your comment but yes. In [6fd328f](https://github.com/yeatmanlab/roar-dashboard/pull/1658/commits/6fd328fc1b5eb28e93037c46d88b4f5c49838358), I changed it to a positive assertion in the `if` clause.

amyyeung17 (approved):

ksmontville (approved):

PR #1676, opened 2026-04-09
Add GET /v1/schools/:schoolId/classes endpoint

richford:
## Summary

This PR adds a paginated sub-resource endpoint to list active classes within a school. It follows the 5-layer architecture (Contract → Route → Controller → Service → Repository) and the standard sub-resource authorization pattern.

### Contract

Extends `SchoolsContract` with `listClasses` at `/:schoolId/classes`. Adds `SchoolClassSchema` with class metadata (name, type, grades, schedule info), `SchoolClassesListQuerySchema` composing pagination, sorting (`name`, `createdAt`), and filtering (`grade`, `classType` via `createFilterQuerySchema`). `strictStatusCodes: true`.

### Authorization

Follows the sub-resource pattern:
1. Verify school exists (404 if not)
2. Super admins bypass all checks
3. Verify user has access to the school via `OrgAccessControls`
4. Require supervisory role. Supervised users (students, guardians) get 403

### Active-only filter

Only returns classes where `rosteringEnded IS NULL`. This is enforced at the repository level.

### New/modified files

- `packages/api-contract/src/v1/schools/schema.ts`: `SchoolClassSchema`, query/response schemas
- `packages/api-contract/src/v1/schools/contract.ts`: `listClasses` endpoint
- `apps/backend/src/routes/schools.ts`: route handler
- `apps/backend/src/controllers/schools.controller.ts`: `listClasses`, `transformSchoolClass`
- `apps/backend/src/services/school/school.service.ts`: `listSchoolClasses`, `authorizeSchoolSubResourceAccess`
- `apps/backend/src/repositories/class.repository.ts`: `listBySchoolId`
- `apps/backend/src/repositories/school.repository.ts`: `getUserRolesForSchool`
- `apps/backend/src/repositories/access-controls/org.access-controls.ts`: `getUserRolesForOrg`

## Review focus

- **Authorization pattern:** Verify `authorizeSchoolSubResourceAccess` correctly enforces supervisory role
- **Grade filter SQL:** Uses `@> ARRAY[value]::app.grade[]` for PostgreSQL array contains. Verify this matches the enum type
- **`getUserRolesForOrg`:** New method on `OrgAccessControls` checks both direct org membership and class membership within the org

Resolves [Issue 1499](https://github.com/yeatmanlab/roar-project-management/issues/1499)

## Types of changes

What types of changes does this pull request introduce?

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
[...]

PR #1677, opened 2026-04-09
Add GET /v1/districts/:districtId/schools endpoint

richford:
## Summary

This PR adds a paginated sub-resource endpoint to list active schools within a district. Reuses `SchoolDetailBaseSchema` for the response shape and supports `?embed=counts`. Follows the 5-layer architecture and the standard sub-resource authorization pattern.

### Contract

Extends `DistrictsContract` with `listSchools` at `/:districtId/schools`. Adds `DistrictSchoolsListQuerySchema` composing pagination, sorting (`name`, `abbreviation`), and optional `counts` embed. `strictStatusCodes: true`.

### Authorization

Same sub-resource pattern as #1676:
1. Verify district exists (404 if not)
2. Super admins bypass all checks
3. Verify user has access to the district via `OrgAccessControls`
4. Require supervisory role. Supervised users get 403

### Embed support

`?embed=counts` returns `{ users, classes }` per school. Uses bulk fetch + Map pattern to avoid N+1 queries.

### Active-only filter

Only returns schools where `rosteringEnded IS NULL`.

### New/modified files

- `packages/api-contract/src/v1/districts/schema.ts`: query/response schemas
- `packages/api-contract/src/v1/districts/contract.ts`: `listSchools` endpoint
- `apps/backend/src/routes/districts.ts`: route handler
- `apps/backend/src/controllers/districts.controller.ts`: `listSchools`, transform helpers
- `apps/backend/src/services/district/district.service.ts`: `listDistrictSchools`, `verifyDistrictAccess`
- `apps/backend/src/repositories/school.repository.ts`: `listAllByDistrictId`, `listAuthorizedByDistrictId`
- `apps/backend/src/repositories/district.repository.ts`: `getUserRolesForDistrict`
- `apps/backend/src/repositories/access-controls/org.access-controls.ts`: `getUserRolesForOrg`

## Review focus

- **Authorization pattern:** Verify `verifyDistrictAccess` correctly enforces supervisory role
- **Embed counts query:** Verify bulk count resolution avoids N+1
- **`getUserRolesForOrg`:** Same new method as #1676 — both PRs add it independently since they target the same base

Resolves [Issue 1723](https://github.com/yeatmanlab/roar-project-management/issues/1723)

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist

### DIFF
--- a/apps/backend/src/services/district/district.service.ts
+++ b/apps/backend/src/services/district/district.service.ts
@@ -200,10 +200,7 @@ export function DistrictService({
         const userRoles = await districtRepository.getUserRolesForDistrict(userId, districtId);
 
         if (!hasSupervisoryRole(userRoles)) {
-          logger.warn(
-            { userId, districtId, userRoles },
-            'Supervised user attempted to list district schools',
-          );
+          logger.warn({ userId, districtId, userRoles }, 'Supervised user attempted to list district schools');
           throw new ApiError(ApiErrorMessage.FORBIDDEN, {
             statusCode: StatusCodes.FORBIDDEN,
             code: ApiErrorCode.AUTH_FORBIDDEN,
@@ -230,18 +227,11 @@ export function DistrictService({
 
       // For regular users, apply access control filtering
       const allowedRoles = rolesForPermission(Permissions.Organizations.LIST);
-      return await schoolRepository.listAuthorizedByDistrictId(
-        { userId, allowedRoles },
-        districtId,
-        queryParams,
-      );
+      return await schoolRepository.listAuthorizedByDistrictId({ userId, allowedRoles }, districtId, queryParams);
     } catch (error) {
       if (error instanceof ApiError) throw error;
 
-      logger.error(
-        { err: error, context: { userId, districtId, options } },
-        'Failed to list district schools',
-      );
+      logger.error({ err: error, context: { userId, districtId, options } }, 'Failed to list district schools');
 
       throw new ApiError('Failed to retrieve district schools', {
         statusCode: StatusCodes.INTERNAL_SERVER_ERROR,
--- a/apps/backend/src/services/district/district.service.ts
+++ b/apps/backend/src/services/district/district.service.ts
@@ -12,7 +12,7 @@ import { ApiErrorMessage } from '../../enums/api-error-message.enum';
 import { logger } from '../../logger';
 import type { PaginatedResult } from '../../repositories/base.repository';
 import type { AuthContext } from '../../types/auth-context';
-
+import type { EnrolledUsersQuery, EnrolledOrgUserEntity, ListEnrolledUsersOptions } from '../../types/user';
 /**
  * Options for listing districts
  */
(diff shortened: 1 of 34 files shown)


richford:
@claude please review this PR

richford:
@claude, please re-review this PR

richford:
@claude please re-review

richford:
@claude please re-review

Emily-ejag (dismissed):

apps/backend/src/services/district/district.service.ts
```diff
@@ -151,9 +168,127 @@ export function DistrictService({
...
+
+      // For regular users, apply access control filtering
+      const allowedRoles = rolesForPermission(Permissions.Organizations.LIST);
+      return await schoolRepository.listAuthorizedByDistrictId({ userId, allowedRoles }, districtId, queryParams);
+    } catch (error) {
+      if (error instanceof ApiError) throw error;
+
+      logger.error({ err: error, context: { userId, districtId, options } }, 'Failed to list district schools');
+
+      throw new ApiError('Failed to retrieve district schools', {
```
ksmontville:
```suggestion
      throw new ApiError(ApiErrorMessage.INTERNAL_SERVER_ERROR, {
```

ksmontville (changes requested):

apps/backend/src/services/district/district.service.ts
richford: I'd like to push back on this. "Failed to retrieve district schools" is a safe, specific message. It doesn't leak any internals and gives the consumer much better information about what failed than "An internal server error occurred." I think this is permissible given our rule about API error messages.

ksmontville: While it doesn't leak anything, I don't think that it really adds any actionable information, either. The logger.error() directly above already outputs the same information with the same context (plus `options`) for developers.

This is the final 'catch all' error path and could be caused by any number of things... server outage, memory leak, network issues, etc. The reason could be entirely unrelated to the action of listing schools in a district itself. These errors should communicate that something is wrong on our end and not that the user did anything incorrectly when interacting with the endpoint.

I don't see any reason to not keep these 500 responses as generic as possible. Makes testing easier (no need to match the string exactly in the test case), and makes changing the generic response easier down the line.

Emily-ejag (dismissed):

Emily-ejag (dismissed):

apps/backend/src/services/district/district.service.ts
```diff
@@ -184,6 +243,80 @@ export function DistrictService({
...
```
ksmontville:
```suggestion
      throw new ApiError(ApiErrorMessage.INTERNAL_SERVER_ERROR, {
```
True that it doesn't leak any information, but it also doesn't provide any actionable information. Having a string error message makes testing harder (need to match the string in the test case), and a 500 error represents an unknwo error (network error, OoM error, server down...) and I think that the error message ought to reflect as such; the cause of the error isn't necessarily related to the actof listing schools in a district (we also have the logger.error() above for additional information on the developer side).

ksmontville (changes requested):

ksmontville (approved):

Emily-ejag (approved):

PR #1697, opened 2026-04-16
Add `POST /administrations` endpoint 

kellyel:
## Proposed changes
This PR adds the `POST /administrations` endpoint. Specifications for this endpoint can be found here: https://github.com/yeatmanlab/roar-project-management/issues/1534

Additionally, this PR adds a method to the authorization service. This method mimics the behavior of createTuples but will throw errors. This is to facilitate allowing us to deleted a created administration in the event that the FGA tuples fail to be created properly.

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist


kellyel:

kellyel:
The minor comment from Claude's latest review, Naming discrepancy between request schema and DB, will be addressed separately as this discrepancy needs to be discussed with other individuals. With that in mind, @claude please review this PR again

kellyel:
@claude I have resolved those comments, please re-review

kellyel:
@claude please re-review this PR

kellyel:
@claude please review the changes to this PR since last review.

apps/backend/src/services/administration/administration.service.ts
```diff
@@ -985,6 +1005,226 @@ export function AdministrationService({
...
+
+      logger.info({ userId, administrationId: created.id }, 'Administration created successfully');
+
+      return created;
+    } catch (error) {
+      if (error instanceof ApiError) throw error;
+
+      logger.error({ err: error, context: { userId } }, 'Failed to create administration');
+
+      throw new ApiError('Failed to create administration', {
```
ksmontville, line 1219:
```suggestion
      throw new ApiError(ApiErrorMessage.INTERNAL_SERVER_ERROR, {
```
I think that we should just keep these as generic as possible, and we have the `logger.error` statement above for developer debugging.

apps/backend/src/services/administration/administration.service.test.ts
```diff
@@ -3552,4 +3558,407 @@ describe('AdministrationService', () => {
...
+        classRepository: mockClassRepo,
+        groupRepository: mockGroupRepo,
+        taskVariantRepository: mockTaskVariantRepo,
+        agreementRepository: mockAgreementRepo,
+      });
+
+      // Act & Assert
+      await expect(service.create(superAdminAuthContext, validRequest)).rejects.toThrow(
+        expect.objectContaining({
+          message: 'Failed to create administration',
```
ksmontville, line 3957:
```suggestion
          message: ApiErrorMessage.INTERNAL_SERVER_ERROR,
```

packages/api-contract/src/v1/administrations/schema.ts
```diff
@@ -515,3 +515,62 @@ export type AdministrationAgreementsListQuery = z.infer<typeof AdministrationAgr
...
+  description: z.string().optional(),
+  dateStart: z.string().datetime(),
+  dateEnd: z.string().datetime(),
+  isOrdered: z.boolean().optional().default(false),
+  orgs: z.array(z.string().uuid()).optional().default([]),
+  classes: z.array(z.string().uuid()).optional().default([]),
+  groups: z.array(z.string().uuid()).optional().default([]),
+  taskVariants: z.array(CreateAdministrationTaskVariantSchema).min(1),
+  agreements: z.array(z.string().uuid()).optional().default([]),
+});
```
ksmontville, line 568:
```suggestion
})
.strict();
```

packages/api-contract/src/v1/administrations/schema.ts
```diff
@@ -515,3 +515,62 @@ export type AdministrationAgreementsListQuery = z.infer<typeof AdministrationAgr
...
+
+/**
+ * Task variant input for creating an administration.
+ * Each task variant includes ordering and optional eligibility/requirement conditions.
+ */
+export const CreateAdministrationTaskVariantSchema = z.object({
+  taskVariantId: z.string().uuid(),
+  orderIndex: z.number().int().min(0),
+  conditionsEligibility: ConditionSchema.nullable().optional(),
+  conditionsRequirement: ConditionSchema.nullable().optional(),
```
ksmontville, line 531:
```suggestion
  conditionsEligibility: ConditionSchema.optional(),
  conditionsRequirement: ConditionSchema.optional(),
```
Does the current frontend logic directly set these to `null` at creation time? If they don't, then I think that we can do without `nullable()` here since the downstream service logic will re-cast them from `undefined` to `null`.

If the frontend does explcitly set these fields to `null` rather than simply omitting them (`undefined`), or if that is the desired behavior, then ignore this comment.

packages/api-contract/src/v1/administrations/schema.ts
```diff
@@ -515,3 +515,62 @@ export type AdministrationAgreementsListQuery = z.infer<typeof AdministrationAgr
...
+});
```
ksmontville, line 532:
```suggestion
})
.strict();
```
We want `.strict()` here as well, since the top level `.strict()` in `CreateAdministrationRequestSchema` only checks top-level keys.

ksmontville (changes requested):

apps/backend/src/services/administration/administration.service.ts
```diff
@@ -985,6 +1005,226 @@ export function AdministrationService({
...
+          orderIndex: tv.orderIndex,
+          conditionsAssignment: tv.conditionsEligibility ?? null,
+          conditionsRequirements: tv.conditionsRequirement ?? null,
+        })),
+        agreementIds,
+      };
+
+      // Create the administration with all related entities in a single transaction
+      const created = await administrationRepository.createWithAssignments(createInput);
+
```
Emily-ejag, line 1210: Should we create FGA tuples here? since when the super admin later tries to list administrations, the newly created administration won't be in the list because no FGA tuples were created for it.
Maybe something like:

```suggestion

// Create FGA tuples so the administration is accessible
// The creator (super admin) should have read access
await authorizationService.createTuple(
  userId,
  FgaRelation.CAN_READ,
  `${FgaType.ADMINISTRATION}:${created.id}`,
);
 
```

apps/backend/src/services/administration/administration.service.ts
```diff
@@ -985,6 +1005,226 @@ export function AdministrationService({
...
+      if (orgIds.length > 0) {
+        const { items: existingDistricts } = await districtRepository.listByIds(orgIds, {
+          page: 1,
+          perPage: orgIds.length,
+        });
+        const { items: existingSchools } = await schoolRepository.listByIds(orgIds, {
+          page: 1,
+          perPage: orgIds.length,
+        });
+        const existingOrgs = [...existingDistricts, ...existingSchools];
```
Emily-ejag, line 1104: Question here:

You are passing the same array (`orgIds`) to both district and school repositories. So I believe:

- Districts will return only districts from that array
- Schools will return only schools from that array
- Together they'll cover all the IDs

But what if:

- A school ID is passed to `districtRepository.listByIds()` and it returns empty (correct)
- Then the same ID is passed to `schoolRepository.listByIds()` and it's found (correct)
- BUT what if the repositories don't filter by type? Then you might get false positives

I think the API contract doesn't distinguish between district and school IDs:
`orgs: z.array(z.string().uuid()).optional().default([]),`

So someone could pass:

1. orgs: ['district-1', 'school-1'] ✅ Works
2. orgs: ['school-1', 'school-1'] ❓ What happens?
3. orgs: ['invalid-id'] ❓ What error message?


I don't know what would be the best approach here? @kellyel @ksmontville

apps/backend/src/services/administration/administration.service.ts
```diff
@@ -985,6 +1005,226 @@ export function AdministrationService({
...
+      const existingTaskVariantIdSet = new Set(existingTaskVariants.map((tv) => tv.id));
+      const missingTaskVariants = taskVariantIds.filter((id) => !existingTaskVariantIdSet.has(id));
+      if (missingTaskVariants.length > 0) {
+        throw new ApiError(ApiErrorMessage.NOT_FOUND, {
+          statusCode: StatusCodes.NOT_FOUND,
+          code: ApiErrorCode.RESOURCE_NOT_FOUND,
+          context: { userId, missingTaskVariants },
+        });
+      }
+
```
Emily-ejag, line 1166: Should we also check the variant status? Since the variant could be `DRAFT` or `DEPRECATED`?

Emily-ejag (changes requested):

apps/backend/src/services/administration/administration.service.ts
ksmontville: Super admins always bypass FGA checks in the service layer, so no need to write any tuples for them.

Although, now I am actually wondering if we need to write an FGA tuple that associates this administration with the orgs/classes/groups/families that were just input?

apps/backend/src/services/administration/administration.service.ts
ksmontville: This is a good thing to consider, and fortunately each of the `listByIds` methods contains a `WHERE` clause which filters by `orgType`, so I think we are covered here.

```ts
    const where = includeEnded
      ? eq(orgs.orgType, OrgType.SCHOOL)
      : and(eq(orgs.orgType, OrgType.SCHOOL), isNull(orgs.rosteringEnded));
```

apps/backend/src/services/administration/administration.service.ts
ksmontville: I believe that all `task-variant` entities are created with `status=published` by default right now (since currently in Firestore we have neither a `task-variant` nor a `status` for it). 

But we might need to do this in the future, so it might not be a bad idea to put that check in now.

apps/backend/src/services/administration/administration.service.ts
Emily-ejag: So the super admin will always see their own administration, got it! 

Current behavior:

- Admin creates administration with orgs: ['district-1', 'school-1']
- Database records are created ✅
- No FGA tuples are created ❌

What happens later:

- A teacher with access to district-1 calls `listOrgs(administrationId)`
- The service checks: "Can you list users in this administration?" → NO FGA tuple exists
- Result: 403

Looking at your authorization model in listOrgs (lines 231-235):

apps/backend/src/services/administration/administration.service.ts
Emily-ejag: Thanks Kyle! That's great to know.

apps/backend/src/services/administration/administration.service.ts
```diff
@@ -734,6 +761,278 @@ export function AdministrationService({
...
+        } catch (deleteError) {
+          // Compensation failed - log for manual intervention
+          logger.error(
+            { err: deleteError, administrationId: created.id },
+            'Compensation failed: administration exists without FGA tuples. Manual intervention required.',
+          );
+        }
+
+        // Re-throw the original FGA error
+        throw fgaError;
```
ksmontville:
```suggestion
        throw new ApiError(ApiErrorMessage.EXTERNAL_SERVICE_UNAVAILABLE, {
          statusCode: StatusCodes.INTERNAL_SERVER_ERROR,
          code: ApiErrorCode.EXTERNAL_SERVICE_FAILED,
          context: { userId, administrationId: created.id },
          cause: fgaError,
        });
```
This looks good, but I think that it could be more valuable with a more detailed error message (particularly when we start developing on the front end to distinguish between roar_core database errors and FGA datbase errors). This way we know a little bit more about the problem, without revealing too much information about our access controls to the client.

And in `ApiErrorMessage.enum`, add this as a new member:

```ts
  /** Generic external service error - use when a downstream dependency fails */
  EXTERNAL_SERVICE_UNAVAILABLE = 'An external service is currently unavailable',
```

And then we can use this in a few more places in `authorization.service` for consistency.

apps/backend/src/services/authorization/authorization.service.ts
```diff
@@ -45,6 +45,32 @@ export function AuthorizationService({
...
+   * @throws {ApiError} INTERNAL_SERVER_ERROR if the FGA write fails
+   */
+  async function writeTuplesOrThrow(tuples: TupleKey[]): Promise<void> {
+    if (tuples.length === 0) return;
+    try {
+      await client.writeTuples(tuples);
+      logger.debug({ tupleCount: tuples.length }, 'FGA tuples written successfully');
+    } catch (error) {
+      logger.error({ err: error, tupleCount: tuples.length }, 'Failed to write FGA tuples');
+      throw new ApiError(ApiErrorMessage.INTERNAL_SERVER_ERROR, {
```
ksmontville:
```suggestion
      throw new ApiError(ApiErrorMessage.EXTERNAL_SERVICE_UNAVAILABLE, {
```

apps/backend/src/services/authorization/authorization.service.ts
```diff
@@ -45,6 +45,32 @@ export function AuthorizationService({
...
 
+  /**
+   * Write tuples to the FGA store, throwing on failure.
+   *
+   * Unlike `writeTuples`, this method propagates errors to callers. Use this
+   * when tuple creation must succeed for the operation to be valid (e.g., in
+   * a Saga pattern where DB writes should be rolled back if FGA fails).
+   *
+   * @param tuples - Array of TupleKey objects to write
+   * @throws {ApiError} INTERNAL_SERVER_ERROR if the FGA write fails
```
ksmontville:
```suggestion
   * @throws {ApiError} EXTERNAL_SERVICE_UNAVAILABLE if the FGA write fails
```

apps/backend/src/services/authorization/authorization.service.ts
ksmontville:
```suggestion
      throw new ApiError(ApiErrorMessage.EXTERNAL_SERVICE_UNAVAILABLE, {
```

ksmontville:
```suggestion
      throw new ApiError(ApiErrorMessage.EXTERNAL_SERVICE_UNAVAILABLE, {
```

ksmontville:
```suggestion
      throw new ApiError(ApiErrorMessage.EXTERNAL_SERVICE_UNAVAILABLE, {
```

ksmontville (changes requested):

Emily-ejag:
LGTM 🚀 I will approve after @ksmontville suggestions

Emily-ejag (dismissed):

ksmontville (dismissed):

Emily-ejag (approved):

ksmontville (approved):

PR #1700, opened 2026-04-16
Assessment SDK Integration Tests

Emily-ejag:
## Proposed changes
 
Implements a complete integration test infrastructure for the Assessment SDK that validates real HTTP behavior against a locally spawned backend with seeded test data.
 
### What Was Implemented
 
**Backend Infrastructure:**
- Automatic baseFixture seeding on startup (truncate + seed via Fishery factories)
- TestAuthProvider: Bypasses Firebase verification in test mode
- Test fixture endpoint: Exposes seeded task variants and test user authId
- Dynamic test route registration: Keeps test code out of production bundle
 
**SDK Test Infrastructure:**
- Vitest config: Separate unit vs integration test projects with dedicated global setup/teardown
- Global setup: Spawns backend process, waits for health check (backend auto-seeds baseFixture on startup)
- Global teardown: Kills backend process cleanly
- Test helpers: SDK initialization, auth context with real user authId from fixture
- Dynamic fixture fetching: Tests fetch task variants and user authId at runtime (no hardcoded UUIDs)

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist


Emily-ejag:
@claude please review

Emily-ejag:
Here are some questions for the team about how we should handle the remaining setup and configuration.
---

## 1. How should we seed the test database?

Tests need a task variant with ID `550e8400-e29b-41d4-a716-446655440000` to exist. Which approach?

- **A)** Create it dynamically in `beforeAll()` via API call
- **B)** Use backend's `TaskVariantFactory` to seed with known ID
- **C)** Use backend's existing `baseFixture` seeding
- **D)** Keep hardcoded UUID and document that backend must be pre-seeded

---

## 2. How should we handle authentication tokens?

Currently one token is shared across all tests. Should we:

- **A)** Generate a unique token per test (most isolated, slightly slower)
- **B)** Generate a unique token per describe block (balanced)
- **C)** Keep one shared token for all tests (fastest, less isolated)

---

## 3. How should we handle the backend build requirement?

Tests need a running backend. Should we:

- **A)** Use `npm run dev` (no pre-build needed, slower startup)
- **B)** Keep `npm run start` and document the build requirement
- **C)** Auto-build in global setup before starting backend (fully automated, slower)

Emily-ejag:

Emily-ejag:

Emily-ejag:
@claude please re-review

Emily-ejag:
@claude please re-review

Emily-ejag:

Emily-ejag:
@claude I believe I have clarify the `INTEGRATION_TESTS.md`

maximilianoertel (changes requested):
Great start getting the SDK integration test infra off the ground! I have some architectural feedback before this goes further, because the current approach diverges from the patterns we've established in the backend and introduces some issues that will be painful to live with.

The core approach is right. The SDK is a published package that talks to the backend over HTTP, so spawning a real backend process and having the SDK make real HTTP requests against it is conceptually the correct architecture for SDK integration tests. This is fundamentally different from the backend's own route integration tests, and that's fine as the SDK is an external client, and its tests should reflect that.

The problems are in the execution, not the architecture.

#### 1. Production code was modified to support testing

This is the biggest issue. Test infrastructure should never leak into production code.
- `server.ts` now has DB seeding logic in the production startup path
- `routes/index.ts` dynamically imports test routes with an acknowledged race condition 
- `auth.service.ts` swaps to `TestAuthProvider` in test mode
- A new `routes/test.ts` exposes `baseFixture` over HTTP

The backend's own integration tests achieve all of this (auth mocking, fixture access, DB seeding) without touching a single production file.

Instead, we should create a dedicated test server entrypoint in the backend (e.g. `apps/backend/src/server-test.ts`) that reuses the existing test-support infrastructure: `initializeDatabasePools`, `truncateAllTables`, `seedBaseFixture`, FGA initialization and finally also starts Express. All the test-specific setup lives in this script, not in `server.ts` or `auth.service.ts`. The production code stays untouched.


#### 2. No FGA authorization, all tests use `isAnonymous: true`
Because the standalone server doesn't initialise OpenFGA, all SDK tests sidestep authorization entirely. This means we're not actually testing one of the most critical paths end-to-end.

The test server entrypoint should initialize FGA the same way the backend's `vitest.integration.globalSetup.ts` does to create a store, deploy the authorization model and sync tuples from Postgres. Then SDK tests can exercise real authorization with non-anonymous runs and proper role-based access. That's crucial here.


#### 3. Fixture data exposed via HTTP endpoint

The `GET /v1/test/fixture` endpoint is a workaround for the SDK not being able to import `baseFixture` directly. Since the SDK tests run out-of-process, they genuinely need a way to discover fixture IDs.

Instead of the current approach, the test server entrypoint can write fixture data (variant IDs, user auth IDs, etc.) to a temp JSON file on startup. The SDK's global setup reads that file instead of making an HTTP call. No new route needed, no race condition, no production code change. Alternatively, the test server could try to log the fixture data to stdout in a parseable format that the global setup captures.


#### 4. `TestAuthProvider` in production auth service

The `TestAuthProvider` approach (token string = Firebase UID) is a reasonable idea for SDK tests, but it shouldn't live in the production `auth.service.ts`.

Instead, the test server entrypoint can mock `AuthService.verifyToken` at startup (the same pattern `authenticateAs` uses in route tests), or register a simple token-as-UID provider. Either way, that should be done in the test entrypoint, not in production code.

<br>

The test server entrypoint is the key piece here. It's a thin script that composes existing backend test-support utilities (`initializeDatabasePools`, `truncateAllTables`, `seedBaseFixture`, FGA init, etc.) into a runnable server. All the building blocks already exist; they just need to be wired up outside of production code.

Emily-ejag:
@claude can you please review and make sure I am following @maximilianoertel's feedfack?

Emily-ejag:
@claude can you please re-review

Emily-ejag:
@claude please re-review

Emily-ejag:
@claude can you check if the minor issues were addressed?

packages/assessment-sdk/package.json
```diff
@@ -30,7 +30,11 @@
     "format": "prettier --write .",
     "format:check": "prettier --check .",
     "typecheck": "tsc --noEmit",
-    "test": "vitest",
+    "test": "vitest run --exclude '**/*.integration.test.ts'",
```
richford: The issue's acceptance criteria include "CI can run these tests." But the `ci.yml` workflow file is unchanged. The only test step is `npm run test`, which delegates to turbo run test, and the SDK's test script excludes the integration tests.

So. the integration tests are not invoked anywhere in CI. I think we need a CI step like:
```yaml
- name: Run SDK integration tests
  env:
    CORE_DATABASE_URL: postgresql://postgres:postgres@localhost:5432/roar_core_test
    ASSESSMENT_DATABASE_URL: postgresql://postgres:postgres@localhost:5432/roar_assessment_test
    FGA_API_URL: http://localhost:8080
  run: npm run test:integration -w packages/assessment-sdk
```

or maybe a turbo task wired up to make the `npm run test` command cover integration tests.

packages/assessment-sdk/src/receiver/roar-api.integration.test.ts
```diff
@@ -0,0 +1,441 @@
...
+        expect(runId).toMatch(/^[0-9a-f-]{36}$/); // UUID format
+      }
+    });
+
+    it('should return 403 for administration outside user hierarchy', async () => {
+      // This test verifies the negative case: user is denied access to an administration
+      // they are not authorized for. schoolAStudent is enrolled in district (and schoolA),
+      // but has no FGA tuples for districtB or its administrations.
+      // Without this test, the FGA can_create_run check could be silently disabled
+      // and the test suite would still pass.
```
richford, line 436: I think this test passes for reasons other than the FGA `can_create_run` check. The flow in apps/backend/src/services/run/run.service.ts is:

1. `verifyAdministrationAccess(authContext, body.administrationId)`: runs an FGA `can_read` check first
2. Only if `can_read` passes does the code reach `requirePermission(CAN_CREATE_RUN)`.

`schoolAStudent` has no role on `districtB`, so the `CAN_READ` check fails and you never reach the `CAN_CREATE_RUN` line. The test's assertion is correct, but it's verifying `can_read`, not `can_create_run`.

So, for example, if someone later came in and silently relaxed `can_create_run` to `student or teacher`, this test would not catch it. To actually test the `can_create_run` check, we need a user who has `can_read` but not `can_create_run`. Like a `schoolATeacher` (supervisory tier, but not a student on the administration).

packages/assessment-sdk/INTEGRATION_TESTS.md
richford, line 1: I'd like to suggest we delete this file.

A few reasons:

- The repo doesn't have per-package documentation as a pattern. there's no INTEGRATION_TESTS.md under apps/backend/ (line 411 here actually links to one that doesn't exist), no per-package README.md, and CLAUDE.md says "Never create documentation files (*.md) or README files unless explicitly requested by the User."

- Most of the content also isn't earning its keep:
  - The "Test Cases" checklist duplicates the test file and will go stale on the first new test
  - "Architecture Decisions" defends choices that aren't controversial. For example, "Why use SDK client instead of raw HTTP requests?" Well, it's an SDK.
  - "Future Improvements" is a wishlist nobody owns
  - "Authentication Token Strategy" is prose that re-explains four lines of code
  - The fixture-file JSON example, the references section, and the overlapping "Architecture" / "Test Server Initialization" / "Architecture Decisions" sections all describe code that's right there in the diff

- The useful stuff is mostly the prerequisites (test DB names, env vars, OpenFGA running) and the troubleshooting hints (port 4001 in use, `dist/server-test.js` missing because of the `BUILD_TEST_SERVER=true` build gate). I'd rather see that as a comment block at the top of `vitest.integration.globalSetup.ts`, where someone debugging a failure will find it. The architectural rationale already lives in the commit history and the PR description.

If we decide later we want broader contributor onboarding docs for the SDK (and we probably should), that's worth a separate conversation and PR.

apps/backend/src/server-test.ts
```diff
@@ -0,0 +1,205 @@
...
+ *   NODE_ENV=test node dist/server-test.js
+ */
+
+import 'dotenv/config';
+import fs from 'fs';
+import http from 'http';
+import type { Express } from 'express';
+import { initializeDatabasePools, closeDatabasePools } from './db/clients';
+import { truncateAllTables, runMigrations } from './test-support/db';
+import { seedBaseFixture } from './test-support/fixtures';
```
richford, line 33: nitpick: The JSDoc for `seedBaseFixture` lists the places where it is used and mentions "called by vitest.setup.ts in `beforeAll`." But now it's used here as well.

I don't think we should maintain a list of call sites in the JSDoc. Instead, the JSDoc just needs to convey what it does and how it's used. Like for the `baseFixture`

```ts
/**
 * The seeded base fixture.
 *
 * Populated by `seedBaseFixture()` and treated as read-only by consumers.
 * Accessing this before `seedBaseFixture()` has resolved will return `undefined`.
 *
 * @example
 * ```typescript
 * import { baseFixture } from '../test-support/fixtures';
 *
 * it('user in school sees administration assigned to district', async () => {
 *   const results = await repo.query({ userId: baseFixture.schoolAStudent.id });
 *   expect(results).toContain(baseFixture.administrationAssignedToDistrict.id);
 * });
 * ```
 */
export let baseFixture: BaseFixture;
```

Same idea for `seedBaseFixture`. Say what it does and what its contract is, not which files happen to call it today.

You didn't create this problem, but extending its use revealed the stale JSDoc comment so let's fix it here.

turbo.json
```diff
@@ -16,7 +16,10 @@
     "CORE_DATABASE_USER",
     "ASSESSMENT_DATABASE_NAME",
     "ASSESSMENT_DATABASE_USER",
-    "LOG_LEVEL"
+    "LOG_LEVEL",
+    "TEST_FIXTURE_FILE",
+    "FGA_API_URL",
+    "BACKEND_PORT"
```
richford: Should `"BUILD_TEST_SERVER"` be included here? Otherwise cached production builds could be served when a test build is needed. Is that correct?

richford (changes requested):

packages/assessment-sdk/package.json
maximilianoertel: All apps/packages should follow the command structure established here: 

https://github.com/yeatmanlab/roar-dashboard/blob/38175db6d0e37d677923038628604076f27c8f98/apps/backend/package.json#L14-L19

`npm run test` should always execute all tests, with sub-commands available for development purposes. Diverging from this will create confusion when jumping from one app/package to another

Emily-ejag: Got it, thank you both. I will change it to `"test": "vitest run",`

packages/assessment-sdk/INTEGRATION_TESTS.md
Emily-ejag: Yeah I think once the code is in place, this file becomes redundant. I wanted to create it to explain How the tests server works, the prerequisites, etc. I think I can put them as code comments.

apps/backend/src/server-test.ts
Emily-ejag: Thanks @richford I have added updated the JSDoc here: https://github.com/yeatmanlab/roar-dashboard/pull/1700/changes#diff-345d826dd440932ea01244f794658d795a3f76b63fa926392bb939c1b894635fR29

Emily-ejag:
@claude can you please review and make I have solve the comments and not introduced new issues?

Emily-ejag:
@claude can you please re-review

Emily-ejag:
@claude can you please re-review

Emily-ejag:
@claude can you please re-review

Emily-ejag:
@claude please re-review

packages/assessment-sdk/vitest.integration.globalSetup.ts
```diff
@@ -0,0 +1,252 @@
...
+ * Builds the backend if not already built.
+ * Required because the global setup runs `node dist/server-test.js` directly.
+ */
+async function buildBackendIfNeeded(backendDir: string): Promise<void> {
+  const testServerBin = path.join(backendDir, 'dist', 'server-test.js');
+
+  if (existsSync(testServerBin)) {
+    console.log('[SDK Integration Tests] Test server binary already built, skipping build');
+    return;
+  }
```
richford, line 68: This part just tests for existence, not whether `dist/server-test.js` is outdated. Can we do this instead?


```suggestion
  const testServerBin = path.join(backendDir, 'dist', 'server-test.js');
   const sourceFile = path.join(backendDir, 'src', 'server-test.ts');
   
   if (existsSync(testServerBin) && existsSync(sourceFile)) {
     const binStats = statSync(testServerBin);
     const srcStats = statSync(sourceFile);
     if (binStats.mtime > srcStats.mtime) {
       console.log('[SDK Integration Tests] Test server binary is up-to-date');
       return;
     }
   }
```

packages/assessment-sdk/src/test-support/sdk-test-helper.ts
```diff
@@ -0,0 +1,120 @@
...
+export function createTestAuthContext() {
+  if (!cachedTestToken) {
+    cachedTestToken = 'test-token-' + Math.random().toString(36).slice(2);
+  }
+
+  return {
+    getToken: async () => cachedTestToken!,
+    refreshToken: async () => cachedTestToken!,
+  };
+}
```
richford, line 58: This function creates `cachedTestToken` including a random number, but later overwrites it with a real `data.testUser.authId` (line 113). The JSDoc says "The token must be a real user's authId from the seeded database." If `getBaseFixtureData()` is called *after* `initTestSdk()`, the token will be overwritten, but the SDK is already initialized with a wrong token. This only works because `beforeAll` is guaranteed to call `getBaseFixtureData()` before any requests, but I think this is a latent bug that will surface if we change test structure.

Can we either (a) don't pre-populate the token, or (b) delay SDK initialization until after `getBaseFixtureData()`.

```suggestion
   export function createTestAuthContext() {
     return {
       getToken: async () => {
         if (!cachedTestToken) {
           throw new Error('Test token not initialized. Call getBaseFixtureData() first.');
         }
         return cachedTestToken;
       },
       refreshToken: async () => cachedTestToken!,
     };
   }
```

packages/assessment-sdk/vitest.integration.globalSetup.ts
```diff
@@ -0,0 +1,252 @@
...
+  // Spawn the test server process using the dedicated server-test.ts entrypoint
+  // This entrypoint initializes databases, seeds fixtures, sets up FGA, and mocks auth
+  backendProcess = spawn('node', ['dist/server-test.js'], {
+    cwd: backendDir,
+    env: {
+      ...process.env,
+      // Use 'production' to avoid pino-pretty transport which crashes in bundled ESM
+      // (thread-stream uses __dirname, unavailable in ES modules).
+      // The test server behavior is controlled by explicit setup steps, not NODE_ENV.
+      NODE_ENV: 'production',
```
richford, line 198:
```suggestion
      // Use 'production' to avoid pino-pretty transport which crashes in bundled ESM
      // (thread-stream uses __dirname, unavailable in ES modules).
      // Trade-off: logs are JSON only, less readable during debugging.
      // For human-readable logs, rebuild locally: NODE_ENV=development npm run test -w packages/assessment-sdk
      NODE_ENV: 'production',
```

packages/assessment-sdk/vitest.integration.globalSetup.ts
```diff
@@ -0,0 +1,252 @@
...
+
+export async function teardown() {
+  // @ts-expect-error globalThis doesn't have __BACKEND_PROCESS__ in type definitions
+  const proc = globalThis.__BACKEND_PROCESS__ as ReturnType<typeof spawn> | undefined;
+  if (proc) {
+    proc.kill();
+    await new Promise<void>((resolve) => proc.on('close', resolve));
+    console.log('[SDK Integration Tests] Backend process terminated');
+  }
+}
```
richford, line 280: Nitpick suggestion to add explicit cleanup:

```suggestion
export async function teardown() {
  // @ts-expect-error globalThis doesn't have __BACKEND_PROCESS__ in type definitions
  const proc = globalThis.__BACKEND_PROCESS__ as ReturnType<typeof spawn> | undefined;
  if (proc && !proc.killed) {
    proc.kill('SIGTERM');
    const timeout = new Promise<void>((resolve) => setTimeout(resolve, 5000));
    const close = new Promise<void>((resolve) => proc.on('close', resolve));
    await Promise.race([timeout, close]);
    console.log('[SDK Integration Tests] Backend process terminated');
  }
}
```

apps/backend/src/server-test.ts
```diff
@@ -0,0 +1,208 @@
...
+ * @param fixtureFile - Path to write the fixture JSON file
+ */
+async function writeFixtureFile(fixtureFile: string): Promise<void> {
+  const { baseFixture } = await import('./test-support/fixtures');
+
+  if (!baseFixture) {
+    throw new Error('[server-test] baseFixture not seeded');
+  }
+
+  const fixtureData = {
```
richford: The fixture shape is duplicated. This object literal is the producer, and the inline return type of `getBaseFixtureData()` in `packages/assessment-sdk/src/test-support/sdk-test-helper.ts` is the consumer. They have to stay in sync by hand. If a field is added here without updating the consumer type, callers don't get the new field. And if a field is renamed there without updating here, JSON parsing silently produces `undefined`.

Can we extract a shared `TestFixture` type into a location both can import (e.g. `apps/backend/src/test-support/fixtures/test-fixture.type.ts`, then re-export from a path the SDK can reach). At minimum, a `TestFixture` type defined once and used as `const fixtureData: TestFixture = { ... }` here, with the consumer importing it via the SDK's existing dependency on the contract package or test-support.

If extracting feels out of scope, a JSDoc block on this object pointing at the consumer (and vice versa) prevents drift from becoming silent. But I'd prefer a shared type.

richford (changes requested):
@Emily-ejag, this is looking really solid. We're very close. I left some inline comments with minor suggestions. Feel free to push back if you think they are off base. 

Additionally, there are four public functions that need JSDoc with `@param` / `@returns`:

- createTestAuthContext (sdk-test-helper.ts:40)
- globalSetup (vitest.integration.globalSetup.ts:143)
- teardown (vitest.integration.globalSetup.ts:244)
- TestAuthProvider.verifyToken (test-auth.provider.ts:13, the class has JSDoc but the method doesn't).

I think we're on the final rounds of iteration, which reflect the ambitious scope of your task, not any failings on your part. We can land this very soon.

Emily-ejag:
@claude can you please review?

Emily-ejag:
@claude can you check that I solve that remaining issue?

kellyel (approved):

richford (approved):

PR #1702, opened 2026-04-17
Add `GET /v1/families/:familyId/users`

amyyeung17:
### Summary
This PR implements the `GET /v1/families/:familyId/users endpoint` to list all users enrolled in a family. 

###  Features

- Retrieves all users under a family (direct membership as a child or parent) 
- Uses FGA pattern established by https://github.com/yeatmanlab/roar-dashboard/pull/1699
- Super admin can list users from any families 
- Similar to other user endpoints, allows filtering by roles and grade

This resolves [Issue 1551](https://github.com/yeatmanlab/roar-project-management/issues/1551)

## Types of changes

What types of changes does this pull request introduce?

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Refactoring (non-breaking change that does not add functionality but makes code cleaner or more efficient)
- [ ] Documentation Update
- [ ] Tests (new or updated tests)
- [ ] Style (changes to code styling)
- [ ] CI (continuous integration changes)
- [ ] Repository Maintenance
- [ ] Other (please describe below)

## Checklist

- [x] I have read the [guidelines for contributing](https://github.com/yeatmanlab/roar-dashboard/blob/main/.github/CONTRIBUTING.md).
- [x] The changes in this PR are as small as they can be. They represent one and only one fix or enhancement.
- [x] Linting checks pass with my changes.
- [x] Any existing unit tests pass with my changes.
- [x] Any existing end-to-end tests pass with my changes.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] If this PR fixes an existing issue, I have added a unit or end-to-end test that will detect if this issue reoccurs.
- [x] I have added JSDoc comments as appropriate.
- [x] I have added the necessary documentation to the [roar-docs repository](https://github.com/yeatmanlab/roar-docs).
- [x] I have shared this PR on the roar-pr-reviews channel (if I have access)
- [x] I have linked relevant issues (if any)

PR #1713, opened 2026-04-24
`SDK` Rename bare runs endpoints to user scoped endpoints

Emily-ejag:
## Proposed changes
Refactor the assessment SDK and backend runs endpoints from bare routes (`/runs`, `/runs/:runId/event`) to user-scoped routes (`/user/:userId/runs`, `/user/:userId/runs/:runId/event`). The `userId` is extracted from the authenticated user's permissions, with FGA enforcing access control.

**Key architectural decision:** User-scoped routes enable explicit authorization checks via FGA, preventing accidental data leaks and making permission boundaries clear in the API contract.

### 1. Fixed Run Ownership
- **Issue**: Runs were being created with `authContext.userId` instead of `targetUserId`
- **Impact**: Super admins creating runs for other users would incorrectly own those runs
- **Fix**: Changed to use `targetUserId` so runs are owned by the intended user
- **Security**: Prevents privilege escalation where super admins could own runs created for others

### 2. Updated Contract Documentation
- **Create endpoint**: 
  - Clarified that super admins have unrestricted access
  - Users may create runs for themselves or on behalf of a child user if they hold `CAN_READ_CHILD` permission
  - Returns 403 if user lacks access to the provided administration context or does not have permission to create runs for the specified user
- **Event endpoint**: 
  - Updated to reflect strict ownership enforcement (no parent/guardian bypass for event posting)
  - Only the run owner can post events
  - Super admins do NOT bypass ownership checks for run events

### 3. Removed Dead Code
- **Issue**: Service-level validation for `isAnonymous && administrationId` was unreachable
- **Reason**: Zod schema validation catches this at the request level (returns 400 BAD_REQUEST)
- **Fix**: Removed dead code and added clarifying comment explaining the validation is handled at request level

### 4. Improved Tests
  - Added integration test verifying run ownership after creation
  - Uses `RunRepository.getById()` to fetch created run from database
  - Asserts `run.userId === tiers.student.id` to confirm the `userId: targetUserId` fix works correctly

### FGA Pattern Used
Follows the established FGA authorization pattern from the codebase:

1. **Service layer** enforces authorization (not controller)
2. **Two-step verification:**
   - Step 1: Verify resource exists (returns 404 if not)
   - Step 2: Verify user has access via FGA (returns 403 if not, skipped for super admins)
3. Uses `ApiErrorMessage` enum for error responses (never raw strings)
4. Returns 403 Forbidden if user lacks access

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist


Emily-ejag:
@claude please review this PR

Emily-ejag:

Emily-ejag:
@claude can you please re-re-review

Emily-ejag:
@claude can you please re-re-re-review?

apps/backend/src/routes/runs.integration.test.ts
```diff
@@ -113,70 +116,111 @@ function buildEngagementEventBody() {
...
+        .set('Authorization', 'Bearer token')
+        .send(buildCreateRunBody());
 
       expect(res.status).toBe(StatusCodes.FORBIDDEN);
       expect(res.body.error.code).toBe(ApiErrorCode.AUTH_FORBIDDEN);
     });
+
+    // TODO: Add test for caregiver with CAN_READ_CHILD permission creating run for child user
+    // This requires setting up FGA tuple: caregiver has CAN_READ_CHILD on student user
+    // Once set up, this test should verify caregiver can create run for their child
```
richford: Can we address this TODO in this PR? If not, please create an issue for it, add it to the BERF project, and reference it in a comment or in this PR description.

richford (changes requested):

apps/backend/src/routes/runs.integration.test.ts
Emily-ejag: Yeah I can try to create a FGA tuple granting the caregiver the permission and then verify that the run is owned by the student

Emily-ejag: After chatting with @kellyel and @ksmontville, we realized this `TODO` is actually incorrect. Parents/caregivers should be able to read their child’s runs, but they should not create runs for them.

So the test failing when trying to create a run as the caregiver is the expected behavior.

I’ll update the TODO/test approach to follow:

1. Authorize as the child and create the run.
2. Authorize as the caregiver/parent.
3. Verify the caregiver can read the child’s run.

Emily-ejag:
@claude please review

Emily-ejag:
@claude re-review please

Emily-ejag:
@claude please re-review

Emily-ejag:
@claude I have clarified the JSDoc

Emily-ejag:
@claude can you check if I have not introduce any new error on my last commit?

Emily-ejag:
@claude can you check if my last fix did not introduce any issues?

apps/backend/src/services/run-event/run-event.service.ts
```diff
@@ -34,34 +34,56 @@ export function RunEventService({
...
+   * Verifies that the authenticated user has access to post events for the target user's run.
+   *
+   * Run events enforce strict ownership — only the run owner can post events to their own runs.
+   * Even super admins cannot post events on behalf of other users.
+   *
+   * @param authContext - User's auth context (id and super admin flag)
+   * @param targetUserId - The user ID who owns the run
+   * @throws {ApiError} FORBIDDEN if user doesn't own the run
+   */
+  async function verifyUserAccess(authContext: AuthContext, targetUserId: string): Promise<void> {
```
richford, line 59: The access policy here diverges from `RunService.verifyUserAccess` (in `run.service.ts`) in a way that breaks the parent-guardian workflow we want.

I think the actual product flow is a parent logs in, selects their child, taps "start assessment," hands the device to the child, and the child taps through the trials. The child does not have to log in on their own. So every HTTP request that goes out during the session, e.g.,

- `POST /v1/user/<child>/runs`
- `POST /v1/user/<child>/runs/<runId>/event`

carries the parent's auth token. The parent's uid is the `requesterUserId` on every event call.

`RunService.create` correctly accepts this: it allows the parent via `CAN_READ_CHILD` + `CAN_CREATE_RUN_FOR_CHILD`. But here, `RunEventService.verifyUserAccess` does not. It has a strict equality check `requesterUserId !== targetUserId` returns 403 on every single trial post, completion, or abort. So, as written, this PR enables parents to create a run for their child and then immediately blocks every subsequent action that completes that run. The same problem applies to a super-admin running the same workflow for support purposes.

Proposed fix: events should use the same access policy as `RunService.create`. Extract a shared helper that both services consume and re-check `CAN_CREATE_RUN_FOR_CHILD` for the cross-user, non-super-admin path so cross-user permission is verified at every event call rather than just at run-creation time. And then update tests accordingly.

richford (changes requested):

apps/backend/src/services/run-event/run-event.service.ts
Emily-ejag: @richford yeah this makes sense, I have created a shared `verifyRunEventAccess` helper allowing parents to post run events for their children using the `CAN_CREATE_RUN_FOR_CHILD` permission, completing the parent-child assessment workflow.

Emily-ejag:
@claude can you check if I resolved @richford's comment and did not introduce new issues?

Emily-ejag:
@claude I believe I have updated the JSDocs

apps/backend/src/routes/runs.integration.test.ts
richford, line 1: Can you add integration tests for posting events as well.

```ts
it('parent with CAN_CREATE_RUN_FOR_CHILD can post events for their child', async () => {
  // Set up parent + child in same family, parent role grants CAN_CREATE_RUN_FOR_CHILD
  // Create run as the child
  // Authenticate as parent, post each of the four event types — assert 200 for all four
});

it('parent without CAN_CREATE_RUN_FOR_CHILD cannot post events for their child', async () => {
  // Create parent and child with no family relationship (or a role lacking the permission)
  // Authenticate as parent, post any event → 403
});

it('parent on family A cannot post events for child in family B', async () => {
  // Cross-family isolation — exact analog of the create-side test
});
```

apps/backend/src/services/run-event/run-event.service.ts
richford: I don't see a `verifyRunEventAccess` helper. I do see that the two `verifyUserAccess` methods are now much more aligned, which is good, although they're not totally identical. Is that intentional. If not, could I ask you to create a shared service layer helper/utility for this?

```ts
export async function verifyTargetUserAccess(
  authContext: AuthContext,
  targetUserId: string,
  permission: FgaRelation,
  deps: { familyRepository: FamilyRepository; authorizationService: AuthorizationService },
): Promise<string[]> { /* ... */ }
```

or something like that?

richford (changes requested):

Emily-ejag:
@claude can you check if I solved @richford's comments without introducing any issues?

Emily-ejag:
@claude what about now?

apps/backend/src/services/run-event/run-event.service.test.ts
```diff
@@ -38,23 +49,131 @@ describe('RunEventService', () => {
...
+    mockAuthorizationService = {
+      hasAnyPermission: vi.fn(),
+    };
+
     runEventsService = RunEventService({
       runRepository: runRepository,
       runTrialsRepository: runTrialsRepository,
       runTrialInteractionsRepository: runTrialInteractionsRepository,
+      familyRepository: mockFamilyRepository as unknown as FamilyRepository,
+      authorizationService: mockAuthorizationService as unknown as ReturnType<typeof AuthorizationService>,
```
richford: These two `as unknown as X` violate our typed mock factory pattern. The sibling test file at `apps/backend/src/services/run/run.service.test.ts` already has the right pattern:

```ts
import { createMockFamilyRepository, type MockFamilyRepository } from '../../test-support/repositories';
import { createMockAuthorizationService, type MockAuthorizationService } from '../../test-support/services';

let familyRepository: MockFamilyRepository;
let authorizationService: MockAuthorizationService;

beforeEach(() => {
  // ...
  familyRepository = createMockFamilyRepository();
  authorizationService = createMockAuthorizationService();

  runEventsService = RunEventService({
    runRepository,
    runTrialsRepository,
    runTrialInteractionsRepository,
    familyRepository,           // no cast
    authorizationService,       // no cast
  });
});
```

apps/backend/src/services/authorization/verify-target-user-access.ts
richford, line 1: Thanks for adding this helper! Can you add a dedicated `verify-target-user-access.test.ts` next to it so the contract is pinned independently of callers. Suggested cases:

- super-admin → [], no repo/FGA calls
- self-access → [], no repo/FGA calls
- cross-user with permission → returns family IDs
- cross-user without permission → throws 403 with requiredPermission in context
- cross-user, target in no families → still throws (pin the no-short-circuit behavior)
- repository / FGA service rejections → propagated

Use `createMockFamilyRepository` and `createMockAuthorizationService` to stay consistent with run.service.test.ts.

richford (changes requested):

apps/backend/src/services/run-event/run-event.service.test.ts
Emily-ejag: yeah you are right about the unknowns . I was creating a partial mock object that only implements two methods, since `FamilyRepository` is the full interface with many more methods Typescript will not allow it. At pointed out on daily, yes there are the Mocks so I can add those :) Thank you!

apps/backend/src/services/authorization/verify-target-user-access.ts
Emily-ejag: Sure, I will create that file for those cases

Emily-ejag:
@claude can you verify that I have resolved @richford 's comments and also did not introduced any new issues?

richford (dismissed):
Yay! Looks great!

packages/authz/authorization-model.fga
```diff
@@ -246,6 +246,7 @@ type family
     define can_delete_child: parent
     define can_read_child: parent
     define can_consent_for_child: parent
+    define can_create_run_for_child: parent
```
amyyeung17, line 249: Perhaps we can add some tests for this in `authz/test/family-permissions.fga.yaml`?

packages/api-contract/src/v1/runs/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from './schema';
 export * from './contract';
+export type { CreateRunRequestBody, RunEventBody, CreateRunResponse } from './schema';
```
amyyeung17: Just curious, do we need to a separate export for these types? `export * from './schema'` should include them automatically.

apps/backend/src/test-support/fga/fga-test-tuples.helper.ts
```diff
@@ -73,3 +75,23 @@ export async function writeFgaGroupMembership(userId: string, groupId: string, r
...
+ * @param familyId - The family ID
+ * @param role - The user's role in the family ('parent' or 'child')
+ * @param joinedOn - When the family membership began, or null for unknown
+ * @param leftOn - When the family membership ended, or null for indefinite
+ */
+export async function writeFgaFamilyMembership(
+  userId: string,
+  familyId: string,
+  role: UserFamilyRole,
+  joinedOn: Date | null,
```
amyyeung17: I believe `joinedOn` cannot be null based on its definition in the schema (`apps/backend/src/db/schema/core/user-families.ts`). It appears to default to `Date.now()` in the database.

apps/backend/src/services/run/run.service.ts
```diff
@@ -32,52 +34,95 @@ export function RunService({
...
+    // so we don't need to check it here. This is caught at the request level (400 BAD_REQUEST).
 
     if (!isAnonymous) {
+      // When a parent creates a run for their child, the child is the one who needs
+      // access to the administration and CAN_CREATE_RUN permission — not the parent.
+      // The parent's authorization is verified separately via CAN_CREATE_RUN_FOR_CHILD.
+      // Super admins always use their own context (they bypass FGA checks anyway).
+      const isParentForChildRun = requesterUserId !== targetUserId && !isSuperAdmin;
+      const administrationAuthContext: AuthContext = isParentForChildRun
+        ? { userId: targetUserId, isSuperAdmin: false }
```
amyyeung17: This is an almost impossible case, but should we consider if the child can be potentially a super admin? I previously implemented something similar in `getUserAdministration`, and Windsurf brought up this concern.

apps/backend/src/services/run/run.service.ts
```diff
@@ -32,52 +34,95 @@ export function RunService({
...
     const { isAnonymous } = body;
 
-    if (isAnonymous && body.administrationId) {
-      throw new ApiError('administrationId must not be provided for anonymous runs', {
-        statusCode: StatusCodes.UNPROCESSABLE_ENTITY,
-        code: ApiErrorCode.REQUEST_VALIDATION_FAILED,
-        context: { userId },
-      });
-    }
+    // Verify user has access to the target user and get family IDs for potential reuse
```
amyyeung17, line 101: Would it make sense to verify the target user exists first before continuing?

apps/backend/src/services/run/run.service.ts
```diff
@@ -95,13 +140,35 @@ export function RunService({
...
+        // For self-runs, this is the requester. For parent-for-child runs, this is the child.
         await authorizationService.requirePermission(
-          userId,
+          isParentForChildRun ? targetUserId : requesterUserId,
           FgaRelation.CAN_CREATE_RUN,
           `${FgaType.ADMINISTRATION}:${body.administrationId!}`,
         );
       }
+
+      if (isParentForChildRun) {
```
amyyeung17: What are your thoughts about moving this inside the if statement on line 142?

apps/backend/src/routes/runs.ts
```diff
@@ -17,15 +17,20 @@ export function registerRunsRoutes(routerInstance: Router) {
...
     event: {
       // @ts-expect-error - Express v4/v5 types mismatch in monorepo
       middleware: [AuthGuardMiddleware],
       handler: async ({ req, params, body }) =>
-        RunsController.event({ userId: req.user!.userId, isSuperAdmin: req.user!.isSuperAdmin }, params.runId, body),
+        RunsController.event(
+          { userId: req.user!.userId, isSuperAdmin: req.user!.isSuperAdmin },
+          params.userId,
+          params.runId,
+          body,
```
amyyeung17: If I recall correctly, I think we're trying to adapt this pattern moving forward: https://github.com/yeatmanlab/roar-dashboard/pull/1702#discussion_r3127128348

apps/backend/src/repositories/family.repository.ts
```diff
@@ -26,6 +26,20 @@ export class FamilyRepository extends BaseRepository<Family, typeof families> {
...
+   * Get family IDs for a specific user (active memberships only).
+   *
+   * @param userId - The user whose families to retrieve
+   * @returns Array of family IDs the user actively belongs to
+   */
+  async getFamilyIdsForUser(userId: string): Promise<string[]> {
+    const rows = await this.db
+      .select({ familyId: userFamilies.familyId })
+      .from(userFamilies)
+      .where(and(eq(userFamilies.userId, userId), isNull(userFamilies.leftOn)));
```
amyyeung17, line 39: Again, another almost impossible case, should we consider checking if a user's join date is in the future?

amyyeung17 (changes requested):

packages/authz/authorization-model.fga
Emily-ejag: Maybe, happy to discuss this during BERF! Since I already covered this in `runs.integration.test.ts` maybe it will be redundant?

packages/api-contract/src/v1/runs/index.ts
Emily-ejag: Yeah you are right the explicit export is redundant! I will remove it

packages/authz/authorization-model.fga
Emily-ejag: After BERF we agreed on create this tests

apps/backend/src/test-support/fga/fga-test-tuples.helper.ts
Emily-ejag: good catch!

apps/backend/src/services/run/run.service.ts
Emily-ejag: yes, it makes more sense logically to have it inside

apps/backend/src/services/run/run.service.ts
Emily-ejag: I think the `verifyUserAccess` already handles it

apps/backend/src/routes/runs.ts
Emily-ejag: I think I am already following the pattern, unless there is something specific I am missing -- I can't go to a direct comment with that link. Can you point me out please?

apps/backend/src/services/run/run.service.ts
Emily-ejag: I think we use `isSuperAdmin: false` for cross-user runs. If the target user is a super admin, this is an interesting edge case (child account with super admin privileges), but I think that would be caught by `verifyAdministrationAccess` when it checks the target user's permissions?

apps/backend/src/repositories/family.repository.ts
Emily-ejag: I think like you mentioned the `joinedOn` field has a database default of now(), so it will always be set to the current timestamp when a record is created. I think it can never be in the future unless someone manually updates it?

apps/backend/src/routes/runs.ts
amyyeung17: This is what @ksmontville suggested for the route in my PR:

```
export function registerFamiliesRoutes(routerInstance: Router) {
  const FamiliesRoutes = s.router(FamiliesContract, {
    listUsers: {
      // @ts-expect-error - Express v4/v5 types mismatch in monorepo
      middleware: [AuthGuardMiddleware],
      handler: async ({ req: { user }, params: { familyId }, query }) =>
        FamiliesController.listUsers(user!, familyId, query),
    },
 });
```

Emily-ejag: Oh got it!

amyyeung17 (approved):

richford (approved):

PR #1724, opened 2026-04-29
Add `POST /v1/users`

ksmontville:
## Summary

> [!WARNING]
> There is currently a significant authorization gap in this PR: any authenticated user can currently add members to any family if they have the `uuid` of that family entity. This issue must be addressed in a chained PR which closes this gap.

This PR adds the `POST /v1/users` endpoint, which creates a new user account atomically across Firebase Auth, the core database, and the OpenFGA authorization model. It also introduces the `platform_admin` role and the `can_create_users` FGA permission that gates this endpoint for non-super-admin callers.

## Endpoint

```
POST /v1/users
```

**Request body:** email, password, name, optional demographics/identifiers, and a `memberships` array (at least one entry required).

**Response:** `201 Created` with `{ data: { id } }`.

## Authorization behavior

- **Super admin:** allowed unconditionally — skips FGA and only verifies that class entity IDs exist.
- **`platform_admin`:** must hold `can_create_users` on every entity in the `memberships` array:
  - `district` / `school`: checked directly against the entity.
  - `class`: checked against the parent school (the FGA model defines `can_create_users` on `school`, not `class`).
  - `group`: checked directly against the group — groups have no hierarchy, so `platform_admin` on district does **not** inherit to groups; a `platform_admin` on group A cannot create users in group B.
  - `family`: no `can_create_users` check — families are self-managed.
- All other roles: `403 Forbidden`.

## Operation sequence

```mermaid
sequenceDiagram
    participant Client
    participant Service
    participant FGA
    participant Firebase
    participant DB

    Client->>Service: POST /v1/users
    Service->>FGA: requirePermission(can_create_users) per membership
    Service->>DB: existsByUniqueFields (email/pid conflict pre-flight)
    Service->>Firebase: getUserByEmail (Firebase conflict pre-flight)
    Service->>Firebase: createUser
    Service->>DB: createWithMemberships (transaction: users + junction rows)
    Service->>FGA: writeTuples (membership tuples)
    Service-->>Client: 201 { data: { id } }
```

On DB transaction failure, the service compensates by deleting the Firebase Auth account before re-throwing.

## Other changes

- **`platform_admin` role** added to `userRoleEnum`, `UserRole`, and `UserRoleSchema`.
- **`can_create_users` FGA relation** added to the authorization model and `FgaRelation` enum.
- **`isFirebaseError` type guard** (`src/types/firebase.ts`) — narrows `unknown` catch values to `{ code: string }` safely, replacing `as { code?: string }` assertions in the service.
- **`FirebaseErrorCode` enum** (`src/enums/firebase-error-codes.enum.ts`) — replaces raw `'auth/user-not-found'`, `'auth/email-already-exists'`, and `'auth/too-many-requests'` string literals.

## Tests

**Route integration (`users.integration.test.ts` — `POST /v1/users`):**

- Authorization: super admin, platform_admin at district/school/class, platform_admin on group (direct), group-level platform_admin cross-group denial, cross-district denial, admin/educator/unauthenticated denials
- Conflict detection: DB duplicate email (409), Firebase duplicate email (409)
- Entity validation: non-existent district/class entityId (422)
- Request validation: missing required fields, empty memberships array, short password (400)

**Repository integration (`user.repository.integration.test.ts`):**

- `createWithMemberships`: org-only, class+group, all four junction types (org+class+group+family), rollback on invalid entityId, duplicate email constraint

**Service unit (`user.service.create.test.ts`):** authorization branching, Firebase error handling, DB compensation paths.

Resolves #1774

## Types of changes

- [x] Documentation Update
- [ ] Repository Maintenance

## Checklist


ksmontville:
@claude Let's review this PR, focusing on database integrity, FGA coverage, and ensuring that unit and integration tests capture common edge cases and scenarios for user creation, specifically that users in one entity type are restricted from creating users in other entity types where FGA governs such an interaction.

ksmontville:
@claude Feedback addressed, let's review again.

We have identified an authorization gap regarding families (see warning in PR description). This PR is already too large to incorporate those needed changes, and so they will be addressed in a PR chained onto this one. See the TODO in user.servive.ts.

ksmontville:
@claude Fixed, let's review again.

ksmontville:
@claude Feedback addressed, let's re-review.

ksmontville:
@claude Feedback implemented, let's review again. Also search for regressions that may have appeared in the recent rebase commit onto enh/1774/fga-can-create-user.

ksmontville:
@claude Suggestions implemented, let's review again.

ksmontville:
@claude The last nitpicks have been addressed. We also performed a large rebase onto project/backend-refactor (the base branch) so let's check again for regressions.

apps/backend/src/services/user/user.service.ts
```diff
@@ -245,6 +337,453 @@ export function UserService({
...
+    //
+    // Entity existence for non-super-admins is checked implicitly: findClassParentSchool
+    // returns null for a non-existent class, and the FGA model requires district/school/group
+    // to have tuples so requirePermission throws 403 for missing entities.
+    //
+    // Super admins skip FGA but get explicit entity existence checks for all membership
+    // types — without this, an invalid ID would only fail at the DB FK constraint after
+    // Firebase has already created an account that then needs compensating deletion.
+
+    if (!isSuperAdmin) {
```
richford, line 382:
```suggestion
    if (!isSuperAdmin) {
      for (const m of body.memberships) {
        if (isOrgMembership(m) && m.role === UserRole.PLATFORM_ADMIN) {
          logger.warn({ userId, requestedRole: m.role }, 'Non-super admin attempted to grant platform_admin');
          throw new ApiError(ApiErrorMessage.FORBIDDEN, {
            statusCode: StatusCodes.FORBIDDEN,
            code: ApiErrorCode.AUTH_FORBIDDEN,
            context: { userId },
          });
        }
      }
    }
```

And tests to add: super-admin can grant platform_admin (success); non-super platform_admin attempting to grant platform_admin returns 403 even within their own scope.

packages/api-contract/src/v1/users/schema.ts
```diff
@@ -49,6 +52,95 @@ export const UserResponseSchema = z.object({
...
+    password: z.string().min(8),
+    name: CreateUserNameSchema,
+    userType: UserTypeSchema.default('student'),
+    dob: z.string().date().nullable().optional(),
+    grade: UserGradeSchema.nullable().optional(),
+    demographics: CreateUserDemographicsSchema.optional(),
+    identifiers: CreateUserIdentifiersSchema.optional(),
+    memberships: z.array(UserMembershipSchema).min(1),
+  })
+  .strict();
```
richford, line 134: The schema correctly excludes `isSuperAdmin` from `CreateUserRequestBodySchema` and uses `.strict()` to reject unknown keys, so a request body with `isSuperAdmin: true` fails Zod validation before reaching the service.

Can we add a test for this so that a future refactor doesn't silently regress this? Suggestion: test the invariant in `apps/backend/src/routes/users.integration.test.ts`:

```ts
it('rejects request body with isSuperAdmin field', async () => {
  await expectRoute('POST', '/v1/users')
    .as(tiers.platformAdmin)
    .withBody({
      ...validCreateUserBody,
      isSuperAdmin: true, // must not be settable via the API
    })
    .toReturn(StatusCodes.BAD_REQUEST);
});
```

apps/backend/src/repositories/user.repository.ts
```diff
@@ -100,4 +100,107 @@ export class UserRepository extends BaseRepository<User, typeof users> {
...
+  /**
+   * Check whether a user with the given email, username, or assessmentPid already exists.
+   *
+   * Used for pre-flight uniqueness checks before writing to Firebase Auth or the DB,
+   * to surface 409 Conflict early without initiating any external writes.
+   *
+   * @param params - At least one of email, username, or assessmentPid must be provided
+   * @returns true if a matching row exists
+   */
+  async existsByUniqueFields({
```
richford, line 188: This query doesn't filter on `users.rosteringEnded`, so a rostered-out user's email and PID continue to be reserved 

We need to decide if that's what we want. It's the safer behavior (no accidental re-registration of an un-rostered user's identity), but it's not documented or tested anywhere.

I don't think you need to settle this product question right now. But I have two requests:

1. Document the contract on this method: "Treats rostered-out users as still occupying their unique fields; a re-register attempt with the same email or PID returns 409 CONFLICT."
2. Pin it with an integration test: create a user, set `rosteringEnded`, attempt to create a new user with the same email, assert 409. Without this test, a future query refactor that adds `isNull(users.rosteringEnded)` would silently change behavior and reopen email reuse without anyone noticing.

apps/backend/src/services/user/user.service.ts
```diff
@@ -245,6 +337,453 @@ export function UserService({
...
+    return { id: newUserId };
+  }
+
+  /**
+   * Verifies that a membership entity exists for the given entity type and ID.
+   * @param entityType The type of the membership entity to verify.
+   * @param entityId The ID of the membership entity to verify.
+   * @returns A promise that resolves to `true` if the entity exists, `false` otherwise.
+   */
+  async function verifyMembershipEntityExists(entityType: EntityType, entityId: string): Promise<boolean> {
```
richford, line 738: `verifyMembershipEntityExists` and the per-membership repository calls used during create both treat rostered-out entities as still valid membership targets. None of these filter on `rosteringEnded`:

- `districtRepository.getById` → extends `BaseRepository.getById`, plain `select().from(orgs).where(eq(id, params.id)).limit(1)`.
- `schoolRepository.getById` → same BaseRepository implementation.
- `groupRepository.getById` → same.
- `familyRepository.getById` → same.
- `userRepository.findClassParentSchool` → returns `classes.schoolId` directly. `classes.schoolId` is `notNull` with `onDelete: 'restrict'`, so `null` only means the class doesn't exist. The query filters on neither `classes.rosteringEnded` nor the parent school's `rosteringEnded`.

Practical effect: a super-admin (whose path runs through `verifyMembershipEntityExists`) can create a user with a membership at any retired district, school, class, group, or family. No 422 fires. A non-super-admin's gate is the FGA `can_create_users` check, which only enforces this if the rostering-end flow tears down the corresponding FGA tuples.

I'd like to recommend the following policy: retired entities should not accept new memberships. This is a uniform decision across all five entity types. Suggested changes:

- Add a `getActiveById` variant on `BaseRepository` (or extend `getById` with an option), filtering `isNull(<table>.rosteringEnded)`. Migrate `verifyMembershipEntityExists` to use the active variant for all four entity types it calls (`district`, `school`, `group`, `family`). Existing call sites that legitimately need to fetch retired entities (e.g., the rostering-end flow itself, audit pages) keep using getById.
- Update `userRepository.findClassParentSchool` to filter `isNull(classes.rosteringEnded)` AND require the parent school be active, either via an inner join checking `orgs.rosteringEnded IS NULL`, or by returning the school row and letting the service check. Either way, null from this method should mean "class doesn't exist OR class/school is retired" with a single 422 response covering both.
- For the non-super-admin district/school/group path, the only authoritative enforcement is FGA. We're kind of relying on the rostering-end flow (which we haven't written yet) to ensure that the relevant tuples are deleted. Can you also add a pre-flight `verifyMembershipEntityExists` call in the non-super-admin branch too. Same as the super-admin branch, using the active variant, so application code provides defense-in-depth regardless of FGA state.
- The non-super-admin path (the FGA requirePermission calls) only enforces this transitively, via tuple cleanup. If rostering-end doesn't currently delete the entity's can_create_users-relevant tuples, that's a separate gap. Worth a quick check that the rostering-end flow does what we'd expect here, even though it's outside this PR's diff.
- Integration tests for each entity type:
  - "create user fails with 422 when district is rostered out"
  - "create user fails with 422 when school is rostered out"
  - "create user fails with 422 when class is rostered out"
  - "create user fails with 422 when class's parent school is rostered out"
  - "create user fails with 422 when group is rostered out"
  - "create user fails with 422 when family is rostered out"

richford (changes requested):

richford (approved):

kellyel (approved):

Emily-ejag (approved):

PR #1726, opened 2026-04-30
Add OIDC authentication to OpenFGA client

maximilianoertel:
## Summary

This PR adds OIDC authentication to the OpenFGA client so it can authenticate with Google identity tokens in deployed environments (dev, staging, prod). Locally, FGA continues to run without auth, no change needed for local dev or tests.

Unfortunately, the SDK's built-in credential options don't work with Google identity tokens, so this uses the SDK's second constructor parameter to inject a custom Axios instance with a request interceptor that attaches auto-refreshing tokens via `google-auth-library`.

### How it works
- A new `FgaClient.initialize()` async method is called at server startup
- When `FGA_OIDC_AUDIENCE` is set, it creates an OIDC-authenticated Axios instance and pre-initializes the singleton
- Else, when `FGA_OIDC_AUDIENCE` is not set (e.g. in local dev), the client is lazily created without auth on first `getClient()` call
- `getClient()` remains synchronous, zero changes to any of the 9 consuming services

<br>

Resolves https://github.com/yeatmanlab/roar-project-management/issues/1779


maximilianoertel:
@claude review

maximilianoertel:
> ### 🔴 Security: silent unauthenticated fallback on missing token
> **`apps/backend/src/clients/oidc-axios.client.ts` — interceptor**
> 
> ```ts
> if (!authorization) {
>   logger.warn({ audience }, 'OIDC identity token missing ..., request will be unauthenticated');
> } else {
>   config.headers.set('Authorization', authorization);
> }
> return config; // request proceeds without auth
> ```
> 
> If `getRequestHeaders()` returns no `Authorization` value, the FGA request goes out without a token. In a security-sensitive context (authenticating to OpenFGA), this silent degradation is risky: if the token is missing, we should reject the request rather than send it unauthenticated. A misconfigured environment could silently disable authorization checks instead of surfacing a startup error.
> 
> Consider throwing here instead:
> 
> ```ts
> if (!authorization) {
>   throw new Error('OIDC identity token missing — cannot send unauthenticated FGA request');
> }
> config.headers.set('Authorization', authorization);
> return config;
> ```
> 
> The error would be caught by Axios and reject the in-flight FGA call (which would surface as a 500 in the service layer), rather than letting an unauthenticated request through. 


The risk described here is pretty overstated. If the token is missing, FGA simply rejects the request with a 401. An unauthenticated request can't "silently disable authorization checks" because FGA enforces its own auth server-side.

That said, failing early does give a clearer error than a cryptic FGA 401. I've added a `NODE_ENV === 'production'` guard: in production builds, a missing token throws immediately; otherwise it just warns and lets FGA reject the request naturally.

maximilianoertel:
@claude re-review

richford (approved):

ksmontville (approved):

PR #1734, opened 2026-05-01
`SDK` Changelog Strategy 

Emily-ejag:
## Proposed changes

This PR establish a changelog and release strategy for the ROAR Dashboard monorepo using low-friction tooling aligned with our current workflow, while supporting hybrid versioning

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

Ref https://github.com/yeatmanlab/roar-project-management/issues/1756


Emily-ejag:
@claude please review this PR

Emily-ejag:

Emily-ejag:
@claude for the `TODO` I will keep it disabled with a tracking issue to enable it once the front end deployment is ready to go

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,34 @@
...
+      api_contract_release_created: ${{ steps.release.outputs.api-contract--release_created }}
+      assessment_sdk_release_created: ${{ steps.release.outputs.assessment-sdk--release_created }}
+      backend_tag_name: ${{ steps.release.outputs.backend--tag_name }}
+      dashboard_tag_name: ${{ steps.release.outputs.dashboard--tag_name }}
+      api_contract_tag_name: ${{ steps.release.outputs.api-contract--tag_name }}
+      assessment_sdk_tag_name: ${{ steps.release.outputs.assessment-sdk--tag_name }}
+    steps:
+      - name: Run Release Please
+        id: release
+        uses: googleapis/release-please-action@v4
```
maximilianoertel: Why use the outdated v4 and not the latest v5?

.github/workflows/deploy-platform-production.yml
```diff
@@ -2,14 +2,15 @@ name: Deploy Platform to Production
...
-    #   - "roar-platform:v[0-9]+.[0-9]+.[0-9]+-beta.[0-9]+"
+    tags:
+      # Platform releases: trigger on backend tag only.
+      # Release Please creates all three tags (backend, dashboard, api-contract) simultaneously,
+      # but only backend needs infrastructure deployment. api-contract is a TypeScript package
+      # published to npm, not deployed to Cloud Run. Dashboard is deployed as part of the
+      # backend release. Triggering on backend-v* only prevents triple deployments.
+      - "backend-v[0-9]+.[0-9]+.[0-9]+"
+      - "backend-v[0-9]+.[0-9]+.[0-9]+-alpha.[0-9]+"
+      - "backend-v[0-9]+.[0-9]+.[0-9]+-beta.[0-9]+"
```
maximilianoertel: This works but feels quite hacky. Instead, I would recommend using the `workflow_call` functionality. 

```
on: 
  workflow_call:
    inputs:
      version:
        type: string
        required: true
```

The `verify-tag` job in this production workflow would need to be adjusted a little bit as it currently only expects a single tag, but should now verify that all three tags with the `inputs.version` is available on main. Similarly, the `Create Sentry release` step also needs to be updated to use `inputs.version`

Then, I would rename the `release-please.yml` to `release.yml` and add an additional step after the `release-please` to call the platform production deployment workflow, conditional on `backend_release_created`, `dashboard_release_created`, and `api_contract_release_created` being true

maximilianoertel (changes requested):

.github/workflows/release-please.yml
Emily-ejag: Good point, I chose v4 because of my node version (22) and v5 requires 24, but I see that GitHub Actions will handle it automatically, so I agree with increasing it to v5

Emily-ejag:
@claude can you check if I have solved @maximilianoertel 's comments and did not introduce any new error?

Emily-ejag:
@claude what about now?

Emily-ejag:
@claude what about now?

Emily-ejag:
@claude I have kept the inputs and added a step that logs which components were released.

.github/workflows/deploy-platform-production.yml
Emily-ejag: I totally agree with this, I think my approach was simpler but messier. I prefer this approach, thanks

.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
...
+      needs.release-please.outputs.backend_release_created == 'true' &&
+      needs.release-please.outputs.dashboard_release_created == 'true' &&
+      needs.release-please.outputs.api_contract_release_created == 'true'
+    uses: ./.github/workflows/deploy-platform-production.yml
+    secrets: inherit
+    with:
+      version: ${{ needs.release-please.outputs.backend_version }}
+      backend_release_created: ${{ needs.release-please.outputs.backend_release_created == 'true' }}
+      dashboard_release_created: ${{ needs.release-please.outputs.dashboard_release_created == 'true' }}
+      api_contract_release_created: ${{ needs.release-please.outputs.api_contract_release_created == 'true' }}
```
maximilianoertel: These are redundant since the job condition requires all three to be evaluate to `true`. In other words, `backend_release_created`, `dashboard_release_created` and `api_contract_release_created` will always be `true` when `deploy-platform-production` runs 


```suggestion
```

.github/workflows/deploy-platform-production.yml
```diff
@@ -1,15 +1,31 @@
...
+        type: boolean
+        required: true
+      dashboard_release_created:
+        description: "Whether dashboard was released"
+        type: boolean
+        required: true
+      api_contract_release_created:
+        description: "Whether api-contract was released"
+        type: boolean
+        required: true
```
maximilianoertel: See https://github.com/yeatmanlab/roar-dashboard/pull/1734/changes#r3191647296


```suggestion
```

.github/workflows/deploy-platform-production.yml
```diff
@@ -1,15 +1,31 @@
...
-    # tags:
-    #   - "roar-platform:v[0-9]+.[0-9]+.[0-9]+"
-    #   - "roar-platform:v[0-9]+.[0-9]+.[0-9]+-alpha.[0-9]+"
-    #   - "roar-platform:v[0-9]+.[0-9]+.[0-9]+-beta.[0-9]+"
+    tags:
+      # Fallback: trigger on backend tag for manual releases or if release-please.yml is not updated.
+      # Normally, release-please.yml calls this workflow via workflow_call with version inputs.
+      - "backend-v[0-9]+.[0-9]+.[0-9]+"
+      - "backend-v[0-9]+.[0-9]+.[0-9]+-alpha.[0-9]+"
+      - "backend-v[0-9]+.[0-9]+.[0-9]+-beta.[0-9]+"
```
maximilianoertel: These tag triggers need to removed otherwise the workflow will trigger twice: once on tag push and once on workflow call
```suggestion
    tags:
```

.github/workflows/deploy-platform-production.yml
```diff
@@ -31,31 +47,52 @@ jobs:
...
-          TAG_SHA="${{ github.sha }}"
-          if ! git merge-base --is-ancestor "$TAG_SHA" origin/main; then
-            echo "::error::Tag ${{ github.ref_name }} (${TAG_SHA}) is not on the main branch. Aborting."
-            exit 1
-          fi
-          echo "Tag ${{ github.ref_name }} is on main."
+          # Determine version: use inputs.version for workflow_call, github.ref_name for push trigger
+          VERSION="${{ inputs.version || github.ref_name }}"
+          
+          # Remove tag prefix if present (e.g., "backend-v3.27.0" -> "3.27.0")
```
maximilianoertel:
```suggestion
          # Determine version
          VERSION="${{ inputs.version }}"
```

.github/workflows/deploy-platform-production.yml
```diff
@@ -185,4 +222,4 @@ jobs:
         with:
           environment: production
           sourcemaps: ./app/dist/assets/*.js.map
-          version: ${{ github.ref_name }}
\ No newline at end of file
+          version: ${{ inputs.version || github.ref_name }}
```
maximilianoertel:
```suggestion
          version: ${{ inputs.version }}
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
+name: Release Please
```
maximilianoertel:
```suggestion
name: Release
```

maximilianoertel (changes requested):
Thanks for the latest changes. 

In addition to the further change requests below, there is two more issues that need to be pointed out. 

#### Strategy
The PR targets [project/backend-refactor](https://github.com/yeatmanlab/roar-dashboard/tree/project/backend-refactor) but the the `release.yml` workflow will only trigger on `main`. 

So what this PR is doing is correctly preparing for the new release workflow post backend refactor but in the meantime breaking our continuous backend deployments to production. 

Yet, merging to main instead would mean having to adopt the new release workflows now. Not necessarily a problem, but means additional changes to the deploy workflows, deprecating the existing workflows, dealing with the new tags, etc. Alternatively, we finish this PR but do not merge until we're ready to release.

What do you think? Ultimately it's @richford's decision. 


#### SDK Publishing
When we discussed the implementation details in https://github.com/yeatmanlab/roar-project-management/issues/1756, we also discussed the publishing and releasing of the Assessment SDK – actually the issue was originally just for the SDK alone and turned into a monorepo scoped one. This PR doesn't touch any of the SDK pieces though, so just double-checking that there will be a follow-up PR to handle the SDK releases?

.github/workflows/release.yml
Emily-ejag: I thought it would be useful for debugging, but it can be removed

Emily-ejag:
@claude please re-review

.github/workflows/deploy-platform-production.yml
```diff
@@ -31,31 +30,52 @@ jobs:
...
+              exit 1
+            fi
+            
+            echo "✓ Tag ${TAG} verified on main"
+          done
+
+      - name: Log release components
+        run: |
+          echo "Release Summary:"
+          echo "  Version: ${{ inputs.version || github.ref_name }}"
```
amyyeung17: Since we've deleted `github.ref_name` elsewhere, should this be removed as well?

.github/RELEASING.md
```diff
@@ -0,0 +1,416 @@
...
+| api-contract | Shared (platform) | `packages/api-contract` |
+| assessment-sdk | Independent | `packages/assessment-sdk` |
+
+**Platform Release**: When backend, dashboard, or api-contract changes are released, they share a single version number (e.g., `3.26.0`).
+
+**SDK Release**: Assessment SDK is versioned independently (e.g., `0.1.0`) and can be released separately.
+
+## Workflow
+
+### 1. Create a PR with Conventional Commit Title
```
amyyeung17, line 26: Is this referring to the individual commit messages themselves or the PR title? Based on line 64 and 185, I'm inclined to believe it's the latter?

amyyeung17 (changes requested):

richford:
Let's go with option 1, "keep the PR targeting project/backend-refactor — the new workflows sit dormant until backend-refactor merges to main. Existing production deployments are unaffected."

.github/workflows/deploy-platform-production.yml
```diff
@@ -1,15 +1,12 @@
 name: Deploy Platform to Production
 
 on:
-  push:
-    branches:
-      # TODO: Remove once backend refactor is complete
-      - project/backend-refactor
```
maximilianoertel, line 7: Given @richford's [decision to go with option #1](https://github.com/yeatmanlab/roar-dashboard/pull/1734#issuecomment-4390686293) this branch trigger needs be added back and temporarily disable the tag check so that we continue continuously deploying the backend to production during the refactor project

.github/workflows/deploy-platform-production.yml
```diff
@@ -31,31 +28,52 @@ jobs:
...
-          if ! git merge-base --is-ancestor "$TAG_SHA" origin/main; then
-            echo "::error::Tag ${{ github.ref_name }} (${TAG_SHA}) is not on the main branch. Aborting."
-            exit 1
-          fi
-          echo "Tag ${{ github.ref_name }} is on main."
+          # Determine version
+          VERSION="${{ inputs.version }}"
+          
+          # Remove tag prefix if present (e.g., "backend-v3.27.0" -> "3.27.0")
+          VERSION="${VERSION#backend-v}"
```
maximilianoertel: As pointed out previously (https://github.com/yeatmanlab/roar-dashboard/pull/1734#discussion_r3191674076) there is no longer any need to remove tag prefixes since the workflow input expects the actual semver version


```suggestion
```

.github/workflows/deploy-platform-production.yml
maximilianoertel:
```suggestion
          echo "  Version: ${{ inputs.version }}"
```

maximilianoertel (changes requested):

.github/workflows/deploy-platform-production.yml
Emily-ejag: I will add them back

.github/RELEASING.md
Emily-ejag: It's the PR title. When you merge a PR to main using squash merge, the PR title becomes the commit message. This is what Release Please reads to determine version bumps and changelog entries.

Emily-ejag:
@claude can you check if I resolved all the comments and did not introduce any new errors?

Emily-ejag:
@claude what about now?

Emily-ejag:
@claude can you re-review

.release-please.json
```diff
@@ -0,0 +1,70 @@
...
+      "component": "api-contract"
+    },
+    "packages/assessment-sdk": {
+      "changelog-path": "CHANGELOG.md",
+      "version-file": "package.json",
+      "release-type": "node",
+      "component": "assessment-sdk"
+    }
+  },
+  "group-pull-request-title-pattern": "chore: release ${version}",
```
richford, line 72: The config has `group-pull-request-title-pattern` set, which is only used when Release Please produces grouped Release PRs. I.e. when packages are linked. But there's no `plugins` array with a `linked-versions` entry, and each package has its own `component` with no link between them.

Without the `linked-versions` plugin, Release Please versions each package independently. The `release-as` field gives matched bootstrap values (3.26.0 across the three platform packages), but immediately after the first release the versions will diverge:

- A `feat(backend): ...` commit → bumps only backend to 3.27.0; dashboard and api-contract stay at 3.26.0
- A `fix(dashboard): ...` commit → bumps only dashboard to 3.26.1; backend stays at 3.27.0

After a few releases the three are at different versions despite the docs claiming they move together.

The bigger consequence: the deployment trigger condition in `release.yml:46-49` becomes unreachable in practice:

```yaml
if: |
  needs.release-please.outputs.releases_created == 'true' &&
  needs.release-please.outputs.backend_release_created == 'true' &&
  needs.release-please.outputs.dashboard_release_created == 'true' &&
  needs.release-please.outputs.api_contract_release_created == 'true'
```

Without `linked-versions`, the AND condition is true only when conventional commits scoped to all three components happen to land before the Release PR is merged, which might not happen in all release windows. The git tags get created but production isn't updated.

While we're here, also add `node-workspace`. I think it's the canonical pairing with `linked-versions` for npm workspaces. It auto-bumps inter-workspace dep declarations when a package releases. Today the deps all use `"*"` so there's nothing for the plugin to update, but if any of these are ever switched to a concrete pin (`^3.26.0`), the plugin keeps them in sync automatically. Cheap forward-compat insurance:

​```jsonc
{
  "bootstrap-sha": "3a7a1a1577fab6498d7a906d69053de2be9a2e0b",
  "plugins": [
    {
      "type": "linked-versions",
      "groupName": "platform",
      "components": ["backend", "dashboard", "api-contract"]
    },
    {
      "type": "node-workspace"
    }
  ],
  "changelog-sections": [ ... ],
  ...
}
​```

`assessment-sdk` is correctly excluded from the `components` list, so it stays independently versioned — exactly the hybrid model the team agreed on.

With `linked-versions` in place: any conventional commit scoped to `backend`, `dashboard`, or `api-contract` triggers a coordinated bump of all three to the same version, all three `_release_created` outputs fire together, the `group-pull-request-title-pattern` actually applies, and the deployment trigger condition becomes reachable.

[release-please plugin docs](https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md#plugins) — both `linked-versions` and `node-workspace` are documented there.

To test this, it's worth doing one dry-run end-to-end on a throwaway branch before this lands on main: push a few empty commits with conventional titles scoped to different platform components, watch Release Please produce a single grouped Release PR, and confirm the deployment trigger condition fires when the Release PR merges.

.release-please.json
richford, line 1: Independent of the `linked-versions` issue: `@roar-dashboard/api-contract` is in the linked-versions group, but the package isn't published anywhere. Its `package.json` has no `publishConfig`, no workflow in this PR runs `npm publish` for it. But I think the team agreed in [1756](https://github.com/yeatmanlab/roar-project-management/issues/1756) to publish it ("Option A: Publish `api-contract` alongside `assessment-sdk`").

Two consequences worth flagging now:

**1. Release tags will lie about what shipped.** With `linked-versions` enabled, every platform release creates an `api-contract-v3.26.0` tag asserting "api-contract 3.26.0 was released" while nothing actually publishes the package. The release notes auto-generated by Release Please will look correct, but a consumer hitting the registry won't find the version.

**2. The published SDK is broken for external consumers.** `@yeatmanlab/assessment-sdk` ships with `"@roar-dashboard/api-contract": "*"` in its dependencies. When external consumers run `npm install @yeatmanlab/assessment-sdk`, npm tries to resolve `@roar-dashboard/api-contract@*` against the registry, finds nothing, and fails. This is a pre-existing problem this PR doesn't introduce, but it becomes more visible once Release Please starts cutting "official" version tags for api-contract.

Three options to address this. Pick whichever you like:

- **(a)** Land the publishing pipeline in this PR. Add an `npm publish` step for api-contract to GitHub Package Registry (same as the SDK). Replace the SDK's `"@roar-dashboard/api-contract": "*"` with a real semver pin, and let `node-workspace` keep it in sync going forward.
- **(b)** Defer publishing to a follow-up PR (consistent with @maximilianoertel's earlier comment "this PR doesn't touch SDK pieces"), but **remove `api-contract` from the `linked-versions` components list for now**. That way no `api-contract-v*` tags get created until the publishing pipeline is in place.
- **(c)** Defer publishing AND keep `api-contract` in `linked-versions`, but add a section to `RELEASING.md` documenting that `api-contract` release tags are aspirational during the transition, with a link to the tracking issue. This is the lightest-touch option but leaves the SDK broken for external consumers until the follow-up lands.

richford (changes requested):

.release-please.json
Emily-ejag: @richford thank you I didn't think about this issue. After a few releases, the three platform packages CAN have different versions and the the platform release will not be reachable. I will add the plugins to have the versions bump at the same time for the three components. I will add the "cheap forward-compat insurance" that you mentioned. Thanks!

Emily-ejag:
## **Test:**
- Create a throwaway branch off of this branch and create PR changing the push branch to that one.
- Change the trigger release to NOT release a new version with `if: | github.ref == 'ref/hads/amin &&`
- Push empty commits with conventional titles to simulate changes in the components, one for each component
- Wait for the release please to run. Check the GitHub Actions, verify grouped release, titles, all components are bumped together, conditional triggering.

Emily-ejag:
From @richford : 

### **Check:**

- The release-please job fires, scans commits since `bootstrap-sha` on `release-test`, opens a Release PR against release-test 
- The Release PR title and body match what you expect
- Per-component changelog files get the right entries grouped under “Features” / “Bug Fixes”
- The Release PR’s diff shows version bumps for backend, dashboard, api-contract (all to the same version if linked-versions since is configured)
- `assessment-sdk` appears in its own separate Release PR (or block) — confirms hybrid versioning works
- `trigger-platform-deployment` is skipped because of the `github.ref == 'refs/heads/main'` gate

### **Then test the second half (tag creation):**

- Merge the release PR (or release PRs if there are multiple)
- The deployment workflow still shouldn’t fire because of the ref gate.
- Verify that it created new tags

### **Then cleanup:**

- Remove test tags
- Delete the test branch
- Revert workflow changes

.release-please.json
Emily-ejag: Thanks! I think I will go with **b**

Emily-ejag:
@claude can you review if I solved @richford's comments and I did not introduce any new issue?

Emily-ejag:
@claude what about now?

richford (approved):
Thanks @Emily-ejag!

Sorry for sending you down the rabbit hole with the `node-workspace` suggestion and good catch on removing it. This looks good to me. Can you please open two follow-on tickets:

- Update the `bootstrap-sha` before the first actual release
- Create the `api-contract` publishing pipeline

amyyeung17 (approved):

PR #1736, opened 2026-05-04
Add `PATCH /administrations/:id`

kellyel:
## Proposed changes
This PR implements the `PATCH /administrations/:id` endpoint. Specifications can be found here: https://github.com/yeatmanlab/roar-project-management/issues/1559

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist


kellyel:

kellyel:
@claude please re-review this PR.

kellyel:
@claude please review this PR once more.

kellyel:
@claude, please re-review this PR. Note that I have not removed the 400: Bad Request as a possible status from the contract. You are correct that it does not get thrown from the service, however it is thrown by Zod if the request does not align with the specified schema. As a result, this is a possible return status.

apps/backend/src/services/administration/administration.service.ts
```diff
@@ -1453,6 +1455,432 @@ export function AdministrationService({
...
+        const { items: existingDistricts } = await districtRepository.listByIds(request.orgs, {
+          page: 1,
+          perPage: request.orgs.length,
+        });
+        const { items: existingSchools } = await schoolRepository.listByIds(request.orgs, {
+          page: 1,
+          perPage: request.orgs.length,
+        });
+        newDistrictIds = existingDistricts.map((d) => d.id);
+        newSchoolIds = existingSchools.map((s) => s.id);
```
amyyeung17: At first pass, I was slightly confused as to how `newDistrictIds` and `newSchoolIds` relate to the existing ids and their interactions with `missingOrgs`. Could we clarify the naming here?

apps/backend/src/services/administration/administration.service.ts
```diff
@@ -1453,6 +1455,432 @@ export function AdministrationService({
...
+        const existingClassIdSet = new Set(existingClasses.map((c) => c.id));
+        const missingClasses = request.classes.filter((id) => !existingClassIdSet.has(id));
+        if (missingClasses.length > 0) {
+          throw new ApiError(ApiErrorMessage.REQUEST_VALIDATION_FAILED, {
+            statusCode: StatusCodes.UNPROCESSABLE_ENTITY,
+            code: ApiErrorCode.REQUEST_VALIDATION_FAILED,
+            context: { userId, administrationId, missingClasses },
+          });
+        }
+      }
```
amyyeung17, line 1647: It seems like we repeat this logic in several places (create set of ids, validating if org/class/groups exist, throw error if any are missing). What are your thoughts on maybe extracting a utility function?

apps/backend/src/services/administration/administration.service.test.ts
```diff
@@ -4321,4 +4321,583 @@ describe('AdministrationService', () => {
       });
     });
   });
+
+  describe('update', () => {
```
amyyeung17, line 4325: Would it make sense to add some tests for partial update cases? One case I had in mind is verifying that requested deletions are applied correctly while preserving other existing values.

apps/backend/src/services/administration/administration.service.ts
```diff
@@ -1453,6 +1455,432 @@ export function AdministrationService({
...
+      if (!existing) {
+        throw new ApiError(ApiErrorMessage.NOT_FOUND, {
+          statusCode: StatusCodes.NOT_FOUND,
+          code: ApiErrorCode.RESOURCE_NOT_FOUND,
+          context: { userId, administrationId },
+        });
+      }
+
+      // Check authorization (super admin only for updates)
+      if (!isSuperAdmin) {
```
amyyeung17, line 1506: Hm, the [ticket](https://github.com/yeatmanlab/roar-project-management/issues/1559#:~:text=District%20admins%3A%20Can%20update%20administrations%20for%20orgs%20they%20have%20access%20to) says that district admins can update orgs they have access to. Did we decide to limit this privilege to super admin?

amyyeung17 (changes requested):

apps/backend/src/services/administration/administration.service.ts
kellyel: This is a super admin only endpoint, I will update the card.

amyyeung17 (dismissed):

amyyeung17 (dismissed):

Emily-ejag (dismissed):

apps/backend/src/services/utils/validations.utils.ts
richford, line 1: This file's location makes it seem like a repository util. A util should belong to the lowest layer that consumes it. But `verifyEntitiesExist` has no repository concerns: it's a pure function over `{ id: string }[]` arrays that throws an `ApiError`. No Drizzle, no DB client, no SQL. And `git grep verifyEntitiesExist` shows the only consumer is `services/administration/administration.service.ts`. So I think this is a service helper and not a repository helper.

Could you move this to a service-layer location? E.g., `apps/backend/src/services/utils/validations.utils.ts` (or `apps/backend/src/services/administration/utils/validations.utils.ts` if you'd rather keep it administration-scoped until a second consumer materializes). The unit tests at validations.utils.test.ts move alongside the file.

I'd lean toward making it available to all services. It's likely that the user creation flow and probably future endpoints will want this too.

richford (changes requested):
Looks really good. Just one nitpick about the utility placement. See below.

richford (dismissed):

Emily-ejag (dismissed):

Emily-ejag (approved):

richford (approved):

PR #1740, opened 2026-05-07
Add LtreeRepository for ltree-based hierarchy queries

richford:
## Summary

This PR adds an `LtreeRepository<TEntity, TTable>` abstract class that extends `BaseRepository` with operations for tables whose rows participate in a PostgreSQL `ltree` hierarchy. The first method is `getDistinctRootOrgIds(ids)` — for a list of node ids, return the unique set of root org ids covering those nodes.

`DistrictRepository`, `SchoolRepository`, and `ClassRepository` are migrated to extend `LtreeRepository` instead of `BaseRepository`. All other repositories (users, runs, agreements, tasks, etc.) continue to extend `BaseRepository` directly — the new class is opt-in for repositories whose driving table has an ltree column.

This is **scaffolding for a follow-up PR**: there is no consumer of `getDistinctRootOrgIds` in this PR. The expected first consumer is in input validation for `POST /users`, which needs to resolve a set of membership district, school, and class IDs back to their roots orgs. Landing the abstraction now keeps that follow-up PR focused on access-control logic.

## Why an opt-in subclass instead of a method on BaseRepository

`BaseRepository` is the universal base. Many repositories (`UserRepository`, `RunRepository`, `AgreementRepository`, etc.) drive tables that have no path column. Adding `getDistinctRootOrgIds` to `BaseRepository` would either force a structural constraint that those tables can't satisfy, or push column existence checks to runtime. Putting the method on a separate `LtreeRepository extends BaseRepository` keeps the universal base honest: only repositories whose driving table actually has an ltree column get the method.

## Why "RootOrgIds" not "RootIds"

In this codebase, the top of every ltree path is always a row in `app.orgs`. Districts root in orgs (themselves). Schools root in orgs (their district or higher). Classes root in orgs (the district above their school). The method name reflects that invariant. Callers get back `orgs.id` values, not generic "root row" ids that could mean anything.

This matters because the path label format encodes the org type: a national-level rostering root has a path like `national_<uuid>`, a district-level root has `district_<uuid>`. The method does not promise that the returned roots are districts. They are whichever org sits at the rostering top of each input's path. Today that's almost always a district; the schema permits other types in the future.

## How the SQL works

For each input id, the node's "root" is the row in `ancestorTable` whose path equals the first label of the node's path. PostgreSQL's `subpath(ltree, 0, 1)` extracts that first label. The query is a self-join (or a cross-table join, for classes) on path equality, with `selectDistinct` to deduplicate roots across multiple inputs.

```sql
SELECT DISTINCT root.id
FROM <table>
INNER JOIN orgs root ON root.path = subpath(<table>.<pathColumn>, 0, 1)
WHERE <table>.id = ANY($1);
```

The Drizzle implementation uses `alias()` to give the join target a distinct table identifier. Without it, a self-join on `orgs` would collapse both references to the same alias and the ON clause would be ambiguous.

GiST indexes on `orgs.path` and `classes.org_path` already exist (`orgs_path_gist_idx`, `classes_org_path_gist_idx`), so the underlying scans are well-supported.

## Edge cases

- **Empty `ids`**: short-circuits to `[]` without round-tripping. `inArray([])` would otherwise generate `false` and run a guaranteed-empty query.
- **Duplicate input ids**: `selectDistinct` returns each root once.
- **Nonexistent input ids**: silently dropped by the `inArray` filter; no error.
- **Orphan paths** (root label not present in `ancestorTable`): silently dropped by the inner join. This shouldn't occur in practice (the `trg_orgs_compute_path_insert` and `trg_classes_compute_org_path_insert` triggers maintain referential integrity), but the JSDoc documents it explicitly so a future contributor doesn't have to reverse-engineer the join semantics.

### District-rooted lookup (district → itself)

```typescript
const districts = new DistrictRepository();

// A single district resolves to itself — its path is a single label.
const result = await districts.getDistinctRootOrgIds([baseFixture.district.id]);
// → [{ id: baseFixture.district.id }]
```

### School-rooted lookup (school → parent district)

```typescript
const schools = new SchoolRepository();

// Schools across a single district collapse to one root.
const result = await schools.getDistinctRootOrgIds([
  baseFixture.schoolA.id,
  baseFixture.schoolB.id,
]);
// → [{ id: baseFixture.district.id }]
```

### Class-rooted lookup (class → root district via orgs join)

```typescript
const classes = new ClassRepository();

// Classes spanning two districts return two distinct roots.
const result = await classes.getDistinctRootOrgIds([
  baseFixture.classInSchoolA.id,
  baseFixture.classInDistrictB.id,
]);
// → [{ id: baseFixture.district.id }, { id: baseFixture.districtB.id }]
```

## What this PR does NOT add

- **`getRootIdsForIds(ids)`**: the input→root mapping variant (preserves which input maps to which root, instead of returning the distinct set). YAGNI; add when there's a concrete consumer.
- **Type-filtered roots**: e.g., "give me the root, but only if it's a district." Adds a `where` predicate hook that's speculative without a use case.
- **A `getRoots()` no-input variant**: "all roots in the table" is well-served by `WHERE parent_org_id IS NULL` for orgs; not worth a generic.

## Files changed

- **New:** `apps/backend/src/repositories/ltree.repository.ts` — abstract class with `getDistinctRootOrgIds`.
- **New:** `apps/backend/src/repositories/ltree.repository.integration.test.ts` — integration tests across all three subclasses.
- **Modified:** `apps/backend/src/repositories/district.repository.ts` — extends `LtreeRepository` instead of `BaseRepository`; constructor passes `orgs.path`.
- **Modified:** `apps/backend/src/repositories/school.repository.ts` — same migration as `DistrictRepository`.
- **Modified:** `apps/backend/src/repositories/class.repository.ts` — extends `LtreeRepository`; constructor passes `classes.orgPath, orgs, orgs.path` so roots resolve in the orgs table.

No other files change. Public API of the three subclasses is unaffected — `LtreeRepository` adds a method, doesn't remove or alter anything from `BaseRepository`.

## Tests

`ltree.repository.integration.test.ts` covers, against the real database with `baseFixture`:

**District (self-rooted)**

- Single district → returns itself.
- Multiple districts spanning branches → distinct roots.
- Duplicate input ids → returns each root once.
- Nonexistent ids → empty result, no error.

**School (orgs-rooted in orgs)**

- Single school → parent district id.
- Multiple schools sharing a district → one root.
- Schools across branches (District A and District B) → two distinct roots.
- Deeper hierarchy: a department under School A → still resolves to the District A root (verifies the path is more than two labels deep).

**Class (classes-rooted in orgs)**

- Single class → root district id.
- Multiple classes sharing a root → one root.
- Classes spanning two districts → two distinct roots.
- Newly-created class under an existing school → resolves correctly (verifies the path-computation triggers play well with the join).

**Empty-input short-circuit**

- Injects a `Proxy`-based db client that throws on every property access. Calling `getDistinctRootOrgIds([])` must return `[]` without ever touching the db. Regressions that remove the short-circuit will throw a clear error instead of silently round-tripping.

No unit tests for `LtreeRepository` itself — the SQL semantics (`subpath`, `@>`, GiST behavior) need a real database to validate, and the abstract class has no logic worth mocking out from.

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist

### DIFF
--- a/apps/backend/src/repositories/class.repository.ts
+++ b/apps/backend/src/repositories/class.repository.ts
@@ -5,14 +5,14 @@ import { and, asc, count, desc, eq, isNull, sql } from 'drizzle-orm';
 import type { NodePgDatabase } from 'drizzle-orm/node-postgres';
 import { CoreDbClient } from '../db/clients';
 import type { Class } from '../db/schema';
-import { classes, userClasses, users } from '../db/schema';
+import { classes, orgs, userClasses, users } from '../db/schema';
 import type * as CoreDbSchema from '../db/schema/core';
 import type { ClassType } from '../enums/class-type.enum';
 import type { Grade } from '../enums/grade.enum';
 import type { ParsedFilter } from '../types/filter';
 import type { EnrolledUserEntity, EnrolledUsersSortFieldType, ListEnrolledUsersOptions } from '../types/user';
 import type { PaginatedResult } from './base.repository';
-import { BaseRepository } from './base.repository';
+import { LtreeRepository } from './ltree.repository';
 import {
   ENROLLED_USERS_SORT_COLUMNS,
   getEnrolledUsersFilterConditions,
@@ -51,9 +51,9 @@ export interface CreateClassInput {
   location?: string | undefined;
 }
 
-export class ClassRepository extends BaseRepository<Class, typeof classes> {
+export class ClassRepository extends LtreeRepository<Class, typeof classes> {
   constructor(db: NodePgDatabase<typeof CoreDbSchema> = CoreDbClient) {
-    super(db, classes);
+    super(db, classes, classes.orgPath, orgs, orgs.path);
   }
 
   /**
--- a/apps/backend/src/repositories/district.repository.ts
+++ b/apps/backend/src/repositories/district.repository.ts
@@ -11,7 +11,7 @@ import { OrgType } from '../enums/org-type.enum';
 import type { UserRole } from '../enums/user-role.enum';
 import type { EnrolledUserEntity, EnrolledUsersSortFieldType, ListEnrolledUsersOptions } from '../types/user';
 import type { PaginatedResult } from './base.repository';
-import { BaseRepository } from './base.repository';
+import { LtreeRepository } from './ltree.repository';
 import {
   ENROLLED_USERS_SORT_COLUMNS,
   getEnrolledUsersFilterConditions,
@@ -87,9 +87,9 @@ export interface ListDistrictOptions {
  *
  * Handles data access for districts (orgs with orgType = 'district').
  */
-export class DistrictRepository extends BaseRepository<District, typeof orgs> {
+export class DistrictRepository extends LtreeRepository<District, typeof orgs> {
   constructor(db: NodePgDatabase<typeof CoreDbSchema> = CoreDbClient) {
-    super(db, orgs);
+    super(db, orgs, orgs.path);
   }
 
   /**
--- a/apps/backend/src/repositories/ltree.repository.integration.test.ts
+++ b/apps/backend/src/repositories/ltree.repository.integration.test.ts
@@ -1,0 +1,181 @@
+/**
+ * Integration tests for LtreeRepository.
+ *
+ * Exercises `getDistinctRootOrgIds()` through all three concrete subclasses
+ * (DistrictRepository, SchoolRepository, ClassRepository) against the real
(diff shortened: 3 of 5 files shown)


richford:

richford:
@claude, please re-review.

ksmontville (approved):

amyyeung17 (approved):

PR #1743, opened 2026-05-08
Add RunScoresRepository with natural-key upsert

richford:
## Summary

This PR is the foundation of #1784, "Write run-level scores on `trial` and `complete` run events." It adds a `RunScoresRepository` with an `upsertMany` method, plus the unique constraint and constants those upserts depend on. **No callers yet** — the contract change and `RunEventService.writeTrial` extension land in the next PR in the chain.

Splitting this off keeps the diff focused on infrastructure: schema, migration, repository, tests. The next PR can then be reviewed purely as wiring.

## Schema and migration

`app.run_scores` previously had only a UUID primary key. To support `INSERT ... ON CONFLICT DO UPDATE` on the natural key, this PR adds a `UNIQUE NULLS NOT DISTINCT (run_id, type, domain, name, assessment_stage)` constraint:

- The constraint enforces the invariant the read path in `report.repository.ts` already assumes — at most one current row per natural key. Before this, the assumption was hope-based; after, it's a foreign-key-grade guarantee.
- `assessment_stage` is nullable. Without `NULLS NOT DISTINCT`, two rows with the same `(run_id, type, domain, name)` and a `NULL` stage would coexist (Postgres treats `NULL` as distinct in unique constraints by default). `NULLS NOT DISTINCT` makes them collapse to one row, which matches CAT scoring semantics where the absence of a stage is the same "kind of" score every time.

`NULLS NOT DISTINCT` requires Postgres 15+. CI is on `postgres:18` and the IaC default in `roar-iac` is `POSTGRES_18`, so this is safe.

The migration is hand-written rather than generated. `drizzle-kit generate` couldn't run during development (sandbox tooling issue), but the schema (`run-scores.ts`), SQL (`migrations/assessment/0021_add_run_scores_natural_key_unique.sql`), snapshot (`meta/0021_snapshot.json`), and journal entry are all in place. Running `db:gen:assess` should be a no-op; if it produces a new migration file, that's drift between the hand-written snapshot and what drizzle-kit computes — flag it in review.

The migration's comment includes the pre-flight duplicate-detection query for any environment that could already have duplicates. `run_scores` is currently never written to outside test fixtures, so prod should be empty.

## Constants

`apps/backend/src/constants/run-scores.ts` introduces:

- `SCORE_TYPE` — `computed`, `raw` (mirrors the `score_type` enum)
- `SCORE_DOMAIN` — `composite` (will grow as more domains land)
- `ASSESSMENT_STAGE` — `practice`, `test` (mirrors the `assessment_stage` enum)
- `SCORE_NAME` — `THETA_SE`, `NUM_ATTEMPTED`

Promoted now because both this PR's upsert path and the upcoming `recompute use_for_reporting` ticket reference the same magic strings (`thetaSE`, `numAttempted`, `composite`, `test`). One source of truth keeps the natural key consistent across producers and consumers.

`SCORE_NAME` is intentionally narrow — only the names backend code references by string in queries. Other names live in assessment configs and don't need backend-side constants.

## Repository

`RunScoresRepository extends BaseRepository<RunScore, typeof runScores>`. Single new method:

```ts
async upsertMany(params: { data: NewRunScore[]; transaction?: Transaction }): Promise<{ id: string }[]>
```

Body:

```ts
return db
  .insert(runScores)
  .values(data)
  .onConflictDoUpdate({
    target: [runScores.runId, runScores.type, runScores.domain, runScores.name, runScores.assessmentStage],
    set: {
      value: sql`excluded.value`,
      categoryScore: sql`excluded.category_score`,
    },
  })
  .returning({ id: runScores.id });
```

Notes:

- `excluded.*` references the row Postgres tried to insert; this is how the update picks up the incoming values rather than leaving the prior ones in place.
- `updated_at` is intentionally not set in the application code — the existing `run_scores_set_updated_at` trigger handles it on UPDATE, and `created_at` has `DEFAULT now()` on the column. The general convention going forward: don't set `createdAt`/`updatedAt` from the backend; let DB defaults and triggers own them.
- `data.length === 0` short-circuits to `[]` without hitting the DB.
- Accepts an optional transaction so the upcoming `writeTrial` extension can run trial + interactions + scores atomically.

## Tests

**Unit tests** (`run-scores.repository.test.ts`) — smoke level: instantiability, method shape, empty-input no-op without a DB connection.

**Integration tests** (`run-scores.repository.integration.test.ts`) — exercise the constraint and conflict path against a real DB:

- `returns an empty array for empty input`
- `inserts a single new score row`
- `inserts multiple new score rows in one call`
- `updates the existing row when re-sent with the same natural key` — verifies the upsert path (one row before, one row after, value replaced)
- `treats two NULL assessment_stage rows with otherwise-identical natural keys as the same row` — covers the `NULLS NOT DISTINCT` semantics
- `treats different assessment_stage values as distinct rows for the same (run, type, domain, name)`
- `updates category_score on conflict`
- `isolates scores by run_id`
- `participates in a caller-provided transaction (commit path)` — proves the `transaction` param is wired through
- `rolls back upserted rows when the caller-provided transaction throws` — proves rollback semantics

## Tests support

`createMockRunScoresRepository` added under `test-support/repositories/`, exported from the index, ready for the next PR's service-level unit tests to consume.

## Out of scope

- `ScoreEntrySchema` and the `RunTrialEventSchema` extension — next PR in the chain.
- `RunEventService.writeTrial` upserting scores — next PR in the chain.
- Per-trial score history (`run_trial_scores`) — separate, future ticket.
- Recompute of `use_for_reporting` — separate, independent ticket (#1785).

## Risks

- **Hand-written snapshot.** If `drizzle-kit generate` on a reviewer's machine produces a new migration file, the hand-written snapshot doesn't match drizzle-kit's computed state. Reconcile before merge.
- **Pre-flight duplicates.** The migration will fail on a non-empty `run_scores` table that contains duplicates on the natural key. Production currently has no production writers to this table, but the comment in the SQL file includes the detection query for any environment that might.

Refs [1784](https://github.com/yeatmanlab/roar-project-management/issues/1784)

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist

### DIFF
--- a/apps/backend/migrations/assessment/0021_add_run_scores_natural_key_unique.sql
+++ b/apps/backend/migrations/assessment/0021_add_run_scores_natural_key_unique.sql
@@ -1,0 +1,23 @@
+-- Add a natural-key UNIQUE constraint on app.run_scores so we can use
+-- INSERT ... ON CONFLICT (run_id, type, domain, name, assessment_stage) DO UPDATE
+-- to upsert score rows from the writeTrial path.
+--
+-- The constraint also enforces the read-path invariant in report.repository.ts that
+-- assumes at most one current score row per natural key — drift here would silently
+-- corrupt reporting.
+--
+-- assessment_stage is nullable. Without NULLS NOT DISTINCT, two rows with identical
+-- (run_id, type, domain, name) and a NULL assessment_stage would both be allowed
+-- (Postgres treats NULL as distinct in unique constraints by default). NULLS NOT
+-- DISTINCT (Postgres 15+) makes NULLs equal for uniqueness purposes.
+--
+-- Pre-flight: this migration will fail if duplicates already exist. Run
+--   SELECT run_id, type, domain, name, assessment_stage, COUNT(*)
+--   FROM app.run_scores
+--   GROUP BY 1, 2, 3, 4, 5
+--   HAVING COUNT(*) > 1;
+-- against the target environment first and remediate before applying.
+
+ALTER TABLE "app"."run_scores"
+  ADD CONSTRAINT "run_scores_natural_key_unique"
+  UNIQUE NULLS NOT DISTINCT ("run_id", "type", "domain", "name", "assessment_stage");
--- a/apps/backend/migrations/assessment/0022_add_platform_admin_to_user_role.sql
+++ b/apps/backend/migrations/assessment/0022_add_platform_admin_to_user_role.sql
@@ -1,0 +1,11 @@
+-- Add `platform_admin` to the assessment-side `user_role` enum.
+--
+-- The shared Drizzle enum definition in `apps/backend/src/db/schema/enums.ts` already
+-- includes `platform_admin`, and the core DB has the equivalent migration
+-- (core/0062_young_tomas.sql), but the assessment-side migration was never generated.
+-- This brings the two enum copies back in sync.
+--
+-- Surfaced incidentally by `db:gen:assess` while validating the run_scores natural-key
+-- unique constraint migration (0021).
+
+ALTER TYPE "app"."user_role" ADD VALUE 'platform_admin' BEFORE 'principal';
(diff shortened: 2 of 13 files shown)


richford:
@claude, Review this PR

richford:
@claude, please re-review. `npm run db:gen:assess -w apps/backend` surfaced a missing migration for the platform admin role. The latest commit fixes that. And this also confirms that the hand-written snapshot was correct.

Emily-ejag (approved):
LGTM

kellyel (approved):

PR #1744, opened 2026-05-08
Upsert run-level scores from trial events

richford:
## Summary

This PR completes [1784](https://github.com/yeatmanlab/roar-project-management/issues/1784). It extends `RunTrialEventSchema` with an optional `scores: ScoreEntry[]` field and wires `RunEventService.writeTrial` to upsert each entry into `app.run_scores` inside the existing trial transaction. Per the ticket, this is the run-level half — per-trial score history (`run_trial_scores`) is a separate, future ticket.

Builds on the prior PR in the chain, which added the `RunScoresRepository`, the natural-key unique constraint, and the `SCORE_TYPE` / `SCORE_DOMAIN` / `SCORE_NAME` constants.

## Contract

`packages/api-contract/src/v1/runs/schema.ts`:

- New `ScoreEntrySchema` mirroring the natural-key shape of `app.run_scores`. `assessmentStage` and `categoryScore` are optional. `value` is `string` to match the backend's text-typed `value` column (covers numeric strings, support-level codes, and angle-bracket percentile strings the SDK already produces).
- `RunTrialEventSchema` gains optional `scores: ScoreEntry[]`.
- New `ScoreEntry` type export so the SDK and other consumers can build score arrays without redeclaring the schema.

The contract is shared with the future `run_trial_scores` writer — that ticket adds another consumer of `body.scores` inside `writeTrial`, no further contract change needed.

## Service

`RunEventService` gains a `runScoresRepository` constructor dependency (defaults to `new RunScoresRepository()`).

`writeTrial` does what it did before, plus one additional step at the end of the transaction:

```ts
if (body.scores && body.scores.length > 0) {
  const scoreRows: NewRunScore[] = body.scores.map((s) => ({
    runId,
    type: s.type,
    domain: s.domain,
    name: s.name,
    value: s.value,
    assessmentStage: s.assessmentStage ?? null,
    categoryScore: s.categoryScore ?? null,
  }));

  await runScoresRepository.upsertMany({
    data: scoreRows,
    transaction: tx,
  });
}
```

Notes:

- Empty/missing `scores` short-circuits before the repo call. Existing callers see zero behavior change.
- `assessmentStage` and `categoryScore` are mapped from `undefined` to `null` so the natural-key constraint sees a consistent shape; combined with `NULLS NOT DISTINCT` from the prior PR, this means a re-send with the same key replaces rather than inserts.
- The score upsert runs inside the same transaction as the trial and interactions writes. A failure in the score upsert rolls back the trial and interactions too.
[...]

PR #1745, opened 2026-05-08
Recompute use_for_reporting on trial events

richford:
## Summary

This PR resolves [issue 1785](https://github.com/yeatmanlab/roar-project-management/issues/1785). It ports the Firestore `selectBestRun` cloud function to the new Postgres backend as a service-layer recompute, called as the last step of `RunEventService.writeTrial`. After every trial event, exactly one non-aborted, non-deleted run per `(user_id, administration_id, task_variant_id)` partition is `use_for_reporting = true`, or zero if no eligible runs exist. The read path in `report.repository.ts` already assumed this invariant — this PR makes it something the backend actually enforces rather than something we hope holds.

Builds on #1744, which delivered `RunScoresRepository`, the natural-key upsert, the `SCORE_TYPE`/`SCORE_DOMAIN`/`SCORE_NAME` constants, and the `writeTrial` score-upsert wiring this PR's recompute reads from.

## Selection priority (unchanged from Firestore)

1. Reliable + completed — earliest `created_at`
2. Reliable + incomplete — lowest `thetaSE`, then highest `numAttempted`
3. Unreliable + completed — latest `created_at`
4. Unreliable + incomplete — lowest `thetaSE`, then highest `numAttempted`, then earliest `created_at`

The new schema has no dedicated start timestamp, so `created_at` substitutes for the legacy `timeStarted`. Behaviorally equivalent (runs are created at start), called out in the JSDoc.

## Repository

`RunRepository.recomputeUseForReporting({ userId, administrationId, taskVariantId, transaction? })` — a single SQL statement structured as three CTEs plus an outer UPDATE:

```sql
WITH locked_candidates AS (
  SELECT id, reliable_run, completed_at, created_at FROM app.runs
  WHERE <partition keys> AND deleted_at IS NULL AND aborted_at IS NULL
  FOR UPDATE
),
ranked AS (
  SELECT lc.id, ROW_NUMBER() OVER (ORDER BY <four-tier CASE>) AS rn
  FROM locked_candidates lc
  LEFT JOIN app.run_scores theta_se ON ...
  LEFT JOIN app.run_scores num_attempted ON ...
),
winner AS (
  SELECT id FROM ranked WHERE rn = 1
)
UPDATE app.runs r
SET use_for_reporting = ((r.id = (SELECT id FROM winner)) IS TRUE)
WHERE <partition keys> AND deleted_at IS NULL
```

Three things worth flagging because the original ticket SQL had to be revised:

1. **`FOR UPDATE` is split off into its own CTE.** Postgres disallows `FOR UPDATE` in a CTE that uses window functions; `locked_candidates` acquires the locks and `ranked` reads them. Concurrent recomputes for the same partition serialize at the row-lock level.
2. **Scalar subquery on `winner` instead of `UPDATE … FROM (winner)`.** `UPDATE … FROM` is an inner join: if `winner` is empty (every run aborted), the UPDATE would affect zero rows and previously-true `use_for_reporting` flags would silently persist. The scalar subquery returns `NULL` for an empty winner, and `(r.id = NULL) IS TRUE` evaluates to `false`, which is exactly the "no eligible runs → all false" semantics the ticket asked for.
3. **Outer WHERE deliberately omits `aborted_at IS NULL`.** Aborted (and now-ineligible) runs in the partition need to be reset from any stale `use_for_reporting=true` to `false`, which means they must be in the UPDATE's scope.

`thetaSE` and `numAttempted` lookups are `LEFT JOIN`s with `NULLS LAST` ordering, so runs without those scores still rank deterministically within their tier (this matters most in tier 4 where the final tiebreaker is `created_at`).

The tier priorities are named in `apps/backend/src/constants/best-run.ts` (`BEST_RUN_TIER.RELIABLE_COMPLETED` etc.) and interpolated into the SQL — no bare `1, 2, 3, 4` literals. CAT field names (`thetaSE`, `numAttempted`, the `raw`/`composite`/`test` triple) come from `constants/run-scores.ts` from #1744 .

## Architectural note — business logic in the repository

The four-tier ranking is business logic (tier priority and in-tier tiebreakers can change as product requirements evolve), and it lives in the repository as SQL. That's an intentional deviation from the usual "business logic belongs in services" principle, by deliberate analogy with `repositories/access-controls/`: both are cases where the rules can only be applied efficiently as joins/ordering against many candidate rows in a single round-trip.

Doing the ranking in the service layer would require either (a) fetching every candidate run plus its scores into TypeScript and sorting in JS — wasteful and reintroducing TOCTOU races against concurrent writers, or (b) the service composing SQL fragments and handing them to the repository, which doesn't actually move the logic anywhere meaningful.

The decision is documented in the JSDoc on `recomputeUseForReporting` so future readers can see the trade-off was deliberate. The off-ramp, when a second consumer of the ranking ever appears (e.g., a dry-run preview endpoint), is to extract the ORDER BY expression into an injectable helper along the lines of `AdministrationAccessControls`.

## Service

`RunService.recomputeBestRunForVariant({ userId, administrationId, taskVariantId, transaction? })`:

- Short-circuits if `administrationId === ANONYMOUS_RUN_ADMINISTRATION_ID` — anonymous-run partitions cross users and the concept of "best run" is meaningless there.
- Delegates to the repository, passing through any caller-provided transaction.
- Catches DB errors and wraps them in `ApiError(INTERNAL_SERVER_ERROR, DATABASE_QUERY_FAILED)` with structured logging context. `ApiError` instances bubble up unchanged.

## Wiring

`RunEventService.writeTrial`:

```ts
const run = await assertRunOwnedByUser(runId, targetUserId);
// ... transaction with trial + interactions + scores ...
await runService.recomputeBestRunForVariant({
  userId: run.userId,
  administrationId: run.administrationId,
  taskVariantId: run.taskVariantId,
  transaction: tx,
});
```

The recompute is the **last step** of the existing transaction, so it sees the trial's freshly-written `run_scores` rows. A failure in any earlier step (trial insert, interactions, score upsert) skips the recompute via the rejection path; a failure in the recompute itself rolls back everything.

`RunService` is a new injected dependency on `RunEventService`, mirroring how `AuthorizationService` is already injected. Default: `RunService()`.

## Tests

**Service unit tests** (`run.service.test.ts`) — five new cases under `describe('recomputeBestRunForVariant')`:

- `delegates to runRepository.recomputeUseForReporting with the partition keys`
- `passes the caller-provided transaction through to the repository`
- `short-circuits without hitting the repository for anonymous-run partitions`
- `wraps a repository failure in a 500 ApiError with DATABASE_QUERY_FAILED`
- `re-throws ApiError unchanged without re-wrapping`

**Wiring unit tests** (`run-event.service.test.ts`) — two new cases under `describe('scores')`:

- `calls runService.recomputeBestRunForVariant inside the trial transaction` — verifies partition keys and the transaction sentinel pass through
- `runs the recompute after the score upsert (last step of the transaction)` — verifies call order is `['trial', 'scores', 'recompute']`

**Repository integration tests** (`run.repository.integration.test.ts`) — twelve new cases under `describe('recomputeUseForReporting')`:

- Single-run partition → that run wins
- Tier 1: reliable + completed, earliest `created_at` wins
- Tier 2: reliable + incomplete, lowest `thetaSE` wins
- Tier 2 ties on `thetaSE`: highest `numAttempted` wins
- Tier 3: unreliable + completed, latest `created_at` wins
- Tier 4: unreliable + incomplete, lowest `thetaSE` wins
- Cross-tier priority: tier 1 beats tier 2 even when tier 2 has a better `thetaSE`
- Reliable beats unreliable across completion states (tier 2 beats tier 3)
- Aborted runs excluded from candidates AND reset from any stale `use_for_reporting=true`
- Soft-deleted runs are entirely outside the partition
- No eligible candidates → every run in the partition resets to `false`
- **Score-flip case** — updating an existing `thetaSE` to a lower value changes the winner on the next recompute (the AC's specifically-called-out case)
- Partition isolation — recompute on partition A leaves partition B untouched
- Idempotency — running the recompute twice produces identical state

## Test support

- New `createMockRunService` factory at `test-support/services/run.service.ts`, exported from the index.
- Existing `createMockRunRepository` extended with `recomputeUseForReporting: vi.fn()`.
- The `runs.controller.test.ts` mock for `RunService` was updated to include the new method (one-line fix flagged by `tsc`).

## Out of scope

- A run-delete path — when one lands, it must call recompute too. Add as part of whichever ticket introduces deletion.
- Postgres trigger as a defense-in-depth backstop — defer until drift is observed.
- Backfill of historical rows that don't satisfy the invariant.
- Wiring recompute into `completeRun` / `abortRun` / `updateEngagement` — the AC specifically narrowed scope to `writeTrial` only.

## Risks

- **Service depends on every future write path remembering to call it.** Discoverable via the `RunService.recomputeBestRunForVariant` reference, but this is the kind of invariant that gets quietly broken. Consider a periodic invariant-check integration test that scans for partitions with more than one `use_for_reporting=true` row.
- **The SQL is the riskiest piece.** The integration suite covers all four tiers, the cross-tier priorities, and the score-flip case. Any new score names beyond `thetaSE` / `numAttempted` that affect ranking would require updates to the four-tier expression.
- **`completeRun`, `abortRun`, `updateEngagement` also change ranking-relevant state** (`completed_at`, `aborted_at`, `reliable_run`) but currently don't trigger a recompute per the AC's narrowed scope. If that becomes a pain point, add the calls in a follow-up — none of them carry score writes, so the recompute call there would be the only addition.

## Reference

- Firestore implementation: `roar-firebase-functions/packages/core/src/services/run.service.ts` (`selectBestRun`, `updateBestRunAndAssignmentCompletion`)
- Best-run finder: `roar-firebase-functions/packages/core/src/utils/best-run-finder.util.ts`
- Run comparison: `roar-firebase-functions/packages/core/src/utils/run-comparison.util.ts`
- Read-path invariant assumption: `apps/backend/src/repositories/report.repository.ts:1395`

Resolves [1785](https://github.com/yeatmanlab/roar-project-management/issues/1785)

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist

### DIFF
--- a/apps/backend/migrations/assessment/0021_add_run_scores_natural_key_unique.sql
+++ b/apps/backend/migrations/assessment/0021_add_run_scores_natural_key_unique.sql
@@ -1,0 +1,23 @@
+-- Add a natural-key UNIQUE constraint on app.run_scores so we can use
+-- INSERT ... ON CONFLICT (run_id, type, domain, name, assessment_stage) DO UPDATE
+-- to upsert score rows from the writeTrial path.
+--
+-- The constraint also enforces the read-path invariant in report.repository.ts that
+-- assumes at most one current score row per natural key — drift here would silently
+-- corrupt reporting.
+--
+-- assessment_stage is nullable. Without NULLS NOT DISTINCT, two rows with identical
+-- (run_id, type, domain, name) and a NULL assessment_stage would both be allowed
+-- (Postgres treats NULL as distinct in unique constraints by default). NULLS NOT
+-- DISTINCT (Postgres 15+) makes NULLs equal for uniqueness purposes.
+--
+-- Pre-flight: this migration will fail if duplicates already exist. Run
+--   SELECT run_id, type, domain, name, assessment_stage, COUNT(*)
+--   FROM app.run_scores
+--   GROUP BY 1, 2, 3, 4, 5
+--   HAVING COUNT(*) > 1;
+-- against the target environment first and remediate before applying.
+
+ALTER TABLE "app"."run_scores"
+  ADD CONSTRAINT "run_scores_natural_key_unique"
+  UNIQUE NULLS NOT DISTINCT ("run_id", "type", "domain", "name", "assessment_stage");
--- a/apps/backend/migrations/assessment/0022_add_platform_admin_to_user_role.sql
+++ b/apps/backend/migrations/assessment/0022_add_platform_admin_to_user_role.sql
@@ -1,0 +1,11 @@
+-- Add `platform_admin` to the assessment-side `user_role` enum.
+--
+-- The shared Drizzle enum definition in `apps/backend/src/db/schema/enums.ts` already
+-- includes `platform_admin`, and the core DB has the equivalent migration
+-- (core/0062_young_tomas.sql), but the assessment-side migration was never generated.
+-- This brings the two enum copies back in sync.
+--
+-- Surfaced incidentally by `db:gen:assess` while validating the run_scores natural-key
+-- unique constraint migration (0021).
+
+ALTER TYPE "app"."user_role" ADD VALUE 'platform_admin' BEFORE 'principal';
--- a/apps/backend/migrations/assessment/0023_add_run_scores_raw_requires_stage_check.sql
+++ b/apps/backend/migrations/assessment/0023_add_run_scores_raw_requires_stage_check.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "app"."run_scores" ADD CONSTRAINT "run_scores_raw_requires_stage" CHECK ("app"."run_scores"."type" <> 'raw'::app.score_type OR "app"."run_scores"."assessment_stage" IS NOT NULL);
--- a/apps/backend/migrations/assessment/meta/0021_snapshot.json
+++ b/apps/backend/migrations/assessment/meta/0021_snapshot.json
@@ -1,0 +1,895 @@
+{
+  "id": "a1314e7c-0697-4a39-9023-3becff80bdcb",
+  "prevId": "b2c3d4e5-f6a7-4b8c-9d0e-1f2a3b4c5d6e",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "app.runs": {
+      "name": "runs",
+      "schema": "app",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "uuid",
+          "primaryKey": true,
+          "notNull": true,
+          "default": "gen_random_uuid()"
+        },
+        "user_id": {
+          "name": "user_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "task_id": {
+          "name": "task_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "task_variant_id": {
+          "name": "task_variant_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "task_version": {
+          "name": "task_version",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "administration_id": {
+          "name": "administration_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "use_for_reporting": {
+          "name": "use_for_reporting",
+          "type": "boolean",
+          "primaryKey": false,
+          "notNull": true,
+          "default": false
+        },
+        "reliable_run": {
+          "name": "reliable_run",
+          "type": "boolean",
+          "primaryKey": false,
+          "notNull": true,
+          "default": false
+        },
+        "engagement_flags": {
+          "name": "engagement_flags",
+          "type": "jsonb",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "metadata": {
+          "name": "metadata",
+          "type": "jsonb",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "is_anonymous": {
+          "name": "is_anonymous",
+          "type": "boolean",
+          "primaryKey": false,
+          "notNull": true,
+          "default": false
+        },
+        "completed_at": {
+          "name": "completed_at",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "aborted_at": {
+          "name": "aborted_at",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "deleted_at": {
+          "name": "deleted_at",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "deleted_by": {
+          "name": "deleted_by",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "updated_at": {
+          "name": "updated_at",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "created_at": {
+          "name": "created_at",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "now()"
+        }
+      },
+      "indexes": {
+        "runs_user_administration_task_idx": {
+          "name": "runs_user_administration_task_idx",
+          "columns": [
+            {
+              "expression": "user_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            },
+            {
+              "expression": "administration_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            },
+            {
+              "expression": "task_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        },
+        "runs_user_id_idx": {
+          "name": "runs_user_id_idx",
+          "columns": [
+            {
+              "expression": "user_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        },
+        "runs_user_reporting_run_idx": {
+          "name": "runs_user_reporting_run_idx",
+          "columns": [
+            {
+              "expression": "user_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "where": "\"app\".\"runs\".\"use_for_reporting\" = true",
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        }
+      },
+      "foreignKeys": {},
+      "compositePrimaryKeys": {},
+      "uniqueConstraints": {},
+      "policies": {},
+      "checkConstraints": {
+        "runs_deleted_by_required": {
+          "name": "runs_deleted_by_required",
+          "value": "\"app\".\"runs\".\"deleted_at\" IS NULL OR \"app\".\"runs\".\"deleted_by\" IS NOT NULL"
+        },
+        "runs_anonymous_administration_id": {
+          "name": "runs_anonymous_administration_id",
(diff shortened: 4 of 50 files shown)


richford:
@claude, please review this PR

richford:
@claude please re-review. For reference, here is the originating ticket:

## Summary

Port the Firestore `selectBestRun` cloud function to a service-layer recompute in the new Postgres backend. When a run's state changes — trial write, completion, abort, engagement/reliability flip, or future delete — recompute which run in the partition `(user_id, administration_id, task_variant_id)` should carry `use_for_reporting = true`. The read path (`report.repository.ts`) already assumes at most one `use_for_reporting=true`, non-aborted, non-deleted run per partition; this ticket makes that invariant something the backend actually enforces rather than something we hope holds.

This ticket is independent. It does not depend on any score-writing ticket: the recompute reads whatever's currently in `run_scores` (empty or otherwise) and ranks accordingly. If `run_scores` is empty when this lands, runs will rank by tier and by the non-score tiebreakers — which is correct behavior.

## Background

The Firestore implementation in `roar-firebase-functions/packages/core/src/services/run.service.ts` does two things on every run document write:

1. Pick the "best" run for `(roarUid, assignmentId, taskId)` and stamp `bestRun: true/false` on every run in that group.
2. Fan out to the assignment document — set `runId`, `allRunIds`, `allowRetake`, `startedOn`, `completedOn`, and recompute assignment-level progress.

Only piece 1 is in scope here. The new backend has no `assignments` table; assignment-level state is derived at query time in `report.repository.ts`. The grain also shifts from `taskId` to `taskVariantId` — a student who runs both an English and a Spanish variant of the same task in one administration ends up with two `use_for_reporting=true` rows, one per variant.

Selection priority (unchanged from Firestore):

1. Reliable + completed — earliest started
2. Reliable + incomplete — lowest `thetaSE`, then highest `numAttempted`
3. Unreliable + completed — latest started
4. Unreliable + incomplete — lowest `thetaSE`, then highest `numAttempted`, then earliest started

## Approach

Add `RunService.recomputeBestRunForVariant({ userId, administrationId, taskVariantId, transaction? })` and call it from every `RunEventService` write that can change the partition's ranking. The recompute itself is a single SQL `UPDATE` driven by a window function so it stays atomic.

```mermaid
sequenceDiagram
    Client->>RunEventService: writeTrial / completeRun / abortRun / updateEngagement
    RunEventService->>RunRepository: update run / trial / score state (in tx)
    RunEventService->>RunService: recomputeBestRunForVariant (same tx)
    RunService->>Repository: SELECT FOR UPDATE candidates + scores
    RunService->>Repository: UPDATE runs SET use_for_reporting = (id = winner)
```

The recompute is the **last step** of the transaction so it sees the freshly-written scores and trial/run state.

Why service-layer rather than a Postgres trigger:

- Keeps the four-tier ranking in TypeScript where it can be unit-tested with mocked repositories.
- Avoids the trigger-recursion problem the Firestore version solves with the `cloudSyncTimestamp` heuristic — the function isn't triggered by the write it makes.
- Aligns with the rule "business logic belongs in services" (`backend-layer-architecture`, `backend-controller-no-business-logic`).

A Postgres trigger can be added later as a defense-in-depth measure if drift is observed, without changing the public API.

## Acceptance criteria

- [ ] `RunService.recomputeBestRunForVariant` exists and is called from `RunEventService.writeTrial`.
- [ ] In `writeTrial`, the recompute is the **last step** of the existing transaction so it sees any score writes performed earlier in that same transaction (this matters once the score-write ticket lands; harmless before then).
- [ ] After every call, exactly one non-aborted, non-deleted run per `(user_id, administration_id, task_variant_id)` has `use_for_reporting = true`, or zero if no eligible runs exist.
- [ ] Aborted, deleted, and anonymous runs (`administration_id = ANONYMOUS_RUN_ADMINISTRATION_ID`) are never selected.
- [ ] Anonymous runs short-circuit before the SQL — no recompute attempted.
- [ ] Recompute runs in the same transaction as the triggering state change so a failure rolls both back together.
- [ ] Concurrent recomputes for the same partition are serialized via `SELECT ... FOR UPDATE` on the candidate set.
- [ ] Unit tests cover all four tier branches plus ties, mirroring the cases in `roar-firebase-functions/packages/core/src/services/run.service.test.ts`.
- [ ] Integration test verifies the invariant holds after sequences that mix `writeTrial`, `completeRun`, `abortRun`, and `updateEngagement` against a real database. Use the `run_scores` test factory to seed score rows directly so this test does not depend on the score-write ticket. Specifically cover: a partition where seeding a lower `thetaSE` for one run flips the winner under tier 2/4 ranking.
- [ ] Errors wrap the underlying DB error in `ApiError` with `ApiErrorCode.DATABASE_QUERY_FAILED` and structured logging context (`userId`, `administrationId`, `taskVariantId`).

## Implementation notes

**Score lookup.** thetaSE and numAttempted live in `app.run_scores`, looked up by `(type='raw', domain='composite', assessment_stage='test', name IN ('thetaSE', 'numAttempted'))`. Introduce `constants/run-scores.ts` (or similar) with `SCORE_NAME.THETA_SE` / `SCORE_NAME.NUM_ATTEMPTED` and related fields as part of this PR — never inline magic strings. The score-write ticket can reuse the same constants if it lands later.

**Started timestamp.** New `runs` has no `time_started` column — use `created_at` for the tier 1, 3, and 4 tiebreakers. Behaviorally equivalent since runs are created at start, but call this out in JSDoc.

**SQL shape.** A single statement of the form:

```sql
UPDATE app.runs r
SET use_for_reporting = (r.id = winner.id)
FROM (
  SELECT id FROM (
    SELECT
      r.id,
      ROW_NUMBER() OVER (ORDER BY /* four-tier CASE expression */) AS rn
    FROM app.runs r
    LEFT JOIN app.run_scores theta_se ON theta_se.run_id = r.id AND theta_se.name = 'thetaSE' /* ... */
    LEFT JOIN app.run_scores num_attempted ON num_attempted.run_id = r.id AND num_attempted.name = 'numAttempted' /* ... */
    WHERE r.user_id = $1
      AND r.administration_id = $2
      AND r.task_variant_id = $3
      AND r.deleted_at IS NULL
      AND r.aborted_at IS NULL
    FOR UPDATE OF r
  ) ranked
  WHERE rn = 1
) winner
WHERE r.user_id = $1
  AND r.administration_id = $2
  AND r.task_variant_id = $3
  AND r.deleted_at IS NULL
```

The outer `WHERE` deliberately omits `aborted_at IS NULL` so that previously-true `use_for_reporting` on aborted or newly-ineligible runs gets reset to `false` in the same statement. If the inner subquery returns no rows, every run in the partition ends up `false`, which is the correct behavior.

**Idempotency.** Running the recompute twice with no intervening writes must produce the same result.

**Future call sites.**

- A run-delete path (when it lands) must call recompute — a deleted run could have been the winner. Add the call site as part of whichever ticket introduces deletion.

## Out of scope

- Assignment-level status (`started`, `completed`, `progress`) — derived at query time, no `assignments` table to write to.
- Score writing itself — separate, independent ticket.
- Postgres trigger — defer until drift is observed in practice.
- Backfill of existing rows that don't satisfy the invariant — separate ticket if needed.

## Risks

- The service depends on every future write path remembering to call it. Discoverable via the `RunService.recomputeBestRunForVariant` reference, but this is the kind of invariant that gets quietly broken. Consider a periodic invariant-check integration test that scans for partitions with more than one `use_for_reporting=true` run.
- thetaSE and numAttempted scores will be absent for some runs (incomplete and no scores yet, or completed with only computed scores). The SQL must rank such runs deterministically within their tier and not throw — `LEFT JOIN` plus `NULLS LAST` handles this, but it deserves an explicit test.

## Reference

- Firestore implementation: `roar-firebase-functions/packages/core/src/services/run.service.ts` (`selectBestRun`, `updateBestRunAndAssignmentCompletion`)
- Best-run finder: `roar-firebase-functions/packages/core/src/utils/best-run-finder.util.ts`
- Run comparison: `roar-firebase-functions/packages/core/src/utils/run-comparison.util.ts`
- Firestore trigger: `roar-firebase-functions/packages/roar-assessment/src/controllers/sync.controller.ts`
- New backend runs schema: `apps/backend/src/db/schema/assessment/runs.ts`
- New backend run-scores schema: `apps/backend/src/db/schema/assessment/run-scores.ts`
- Existing service entry point: `apps/backend/src/services/run-event/run-event.service.ts`
- Read-path invariant assumption: `apps/backend/src/repositories/report.repository.ts:1395`
- Related but independent: *Write run-level scores on `trial` and `complete` run events* (introduces real score data into `run_scores`; recompute behavior strengthens once that ticket lands but does not require it)

Emily-ejag (dismissed):
LGTM

kellyel (dismissed):

richford:
@claude, please re-review.

richford:
@claude, I resolved merge conflicts. Please re-review and confirm merge correctness and that I introduced no regression.

kellyel (approved):

Emily-ejag (approved):

PR #1774, opened 2026-05-11
Wire GET /me into app init with TOS gating and TanStack Query ownership

richford:
## Summary

This PR wires the backend `GET /me` endpoint into dashboard app initialization as the single source of truth for the authenticated user's identity, type, and unsigned-TOS status, replacing the Firestore-based user fetch. It handles every `/me` outcome: unsigned TOS blocks navigation behind a signing flow, `auth/rostering-ended` routes to the AccessEnded page, and 500/network errors route to GenericError with retry. `/me` data lives in TanStack Query (not Pinia), and `<router-view>` is gated on `/me` settlement so no protected content paints before a redirect can fire.

This is the first endpoint of the frontend→backend migration and supersedes the monolithic PR that was previously reviewed on this PR. It is resliced into small, reviewable commits and stacks on the seeded-backend e2e stack (PR 1, `enh/e2e-tooling`).

## What changed

**Composables**
- `useMeQuery` — calls `GET /me`, internally gated on `accessToken`; no retry on `auth/rostering-ended` or terminal auth, and the retry policy can't be overridden by callers.
- `useCurrentUser` — derives identity from the `/me` query (replaces the auth-store mirror).
- `useAgreementVersionContentQuery` and `useRecordUserAgreementMutation` (treats a 409 as already-signed and advances the queue).

**TOS signing flow** (container / presentational)
- `SignTos` page + container + `useTosSigningFlow`; renders each agreement in the existing `ConsentModal`; the `?next=` return path is validated with the WHATWG URL parser plus a strict internal-path check.

**App init & routing**
- TanStack Query owns `/me`; the auth-store `meData` mirror is removed; `<router-view>` is gated on `/me` settlement.
- The router `beforeEach` guard is the single owner of TOS routing; `meError`→`globalError` mapping is consolidated in the QueryCache `onError`.
- Route names and global-error types use named constants; `QueryClient` is a module-scoped singleton.

## `/me` lifecycle

```mermaid
sequenceDiagram
  participant FB as Firebase Auth
  participant MQ as useMeQuery
  participant App as App.vue (router-view gated)
  participant R as Router guard
  FB->>MQ: access token
  MQ->>App: GET /me settles
  alt 200 OK
    alt unsigned TOS
      R-->>App: redirect to SignTos
    else
      R-->>App: render destination
    end
  else auth/rostering-ended
    App-->>App: AccessEnded
  else 500 / network
    App-->>App: GenericError (retry)
  end
```

## Review feedback addressed (from previous review rounds on this PR)

- `/me` state owned by TanStack Query, not mirrored in Pinia.
- A single TOS-enforcement mechanism: render gated on `/me` settlement, router guard owns routing (no parallel watcher).
- `useMeQuery` retry policy is caller-proof; `http-status-codes` constants instead of literals; route-name and error-type constants; stricter `?next=` validation; Cypress exercises the real seeded backend (no `/me` stub).

## Test plan

- **Unit:** `useMeQuery`, `useCurrentUser`, the agreement query/mutation, `pickVersionForLocale`, `isInternalPath`, `queryClient`.
- **E2E:** the `/me` smoke spec against the seeded backend (PR 1).
- **Manual:** sign-in → unsigned TOS → sign all → proceeds to destination; rostering-ended → AccessEnded; 500 → GenericError retry.

Resolves [1738](https://github.com/yeatmanlab/roar-project-management/issues/1738)

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist


richford:
@claude, please review this PR. Here is the originating issue for reference:

## Summary

Introduce the `GET /me` backend API call into the frontend app initialization flow, replacing the Firestore-based user data fetch. Handle all error states from `/me` gracefully: rostering-ended users see a dedicated "access ended" screen, users with unsigned TOS are blocked until they sign, and server errors show a generic error page with retry.

## Context

The dashboard currently loads user identity and claims entirely from Firebase Authentication and Firestore. As part of the migration to the new backend API, the `/me` endpoint becomes the single source of truth for user identity, type, and TOS status.

This ticket wires up the `/me` call and handles three critical states:
1. **Rostering ended** (`code: auth/rostering-ended`) -- redirect to the AccessEnded page
2. **Unsigned TOS** (`unsignedAgreements` array is non-empty) -- block navigation and present the TOS signing flow
3. **Server error** (500) -- show the GenericError page with retry

### Prerequisites

- Backend: `GET /me` returns `unsignedAgreements` array (separate ticket)
- Backend: Auth guard returns `auth/rostering-ended` error code (separate ticket)
- Backend: `GET /agreements/:agreementId/versions/:versionId` returns agreement content (separate ticket)
- Backend: `POST /users/:userId/agreements` records user agreement consent (already implemented)
- Frontend: Global error handling infrastructure (separate ticket)

## Implementation Details

### 1. `/me` Query Composable

Create a new TanStack Query composable for the `/me` endpoint:

```javascript
// composables/queries/useMeQuery.js
const useMeQuery = (queryOptions = undefined) => {
  const { isQueryEnabled, options } = computeQueryOverrides([], queryOptions);

  return useQuery({
    queryKey: [ME_QUERY_KEY],
    queryFn: () => apiClient.me.get(),
    enabled: isQueryEnabled,
    retry: (failureCount, error) => {
      // Don't retry on auth/rostering-ended errors
      if (isRosteringEndedError(error) || isAuthError(error)) return false;
      return failureCount < 3;
    },
    ...options,
  });
};
```

Add `ME_QUERY_KEY` to `src/constants/queryKeys.js`.

### 2. App Initialization Flow

Replace the Firestore-based user data fetch in the app initialization with the `/me` query. The flow becomes:

```
Firebase Auth sign-in (existing)
  → Token obtained
  → Call GET /me with Bearer token
  → Success:
      → Store user identity in Pinia (id, userType, nameFirst, nameLast)
      → Check unsignedAgreements:
          → Non-empty: Block navigation, show TOS signing flow
          → Empty: Proceed to requested route
  → Error (auth/rostering-ended):
      → Set global error state
      → Router redirects to AccessEnded page
  → Error (401/auth errors):
      → Redirect to sign-in page
  → Error (500):
      → Set global error state
      → Router redirects to GenericError page
```

### 3. TOS Signing Gate

When `/me` returns a non-empty `unsignedAgreements` array, the user must sign before proceeding:

**Router guard:** Add a check in `router.beforeEach` that blocks all navigation (except to the TOS signing route and sign-out) when unsigned agreements exist:

```javascript
router.beforeEach((to) => {
  if (authStore.hasUnsignedTos && to.name !== 'SignTos' && to.name !== 'SignOut') {
    return { name: 'SignTos' };
  }
});
```

**TOS signing page/route:** Create a dedicated route that:
1. Displays the list of unsigned agreements
2. For each agreement, fetches the content via `GET /agreements/:agreementId/versions/:versionId` (using the version matching the user's locale, falling back to `en-US`)
3. Renders the content in the existing `ConsentModal` component
4. On acceptance, calls `POST /users/:userId/agreements` with the selected `agreementVersionId`
5. After all agreements are signed, invalidates the `/me` query and navigates to the originally requested route

**Locale selection:** The frontend picks the version from `unsignedAgreements[].versions` that matches the user's current locale. If no match, fall back to `en-US`. Since all locale variants are included in the `/me` response, no extra round-trip is needed if the user switches locale.

### 4. Auth Store Updates

Update the Pinia auth store to hold the `/me` response data:

```javascript
// In auth store or a new dedicated store
state: () => ({
  // ... existing state
  meData: null, // { id, userType, nameFirst, nameLast, unsignedAgreements }
}),

getters: {
  hasUnsignedTos: (state) => state.meData?.unsignedAgreements?.length > 0,
  currentUserId: (state) => state.meData?.id,
},
```

### 5. Error Handling Integration

Use the global error handling infrastructure (prerequisite ticket) to route errors:

```javascript
// In the /me query's onError or in a watcher
watch(meQuery.error, (error) => {
  if (!error) return;
  const appState = useAppStateStore();

  if (isRosteringEndedError(error)) {
    appState.setGlobalError({ type: 'rostering-ended' });
  } else if (isAuthError(error)) {
    // Redirect to sign-in
    router.push({ name: 'SignIn' });
  } else {
    appState.setGlobalError({ type: 'server-error' });
  }
});
```

## Directory Structure

```
src/
├── composables/
│   └── queries/
│       └── useMeQuery.js              # New: /me query composable
├── containers/
│   └── SignTos/
│       ├── SignTos.vue                 # New: TOS signing container
│       └── composables/
│           └── useTosSigningFlow.js    # New: TOS signing orchestration
├── pages/
│   └── SignTos.vue                     # New: TOS signing page (thin wrapper)
├── constants/
│   └── queryKeys.js                   # Updated: add ME_QUERY_KEY
└── store/
    └── auth.js                        # Updated: store /me response data
```

## Acceptance Criteria

### /me Integration
- [ ] `GET /me` is called during app initialization after Firebase auth
- [ ] User identity fields from `/me` are stored in Pinia
- [ ] The Firestore user data fetch is replaced by the `/me` call
- [ ] `useMeQuery` composable follows existing query composable patterns (`computeQueryOverrides`, centralized query key)

### Rostering Ended
- [ ] When `/me` returns `auth/rostering-ended`, the user sees the AccessEnded page
- [ ] The user cannot navigate away from the AccessEnded page (except sign-out)
- [ ] The AccessEnded page has a sign-out button that works

### TOS Signing Gate
- [ ] When `/me` returns non-empty `unsignedAgreements`, the user is redirected to the TOS signing flow
- [ ] The user cannot navigate to any other page until TOS is signed
- [ ] The TOS content is fetched via `GET /agreements/:agreementId/versions/:versionId`
- [ ] The correct locale variant is displayed based on the user's current locale
- [ ] If no locale match, falls back to `en-US`
- [ ] The existing `ConsentModal` component is used to display the agreement content
- [ ] On acceptance, `POST /users/:userId/agreements` is called with the correct `agreementVersionId`
- [ ] After signing, the `/me` query is invalidated and the user proceeds to the originally requested route
- [ ] If multiple TOS agreements are unsigned, the user must sign all of them

### Server Errors
- [ ] When `/me` returns 500, the GenericError page is displayed
- [ ] The GenericError page has a retry button
- [ ] When `/me` returns 401, the user is redirected to sign-in

### General
- [ ] No regressions in existing authentication flows
- [ ] Follows container/presentational pattern for the TOS signing flow
- [ ] Query composable follows `computeQueryOverrides` pattern
- [ ] Query key is centralized in `queryKeys.js`
- [ ] TanStack Query does not retry on auth/rostering-ended errors

richford:
@claude, please re-review

richford:
@claude, please re-review

richford:
@claude please re-review

.github/actions/build/action.yml
```diff
@@ -69,6 +69,16 @@ runs:
...
+    # from the dashboard's working directory bypasses turbo, so transitive
+    # workspace builds aren't triggered automatically.)
+    - name: Build api-contract package
+      working-directory: build
+      shell: bash
+      run: npm run build -w packages/api-contract
+
     # Build the application
     - name: Build application
       working-directory: build/apps/dashboard
```
maximilianoertel: This works, but could be replaced with a `npx turbo build --filter=roar-dashboard` from the root of the repo. This would future proof the build as Turbo resolves local build dependencies via the `dependsOn` in `turbo.json`. The same could be applied to the changes made in `.github/workflows/ci.yml`

apps/dashboard/cypress/support/e2e.js
```diff
@@ -15,6 +15,31 @@ beforeEach(() => {
...
+      data: {
+        id: '00000000-0000-0000-0000-000000000000',
+        userType: 'admin',
+        nameFirst: 'E2E',
+        nameLast: 'Test',
+        unsignedAgreements: [],
+      },
+    },
+  }).as('getMe');
+
```
maximilianoertel: I understand why you added this, but I strongly recommend we remove this as it's quite counterproductive for the purpose of our E2E tests. Instead of intercepting these requests, we should:

- Create a new set of E2E test (and QA) dedicated test accounts 
- Use `drizzle-seed` to seed these test accounts against into every test environment to ensure these accounts exist with the correct org memberships, roles, and signed TOS state before any Cypress suite executes
- Let the real `/me` endpoint respond with real data so E2E tests exercise the actual app initialisation flow, including the TOS guard, the auth store hydration, and the router guards, rather than silently bypassing all of it with a hardcoded stub

```suggestion
```

apps/dashboard/src/composables/mutations/useSignOutMutation.js
```diff
@@ -24,8 +25,17 @@ const useSignOutMutation = () => {
       // Cancel all actively fetching queries.
       await queryClient.cancelQueries();
 
+      // Drop any cached `/me` payload before the store reset, so a fast
+      // re-sign-in (different user) doesn't transiently see the previous
+      // user's data through the watcher in `App.vue` while the new `/me`
+      // request is in flight.
+      queryClient.removeQueries({ queryKey: [ME_QUERY_KEY] });
```
maximilianoertel: Did you add this because you noticed odd behaviour during manual testing? In theory, this in redundant as we already cancel in-flight queries through `cancelQueries()` and wipe the entire cache with `queryClient.clear()` 

I know, sometimes we prefer to play it safe, but I'd rather avoid adding redundant cleanup steps that suggest the primary mechanisms aren't trustworthy. If there's a real race condition where stale `/me` data leaks through between `cancelQueries()` and `clear()`, we should fix the underlying issue. 

In other words, I recommend we remove it unless there's a specific reproduction case that motivated it.


```suggestion
```

apps/dashboard/src/composables/mutations/useSignOutMutation.js
```diff
@@ -24,8 +25,17 @@ const useSignOutMutation = () => {
...
+
       // Reset store and delete persisted data. Persisted data should be cleared via the $reset but to be safe, we also
       // remove it manually from sessionStorage to prevent any issues.
+      // `$reset` already nulls `meData` since it's part of state, but the explicit `clearMeData()` is kept as
+      // belt-and-suspenders against any future state-shape changes that might cause `$reset()` to miss it.
+      authStore.clearMeData();
```
maximilianoertel: Same as for https://github.com/yeatmanlab/roar-dashboard/pull/1774/changes#r3266662812 – `$reset` is a built-in method designed to void the state, a custom `clearMeData` should isn't needed here? 

```suggestion
```

apps/dashboard/src/composables/queries/useAgreementVersionContentQuery.js
```diff
@@ -0,0 +1,52 @@
...
+    queryFn: async () => {
+      const client = getRoarApiClient();
+      const result = await client.agreements.getVersionContent({
+        params: {
+          agreementId: toValue(agreementId),
+          versionId: toValue(versionId),
+        },
+      });
+
+      if (result.status === 200) {
```
maximilianoertel: Small nit-pick, but the dashboard has `http-status-codes` installed so could we please replace any status number literals introduced by this PR with their named constants from `http-status-codes`? It's a small thing but the backend already uses them consistently, and it keeps the frontend aligned.


```suggestion
      if (result.status === StatusCodes.OK) {
```

apps/dashboard/src/composables/queries/useMeQuery.js
```diff
@@ -0,0 +1,74 @@
...
+    enabled: isQueryEnabled,
+    retry: (failureCount, error) => {
+      // Rostering-ended and terminal auth errors are not transient; retrying
+      // wastes time and delays the user-facing error UX.
+      if (isRosteringEndedError(error) || isTerminalAuthError(error)) {
+        return false;
+      }
+      return failureCount < MAX_RETRIES;
+    },
+    ...options,
```
maximilianoertel: The custom retry callback which prevents infinite retries on `auth/rostering-ended` and `auth/token-expired` is correctly set on the useQuery call, but since `...options` is spread after it, any caller could accidentally silently disable the non-retriable error handling. The spread order should be flipped: 


```suggestion
    ...options,
    enabled: isQueryEnabled,
    retry: (failureCount, error) => {
      // Rostering-ended and terminal auth errors are not transient; retrying
      // wastes time and delays the user-facing error UX.
      if (isRosteringEndedError(error) || isTerminalAuthError(error)) {
        return false;
      }
      return failureCount < MAX_RETRIES;
    },
```

apps/dashboard/src/store/auth.js
maximilianoertel, line 1: The changes introduced here and in `App.vue` copy server state into Pinia, effectively creating a second source of truth and violating our state management rule: https://github.com/yeatmanlab/roar-dashboard/blob/project/backend-refactor/.ai/rules/frontend-state-management.md#frontend-state-management 

The consumers that read from the store could read directly from the TanStack query instead.

Can we keep the store free of /me data and have consumers use the query directly?

apps/dashboard/src/router/index.js
```diff
@@ -899,6 +905,28 @@ router.beforeEach(async (to, from, next) => {
...
+  if (
+    store.hasUnsignedTos &&
+    to.name !== 'SignTos' &&
+    to.name !== 'SignIn' &&
+    to.name !== 'AccessEnded' &&
+    to.name !== 'GenericError'
+  ) {
+    next({ name: 'SignTos', query: { next: to.fullPath } });
+    return;
+  }
```
maximilianoertel, line 962: The guard uses magic route name strings instead of `APP_ROUTES` path constants that already exist (or should be added) in `constants/routes.js`

apps/dashboard/src/App.vue
```diff
@@ -58,13 +62,76 @@ const loadSessionTimeoutHandler = computed(() => isAuthStoreReady.value && authS
...
+    if (route.name !== 'SignIn') {
+      router.replace({ name: 'SignIn' });
+    }
+  } else {
+    setGlobalError({ type: 'server-error' });
+    if (route.name !== 'GenericError') {
+      router.replace({ name: 'GenericError' });
+    }
+  }
+});
```
maximilianoertel, line 141: Same as https://github.com/yeatmanlab/roar-dashboard/pull/1774/changes#r3266896392

Additionally, a new constant should be introduced for the global error types

apps/dashboard/src/constants/mutationKeys.js
```diff
@@ -1,6 +1,7 @@
 export const ADMINISTRATION_UPSERT_MUTATION_KEY = 'administration-upsert';
 export const ADMINISTRATION_DELETE_MUTATION_KEY = 'administration-delete';
 export const CONSENT_UPDATE_MUTATION_KEY = 'consent-update';
+export const RECORD_USER_AGREEMENT_MUTATION_KEY = 'record-user-agreement';
```
maximilianoertel: Small nitpick, but the existing mutation keys follow an entity-first convention (e.g. `administration-{upsert|delete}`. `record-user-agreement` is verb-first and should ideally be updated to match the existing convention

apps/dashboard/src/containers/SignTos/SignTos.vue
```diff
@@ -0,0 +1,112 @@
...
+    !next ||
+    !next.startsWith('/') ||
+    next.startsWith('//') ||
+    next.startsWith(APP_ROUTES.SIGN_TOS) ||
+    next.startsWith(APP_ROUTES.SIGN_IN)
+  ) {
+    return APP_ROUTES.HOME;
+  }
+  return next;
+});
```
maximilianoertel, line 51: The open-redirect validation could be more robust here. The `startsWith('/')` + `!startsWith('//')` checks works but `/\/evil.com` or `/%2F%2Fevil.com` would _could_ still pass depending on how the browser normalizes it before it hits the router. 

Realistically `router.replace` with an absolute URL won't navigate externally as, if I recall correctly, Vue Router treats it as a relative path and it would 404, so there is no true open redirect vulnerability here. 

However, since `next` actually only needs to support internal application paths, why not use a simple regex that only allows well-formed paths?

apps/dashboard/src/pages/SignTos.vue
```diff
@@ -0,0 +1,17 @@
...
+// Thin routing wrapper for the SignTos flow. All orchestration logic lives in
+// the container (containers/SignTos/SignTos.vue) per the container/presentational
+// pattern documented in `.ai/rules/frontend-layer-architecture.md`.
+//
+// Sign-out reachability: the global `Navigation` component renders even on
+// the SignTos page (it's not in `navbarBlacklist`), so the navbar sign-out
+// button remains the user's escape hatch. The router's TOS guard explicitly
+// allows transitions to `SignIn`, and `useSignOutMutation` redirects there
+// after clearing auth state, so sign-out succeeds without being intercepted
+// by the unsigned-TOS guard.
```
maximilianoertel: The inline documentation throughout this PR is generally strong (👏!) and most of the comments are genuinely valuable. However, some comments explain things that are self-evident from the code. 

For example, the multi-line block here explains that the page is a thin routing wrapper, that navigation and sign-out are handled elsewhere, and that router guards manage access control – none of which is relevant to this file. 

Comments like these add reading load without adding understanding, and they become maintenance liabilities when the code changes but the comment doesn't get updated.


```suggestion
<script setup>
```

maximilianoertel (changes requested):
Great to see the first backend endpoint getting integrated into the dashboard! 🎉

In addition to the inline comments, I wanted to flag a broader architectural concern with the TOS enforcement strategy. Right now there are two overlapping mechanisms doing the same job: the `meData` watcher in `App.vue` (which handles the initial load) and the `beforeEach` router guard (which handles subsequent navigations). 

As you commented in the code, the watcher exists because `/me` resolves after the router has already settled the initial route, so the guard misses it. The guard exists because the watcher doesn't fire on navigations.

In theory this covers a mid-session TOS change (e.g., a new version published while the user is active), but in practice both mechanisms are driven by the same trigger: a `/me` request response. 

The query is cached with a 10-minute stale time, so the guard never independently catches something the watcher doesn't. If a background refetch was to fire and `unsignedAgreements` changes, the watcher would handle it.

I'd prefer we pick one mechanism and commit to it. If we gate the app render on `/me` settlement, the router guard alone is sufficient for both initial load and subsequent navigations — no watcher, no "boot-time gap" comment, no two code paths with subtly different redirect behaviours. This also eliminates the flash where the user could briefly sees protected content before being redirected away. 

This ties into the inline comment about `meData` living in the Pinia store. If we keep TOS state in TanStack Query where it belongs and gate the render on query settlement, most of the complexity in App.vue and the store changes goes away.

richford:
@claude please review this PR. The scope has changed. Following @maximilianoertel's review, I broke this up into two PRs: #1822 now sets up the E2E infrastructure and this PR's scope has narrowed to wiring up `GET /me`. I have updated the PR description accordingly. I have also updated the originating ticket to the following:

> [!Note]
> Rewritten to reflect the PR split and the implementation decisions made during review.
> Ships as **PR 2** of a two-PR chain, stacked on the e2e tooling work.
> Supersedes the monolithic dashboard PR [1774](https://github.com/yeatmanlab/roar-dashboard/pull/1774).

## Summary

Introduce the backend `GET /me` call into dashboard app initialization as the single source of truth for the authenticated user's identity, type, and unsigned-TOS status, replacing the Firestore-based user fetch. Handle every `/me` outcome gracefully: rostering-ended users see a dedicated AccessEnded screen, users with unsigned TOS are blocked behind a signing flow, and server errors show GenericError with retry.

`/me` is owned by **TanStack Query**, not mirrored in Pinia, and `<router-view>` is gated on `/me` settlement so no protected content paints before a redirect can fire.

## Context

The dashboard currently loads user identity and claims from Firebase Auth + Firestore. As the first endpoint of the frontend→backend migration, `/me` becomes the canonical source for identity, type, and TOS status.

Three critical states are handled:
1. **Rostering ended** (`code: auth/rostering-ended`) → AccessEnded page.
2. **Unsigned TOS** (`unsignedAgreements` non-empty) → block navigation, present the TOS signing flow.
3. **Server error** (500 / network) → GenericError with retry.

## Scope

**In scope:** the `/me` query composable, app-init wiring, the TOS signing flow, and the three error states — all under `apps/dashboard/src/`.

**Out of scope:**
- The seeded local-backend **e2e test harness** — split into its own ticket (`ISSUE_e2e-tooling.md`); this work stacks on top of it.
- Backend endpoints (`/me`, agreements, the rostering-ended auth code) — separate backend tickets, already available on `project/backend-refactor`.

## Implementation

### 1. `useMeQuery`
TanStack Query composable for `GET /me`. Internally gated on `authStore.accessToken` (callers don't wire enablement). The retry policy is **caller-proof** — `...options` is spread *before* `enabled`/`retry` so a caller can't accidentally re-enable retries on terminal errors. No retry on `auth/rostering-ended` or terminal auth; up to 3 on transient failures. `ME_QUERY_KEY` is centralized in `constants/queryKeys.js`.

### 2. App initialization
`<router-view>` is gated on `/me` settlement (`isMeSettling`) so the destination cannot paint before TOS/error redirects resolve. On settle: 200 → TOS check; `auth/rostering-ended` → AccessEnded; terminal auth → SignIn; 500/network → GenericError.

### 3. State ownership — TanStack Query, not Pinia
`/me` data is **not** copied into the Pinia auth store (per `frontend-state-management.md`). A `useCurrentUser` composable derives identity (`id`, `userType`, name, `hasUnsignedTos`) directly from the `/me` query, and consumers read from it. There is **no** `meData`/`setMeData`/`hasUnsignedTos`/`currentUserId` mirror in the store.

### 4. TOS signing gate (container/presentational)
The router `beforeEach` guard is the **single owner** of TOS routing — it blocks navigation when `hasUnsignedTos`, allowing only `SignTos`, `SignIn`, `AccessEnded`, and `GenericError` (route names are named constants, not literals). There is no parallel App.vue watcher.

The `SignTos` page (thin wrapper) + container + `useTosSigningFlow` orchestrate signing: list unsigned agreements, fetch each via `GET /agreements/:agreementId/versions/:versionId`, render in the existing `ConsentModal`, and `POST /users/:userId/agreements` on acceptance (a 409 is treated as already-signed). After all are signed, invalidate `/me` and return to the original route. Locale selection is extracted into a `pickVersionForLocale` helper (match current locale, fall back to `en-US`). The `?next=` return path is validated with the WHATWG URL parser plus a strict internal-path check (extracted `isInternalPath`).

### 5. Error handling
`meError` → `globalError` mapping is consolidated in the QueryCache `onError`, using named global-error-type constants. `http-status-codes` constants are used instead of numeric literals throughout.

## Directory structure

```
src/
├── composables/
│   ├── useCurrentUser.js                         # identity from the /me query (no Pinia mirror)
│   └── queries/useMeQuery.js                      # GET /me
├── containers/SignTos/
│   ├── SignTos.vue                               # container
│   ├── isInternalPath.js                         # ?next= validation
│   └── composables/
│       ├── useTosSigningFlow.js
│       └── pickVersionForLocale.js
├── pages/SignTos.vue                             # thin wrapper
└── constants/{queryKeys.js, routes.js, ...}      # ME_QUERY_KEY, route names, error types
```
`store/auth.js` is **not** extended with `/me` data.

## Acceptance criteria

**/me integration**
- [ ] `GET /me` is called during app init after Firebase auth; the Firestore user fetch is replaced.
- [ ] Identity is exposed via `useCurrentUser` (from the `/me` query) — **not** mirrored in Pinia.
- [ ] `useMeQuery` follows the `computeQueryOverrides` pattern with a centralized query key; TanStack Query does not retry on `auth/rostering-ended`, and the retry policy can't be overridden by callers.

**Rostering ended**
- [ ] `auth/rostering-ended` → AccessEnded; user can't navigate away except sign-out, which works.

**TOS signing gate**
- [ ] Non-empty `unsignedAgreements` → SignTos; no other page is reachable until signed.
- [ ] Content fetched via `GET /agreements/:agreementId/versions/:versionId`; correct locale variant, `en-US` fallback; rendered in `ConsentModal`.
- [ ] Acceptance calls `POST /users/:userId/agreements` with the right `agreementVersionId`; multiple agreements must all be signed; afterward `/me` is invalidated and the user returns to the original route.

**Server errors**
- [ ] 500 → GenericError with a working retry; 401 → SignIn.

**General**
- [ ] No auth-flow regressions; container/presentational pattern for SignTos; a single TOS-enforcement mechanism (render gated on settlement + router guard); route names and error types are named constants; `?next=` cannot be an open redirect.

## Dependencies
- Stacks on the **e2e tooling** ticket (`ISSUE_e2e-tooling.md`) — targets that branch.
- Backend: `/me` (with `unsignedAgreements`), agreement content + consent endpoints, the `auth/rostering-ended` code — on `project/backend-refactor`.
- Supersedes PR [1774](https://github.com/yeatmanlab/roar-dashboard/pull/1774).

richford:
@claude please re-review

ksmontville (dismissed):

ksmontville (approved):

PR #1785, opened 2026-05-12
`ref`Initial Monorepo Migration

ksmontville:
## Summary

This PR migrates roar-pa into the monorepo as the exemplar assessment, establishing the patterns and tooling that all future assessment migrations will follow.

## Types of changes

- [ ] Documentation Update
- [x] Repository Maintenance

## Checklist


ksmontville:
@claude Let's review this first-pass monorepo transition. 

Do not review any of the original source code for roar-pa found in apps/assessments/roar-pa/src. 

Only review the monorepo infrastructure set up in /roar-pa root, /roar-pa/shared, and GitHub workflows.

ksmontville:
@claude Feedback addressed, let's review again.

ksmontville:
@claude I have added some clarifying comments, except on firebase.json which we are intentionally leaving as-as. Let's re-review.

ksmontville:
@claude We rebased onto project/backend-refactor. Let's review again, checking for regressions (only the files from the earlier review, not any source code from roar-pa).

amyyeung17 (dismissed):

.github/actions/deploy/action.yml
maximilianoertel, line 1: If we're going to repurpose this composite action, let's rename the action to `deploy-firebase-hosting` and update these: 

https://github.com/yeatmanlab/roar-dashboard/blob/00e00edad1fc124e94497256ab110afbf0675729/.github/actions/deploy/action.yml#L1-L6

package.json
```diff
@@ -21,15 +22,34 @@
...
+    "path-browserify": "^1.0.1",
     "prettier": "^3.6.0",
-    "turbo": "^2.5.8"
+    "process": "^0.11.10",
+    "style-loader": "^3.3.1",
+    "turbo": "^2.5.8",
+    "webpack": "^5.74.0",
+    "webpack-cli": "5.1.4",
+    "webpack-dev-server": "^4.11.0",
+    "webpack-merge": "^5.8.0"
```
maximilianoertel: The root `package.json` should exclusively contain `turbo` as dependency. All other dependencies should be declared directly in each app or package


```suggestion
    "turbo": "^2.5.8"
```

.npmignore
maximilianoertel, line 1: Can this file be removed given that there is nothing in this PR actually publishing to npm?

package.json
```diff
@@ -6,7 +6,8 @@
   },
   "version": "3.26.0",
   "scripts": {
-    "dev": "turbo run dev",
+    "dev": "turbo run dev --filter=!./apps/assessments/*",
+    "dev:pa": "turbo run dev --filter=./apps/assessments/roar-pa",
```
maximilianoertel: As the list will grow, can we standardise the command name as the following please?

```suggestion
    "dev:assessment:pa": "turbo run dev --filter=./apps/assessments/roar-pa",
```

.github/workflows/deploy-platform-production.yml
```diff
@@ -207,4 +207,16 @@ jobs:
...
+  # Deploy assessments
+  # ––––––––––––––––––––––––––––––––––––––––––––––– #
+  deploy-roar-pa:
+    needs: [approval]
+    uses: ./.github/workflows/deploy-assessment.yml
+    secrets: inherit
+    with:
+      environment: platform:production
+      assessment: roar-pa
+      build-script: build:prod
```
maximilianoertel: Platform releases and assessment releases are independent concerns and have different release conditions. Let's please remove this change and on a separate PR implement new `deploy-assessments-{staging/production}.yml` workflows as well as the necessary Release Please config changes


```suggestion
```

.github/workflows/deploy-platform-staging.yml
```diff
@@ -130,4 +130,17 @@ jobs:
...
+  # ––––––––––––––––––––––––––––––––––––––––––––––– #
+  deploy-roar-pa:
+    uses: ./.github/workflows/deploy-assessment.yml
+    secrets: inherit
+    with:
+      environment: platform:staging
+      assessment: roar-pa
+      build-script: build:staging
+      run-cypress: true
+      cypress-base-url: https://roar-phoneme-staging.web.app/
```
maximilianoertel: Same as https://github.com/yeatmanlab/roar-dashboard/pull/1785/changes#r3288719798


```suggestion
```

.github/workflows/deploy-assessment.yml
maximilianoertel, line 1: To follow the patterns/conventions currently established for GitHub Actions workflows, let's please remove this reusable workflow in favour of an a composite `build-assessment` action, analog to the existing `build` action. 

The `deploy-assessment-{env} worfklows requested [here](https://github.com/yeatmanlab/roar-dashboard/pull/1785/changes#r3288719798) should then mimic the structure of the existing `deploy-platform-{env}` workflows and call that action, as well as the composite `deploy-firebase-hosting` action

.github/workflows/deploy-assessment.yml
```diff
@@ -0,0 +1,119 @@
...
+
+# Reusable workflow for building and deploying a ROAR assessment to Firebase Hosting.
+# Convention: each assessment lives at apps/assessments/<name>/ with a firebase.json,
+# a dist/ output dir after build, and npm scripts: build:staging, build:prod.
+
+on:
+  workflow_call:
+    inputs:
+      environment:
+        description: "GitHub environment to deploy to (e.g. platform:staging)"
```
maximilianoertel: Just FYI, the Github environments `platform:{env}` are exclusively reserved to platform (backend + frontend + db + fga) deployments. Assessments should have their independent Github environment, e.g. `assessments:{env}`

.github/workflows/deploy-assessment.yml
```diff
@@ -0,0 +1,119 @@
...
+        run: npm run ${{ inputs.build-script }}
+
+      - name: Set Sentry environment in firebase.json
+        # firebase.json hardcodes sentry_environment=production in Report-To and CSP headers.
+        # For non-default channels (e.g. PR previews), patch it to the channel name so Sentry
+        # groups these events separately from production issues.
+        if: inputs.firebase-channel != 'default'
+        run: |
+          sed -i "s/sentry_environment=production/sentry_environment=${{ inputs.firebase-channel }}/g" \
+            apps/assessments/${{ inputs.assessment }}/firebase.json
```
maximilianoertel: Instead of this approach, we should use the same approach as established by @richford that uses a firebase.template.json and dynamically generates the final firebase.json file (see `apps/dashboard/firebase/admin/`)

.github/workflows/pr-preview-assessments.yml
maximilianoertel, line 1: A couples of points:

**1. Checked in as known-broken**. The opening comment says the workflow doesn't work. This is tech debt on arrival, as a general rule, we shouldn't include anything that isn't functional.

**2. Shouldn't be a separate workflow** _(for now)_. `ci.yml` is already the "everything that happens on
 a PR" workflow — lint, tests, build, and preview deploy. Assessment previews hence belong there too, not in a standalone workflow. In a separate step (but no priority) we should analyse and potentially revise  the CI workflows. 

**3. Hardcoded single assessment with no change detection**. These don't need to run when no changes were made to the assessments or shared assessment files. Additionally, each added assessment would need 
manual duplication.

Instead, we can add change detection and a matrix-driven preview job to `ci.yml` using  an action like `dorny/paths-filter`. Here's a rough/untested example: 

```yml
detect-assessment-changes:
  name: Detect changed assessments
  runs-on: ubuntu-latest
  outputs:
    matrix: '${{ steps.filter.outputs.changes }}'
  steps:
    - name: Checkout repository
      uses: actions/checkout@v6
    - name: Detect changes
      id: filter
      uses: dorny/paths-filter@v3
      with:
        filters: |
          roar-pa:
            - 'apps/assessments/roar-pa/**'
            - 'apps/assessments/shared/**'

e2e-assessments:
    name: 'E2E ${{ matrix.assessment }} (${{ matrix.browser }})'
    needs:
      - detect-assessment-changes
    if: 'needs.detect-assessment-changes.outputs.matrix != ''[]'''
    runs-on: ubuntu-latest
    timeout-minutes: 60
    strategy:
      fail-fast: false
      matrix:
        assessment: '${{ fromJson(needs.detect-assessment-changes.outputs.matrix) }}'
        browser:
          - chrome
          - edge
    steps:
      - name: Checkout repository
        uses: actions/checkout@v6
      - name: Setup environment
        uses: ./.github/actions/setup-node-environment
      - name: Cypress run
        uses: cypress-io/github-action@v6
        env:
          CYPRESS_RECORD_KEY: '${{ secrets.CYPRESS_RECORD_KEY }}'
          GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
        with:
          install: false
          build: 'npm run build:staging'
          start: npm run dev
          browser: '${{ matrix.browser }}'
          working-directory: 'apps/assessments/${{ matrix.assessment }}'
          wait-on: 'http://localhost:8080'
          wait-on-timeout: 120
          record: true
          parallel: true
          spec: cypress/e2e/**/*
          ci-build-id: '${{ github.run_id }}-${{ matrix.assessment }}-${{matrix.browser }}'

preview-assessments:
  name: 'Preview ${{ matrix.assessment }}'
  needs:
    - detect-assessment-changes
  if: 'needs.detect-assessment-changes.outputs.matrix != ''[]'''
  runs-on: ubuntu-latest
  environment: 'assessments:staging'
  timeout-minutes: 15
  strategy:
    fail-fast: false
    matrix:
      assessment: '${{ fromJson(needs.detect-assessment-changes.outputs.matrix) }}'
  steps:
    - name: Build assessment
      uses: ./.github/actions/build-assessment
      with:
        assessment: '${{ matrix.assessment }}'
        build-script: 'build:staging'
        firebase-channel: preview
    - name: Deploy preview
      uses: ./.github/actions/deploy-firebase-hosting
      with:
        firebase-channel: preview
        expires: 7d
        firebase-project-id: '${{ secrets.FIREBASE_PROJECT_ID }}'
        gcp-workload-identity-provider: '${{ secrets.GCP_WORKLOAD_IDENTITY_PROVIDER}}'
        gcp-service-account-id: '${{ secrets.FIREBASE_HOSTING_SERVICE_ACCOUNT_ID }}'
        github-token: '${{ secrets.GITHUB_TOKEN }}'
        dist-path: 'apps/assessments/${{ matrix.assessment }}/dist'
        firebase-config-path: 'apps/assessments/${{ matrix.assessment}}/firebase.json'

```

The current `deploy-assessment.yml` reusable workflow includes a Cypress job that runs post-deploy. However, E2E tests should not run against the deployed preview URL. Instead, just like the dashboard the E2E tests should run in the CI workflow before the deployment.

maximilianoertel (changes requested):
In addition to the inline comments, there's an architectural concern with how `roar-pa` is integrated into the monorepo. This PR copies `roar-pa` into apps/assessments/roar-pa/, adds it to the workspaces config, and replaces the dashboard's dependency from the npm package`@bdelab/roar-pa@2.2.5` to `@roar-dashboard/roar-pa`. Meaning: the dashboard's Vite build now bundles `roar-pa` from source. 

This couples assessment versioning to the platform release cycle, which conflicts our versioning and release strategy. So far, each assessment has had its own release cadence but the workspace * reference removes that independence. 

Assessments really should remain independently versioned and published packages. The monorepo can host assessment source under `apps/assessments/`, but the dashboard should depend on published versions (npm or GitHub Packages), not workspace references. Assessment build, versioning, and deployment and publishing should be fully independent from the platform and have its own Release Please branch — same as `assessment-sdk` for example

ksmontville:
@maximilianoertel Thanks for the thorough review.

A lot of your concerns regarding versioning, deployment, and CI/CD are going to be addressed in the next PR in this chain which addresses the monorepo versioning and deployment strategy.

The plan is to independently version the assessments, the `assessment-schema` package, and the `assessment-sdk` package under the `@roar-platform` package.

1. #1785 
2. #1791 
3. #1794
4. [Planned: Deployment Strategy](https://github.com/yeatmanlab/roar-project-management/issues/1825)
5. [Planned: Versioning Strategy](https://github.com/yeatmanlab/roar-project-management/issues/1818)

For this initial brute-force migration, the goal was to just establish a working researcher development environment that runs locally and is purely ephemeral. We need this proof of concept to function in order to get the researchers and assessment developers on board. Without their approval, a versioning and deployment strategy isn't worth much. So for speed and simplicity we keep everything in source code for these initial implementations.

In the immediate, I can remove the defunct CI workflows, remove `roar-pa` from the platform deployment workflow,  and move assesssment dependencies back into their respective `package.json` as well as remove the unused `.npmignore`. I'm also going to remove the `dev:pa` and similar scripts from root. These scripts are deprecated in step 3 by the ephemeral researcher environment, anyway.

Your suggestions regarding assessment preview environments and assessment deployment specifics will all be implemented in step 4 when we take a more holistic view of versioning and deployment. Once that is in place for `roar-pa`, we can start migrating the rest of the assessments into the monorepo using these patterns.

apps/dashboard/cypress/e2e/participant/default-tests/assessments/playPA.cy.js
```diff
@@ -1,10 +1,10 @@
 import { playPA } from '../../../../support/helper-functions/roar-pa/paHelpers';
 import { isCurrentVersion } from '../../../../support/utils';
 
-const app = '@bdelab/roar-pa';
+const app = '@roar-dashboard/roar-pa';
 let isCurrentAppVersion;
 
-describe('Participant Assessment: ROAR Phoneme', () => {
+describe.skip('Participant Assessment: ROAR Phoneme', () => {
```
richford, line 8: I agree with the decision to convert all of these specs from `describe(...)` to `describe.skip(...)`. The PR description only mentions PA, but the same change lands across ARF, CALF, CVA, Letter, Morphology, SRE, SWR, Syntax, Vocab, and all their network-test variants.

Can you add a tracking ticket to reevaluate the Cypress tests for playing the assessments through the dashboard? We need to decide when/if/how we want to re-enable them. And then can you add a one-line `// SKIPPED until #<ticket>: <reason>` comment immediately above each top-level `describe.skip(...)` lines in each file. I know it's tedious but I think it will be helpful to document the temporary decision here so that we don't lose the thread.

apps/assessments/roar-pa/package.json
```diff
@@ -0,0 +1,79 @@
...
+  "name": "@roar-dashboard/roar-pa",
+  "description": "This is a 3-AFC task to measure phonological awareness",
+  "version": "5.1.0",
+  "author": "AnyaWMa <anyawma08@gmail.com>",
+  "main": "src/index.js",
+  "module": "src/index.js",
+  "type": "module",
+  "scripts": {
+    "//": "build is a no-op: Vite resolves imports via 'main' (src/index.js) directly, so no webpack output is needed for dashboard#dev.",
+    "build": "echo 'roar-pa: skipped (Vite uses src directly)'",
```
richford, line 11: The `"build"` script being `echo 'roar-pa: skipped …'` is misleading and breaks the Turbo task contract. From the root, `turbo run build` will report `roar-pa#build` as successful while producing no artifacts; any downstream task that adds `dependsOn: ['^build']` later will see a phantom-success cycle and silently consume stale or missing output.

Two suggestions:

1. Keep the source-resolved Vite path working but rename this task so its name matches its function, and either delete build or point it at the existing webpack config:

```json
"build:noop": "echo 'roar-pa: skipped (Vite uses src directly)'",
"build": "webpack --mode production --env dbmode=production"
```

2. Or, if `build` really should remain a no-op for now, declare that to Turbo so the graph doesn't lie. Add `apps/assessments/roar-pa/turbo.json`:

```json
{
  "extends": ["//"],
  "tasks": {
    "build": { "outputs": [] }
  }
}
```

Your second PR in the chain (#1791) adds a cross-workspace dependency on `@roar-dashboard/assessment-schema` without wiring `roar-pa#build` into `dependsOn`, so fixing this here helps for the next PR in the chain.

apps/assessments/roar-pa/package.json
richford, line 1: Can you also add an `apps/assessments/roar-pa/README.md` file documenting:

- that the dashboard resolves PA from `src/index.js` via Vite (not from `dist/`)
- why `build` is intentionally a no-op (assuming you go with option 2 in the build-script comment above)
- that the standalone `webpack.config.cjs` still exists for `dev` / `build:dev`

Not blocking.

richford (changes requested):
Thanks @ksmontville for taking this one on. I think scope-wise this PR is clean. I have some inline requests below. And two broader architectural notes here:

- @maximilianoertel's "assessments as published packages vs. workspace refs" question is the only real architectural call this PR opens. And you already acknowledged it as a follow-up. I'm not asking you to re-litigate it here, but please make sure that follow-up actually lands in a chained PR.

- The loose `apps/assessments/shared/` directory with Firebase and Sentry config isn't a real workspace package, so it's resolved via relative imports. I think this is fine for one assessment, but when we migrate more, we need to consider converting this to `packages/assessment-shared` or similar.

apps/dashboard/cypress/e2e/participant/network-tests/roar-morphology/playMorphology-3G.cy.js
richford, line 1: What's with the ` copy` in this filename? Is this a development artifact? Can you rename the file please?

richford (changes requested):
Thanks @ksmontville. You're changes resolved all of my previous comments. I just found this one file that has a weird filename. Other than that, LGTM!

apps/dashboard/cypress/e2e/participant/network-tests/roar-morphology/playMorphology-3G.cy.js
ksmontville: This is actually pre-existing from either `main` or `project/backend-refactor` but I'll rename it here.

richford (approved):

maximilianoertel (approved):
Thanks for the context and removing the deployment-related code from this PR @ksmontville. I didn't manually test the assessment but overall, everything looks good.


One concern though: `apps/assessments/roar-pa/src/assets/` contains a large number of audio and image files checked into the repo. I know this isn't introduced by this PR as it's just migrating existing code, but now that it's in the monorepo it will impact clone times, CI performance, etc. Could we please file a ticket to move assessment static assets (audio, images) to a GCS bucket served via CDN? (cc @richford)

PR #1787, opened 2026-05-14
Add `POST /v1/families` endpoint

richford:
## Summary

Adds a public `POST /v1/families` endpoint that registers a new caretaker, creates a new family, and links the two with `role=parent` in a single atomic operation. This is step 1 of the two-step ROAR@Home registration flow; step 2 (adding children) lands in a PR chained off this branch.

The endpoint replaces the `createNewFamily` Firebase Cloud Function. It is intentionally public — there is no `AuthGuardMiddleware` on the `create` handler — because the caretaker has no identity at the point of registration. Abuse is bounded by the WAF (100 req/s global) and by a new partial unique index on `families.created_by` that enforces "one family per caretaker" at the database level. Tightening to per-IP throttling and a captcha challenge is deferred to follow-up tickets.

## Data flow

```
POST /v1/families  (no auth)
  → FamiliesController.create(body)
    → FamilyService.create(input)
      → Step 1: pre-flight email check (DB + Firebase getUserByEmail)
        → 409 if either hits
      → Step 2: FirebaseAuthClient.createUser({ email, password, displayName })
        → 409 on auth/email-already-exists
        → 429 on auth/too-many-requests
        → 500 on anything else
      → Step 3: DB transaction
          - INSERT users (userType=caregiver, authProvider=[password], authId=firebaseUid)
          - INSERT families (createdBy=caretakerId)
              ↳ families_created_by_uniq_idx fires → 422
          - INSERT user_families (caretakerId, familyId, role=parent)
          - INSERT rostering_provider_ids (provider=dashboard, partner=familyId, entity=caretaker)
        On any failure: rollback DB + compensate-delete the Firebase account
      → Step 4: FGA writeTuplesOrThrow([parent on family])
        On failure: deleteTuples → rollback DB rows (rostering_provider_ids → user_families → families → users) → compensate-delete Firebase
    ← { id: familyId }
```

### Migration

- **`0063_add_families_created_by.sql`** — adds `families.created_by uuid` with a `RESTRICT` foreign key to `users(id)` and a partial unique index `families_created_by_uniq_idx WHERE created_by IS NOT NULL`. The column is nullable so existing rows and any future admin-tool creations remain valid; the partial index only fires when `created_by` is set. The Drizzle snapshot at `meta/0063_snapshot.json` was hand-built to match the existing snapshot format because `drizzle-kit generate` can't run in our sandbox environment, but the SQL itself is what `db:migrate:core` will execute against the database.

[...]

PR #1788, opened 2026-05-14
Add `POST /v1/families/:familyId/users` endpoint

richford:
## Summary

Adds an authenticated `POST /v1/families/:familyId/users` endpoint that lets a caretaker (or super admin) add one or more children to an existing family. Each child gets a Firebase Auth account, a `users` row with `userType=student`, a `user_families` row with `role=child`, a `user_groups` row for the group their activation code resolves to, and a `rostering_provider_ids` row with `partnerId=familyId`. The endpoint enforces a family-size cap of 12 and rejects duplicate emails within the request.

This is step 2 of the two-step ROAR@Home registration flow. It chains off #1787 which is a prerequisite to merge first.

Resolves [1751](https://github.com/yeatmanlab/roar-project-management/issues/1751)

## Data flow

```
POST /v1/families/:familyId/users  (auth required)
  → AuthGuardMiddleware
  → FamiliesController.addChildren(authContext, familyId, body)
    → FamilyService.addChildren
      → Step 1: family exists + not rostered out  (404 / 422)
      → Step 2: parent of family OR super admin   (403 otherwise)
      → Step 3: resolve activation codes via invitation_codes
                (dedupe; 422 if any invalid/expired/group inactive)
      → Step 4: existing user_families + children.length ≤ FAMILY_SIZE_LIMIT  (422)
      → Step 5: pre-flight email uniqueness (DB + Firebase)
                (400 in-request dups; 409 if any email taken anywhere)
      → Step 6: Firebase createUser per child (sequential, cumulative rollback)
                On failure: delete UIDs created so far, surface 409/429/500
      → Step 7: DB transaction
                - INSERT users × N
                - INSERT user_families (role=child) × N
                - INSERT user_groups (role=student) × N
                - INSERT rostering_provider_ids × N
                On failure: roll back DB + delete every Firebase UID
      → Step 8: FGA writeTuplesOrThrow (2 tuples per child: family + group)
                On failure: deleteTuples → rollback DB rows → delete Firebase
    ← { ids: [...] }
```

### API contract (`packages/api-contract/src/v1/families/`)

- **`schema.ts`**:
  - `FAMILY_SIZE_LIMIT = 12` exported constant — used both as the `.max()` on the children array (validation layer) and as the runtime cap in the service
  - `AddChildSchema` — per-child shape, `.strict()`, required `email/password/name/dob/grade/activationCode`, optional `demographics` (matches `CreateUserDemographicsSchema` from POST /users)
  - `AddFamilyChildrenRequestSchema` — `.strict()`, `children` array constrained to `[1, FAMILY_SIZE_LIMIT]`
  - `AddFamilyChildrenResponseSchema` — `{ ids: string[] }`, ids in request order
- **`contract.ts`**: new `addChildren` action — `POST /:familyId/users`, body `AddFamilyChildrenRequestSchema`, responses 201/400/401/403/404/409/422/429/500, `strictStatusCodes: true`

### Repositories

- **`InvitationCodeRepository.findValidByCode(code)`** — looks up a code by its `code` value (case-sensitive — codes are opaque) and returns the row only if `validFrom ≤ now()` and `validTo` is unset or in the future. Used by the service to resolve activation codes; the Firebase-era "2 months old" heuristic is gone, replaced by the schema's explicit `validFrom`/`validTo`.
- **`FamilyRepository.addChildren(childInserts, transaction)`** — inserts N user rows, N user_families rows, and the group memberships in one transaction. The rostering_provider_ids rows are NOT written here; the service writes them via `rosterProviderIdRepository.create({ ..., transaction })` in the same transaction. Mirrors the `createWithCaretaker` split.
- **`FamilyRepository.countActiveMembers(familyId)`** — count of `user_families` with `leftOn IS NULL`. Used for the size cap.
_(There is no SQL-based role-lookup helper in this PR. An earlier draft introduced `FamilyRepository.getUserRolesInFamily` against the inactive backend-authorization-pattern rule that described an SQL-first model — see the rule update below. The current code uses FGA exclusively, matching every other service in the backend.)_

### Service (`apps/backend/src/services/family/family.service.ts`)

- New `addChildren` function implements the saga above. Mirrors the patterns from `FamilyService.create` and `UserService.create` (pre-flight → Firebase → DB transaction → FGA, with full compensation at each step).
- **Cumulative Firebase rollback**: Firebase accounts are created one-by-one in step 6. If creating child N fails, the UIDs already created for children 1..N-1 are deleted before the error is surfaced. This is the same shape as the bulk-import pattern documented in `UserService` but applied per-request rather than per-batch.
- **Activation code dedup**: two children sharing a code resolve via one `invitation_codes` lookup; each child still gets their own `user_groups` row.
- **FGA tuple shape**: per child, one `(child on family)` tuple plus one `(student on group)` tuple. Both use `active_membership` conditions via `familyMembershipTuple` / `groupMembershipTuple`.
- **Compensation order** on FGA failure: `deleteTuples` → DB transaction (`rostering_provider_ids` → `user_groups` → `user_families` → `users`) → Firebase deletes. The DB delete order satisfies the trigger/FK constraints — `prevent_rostered_entity_delete` blocks user deletion while the rostering row exists, and `user_families.familyId` is `ON DELETE RESTRICT`.

### Controller (`apps/backend/src/controllers/families.controller.ts`)

- New `addChildren` handler — passes `authContext` and `familyId` through, maps `AddFamilyChildrenRequest.children` to the service input. Translates ApiError to the typed ts-rest responses (400, 403, 404, 409, 422, 429, 500); non-ApiError re-throws to the global handler.

### Route (`apps/backend/src/routes/families.ts`)

- New `addChildren` entry on the `FamiliesRoutes` router. Uses `AuthGuardMiddleware` (unlike the `create` handler which is public).

### Test mocks

- New `apps/backend/src/test-support/repositories/invitation-code.repository.ts` mirroring the convention for the other repos
- `MockFamilyRepository` gains the three new methods

## Decisions worth flagging

- **Auth check uses FGA, matching every other backend service.** An earlier draft used a direct SQL lookup of `user_families.role`, citing a presumed race between the DB commit and FGA tuple write during family creation. That justification didn't survive a deeper audit:
  1. Every other service in the backend uses `authorizationService.requirePermission(...)` for runtime authorization — the SQL-based pattern (`AccessControlFilter`, `access-controls/` subqueries, etc.) has been fully removed from the codebase. `addChildren` was the lone outlier.
  2. The race window I was worried about isn't observable: the `POST /v1/families` saga rolls back the DB row if the FGA write fails, so external callers can't see a family where DB has the parent role but FGA doesn't. The transient window between DB commit and FGA write inside the saga is small, the FGA backfill job exists to reconcile any drift, and a transient 403 during that window is no worse than the 403 the SQL check would return anyway.
  3. The FGA model already defines `define can_create_child: parent` on the family type, so no model change is needed — just a new `CAN_CREATE_CHILD` entry in `FgaRelation`.

  The branch also rewrites `.ai/rules/backend-authorization-pattern.md` to reflect the FGA-first reality the codebase has been on for a while — that's a separate, self-contained commit and could land on its own if you prefer.
- **Children are added to groups, not orgs**. The original ticket called for `user_orgs` writes for children, but groups in the new schema are flat — they have no parent org tree — so there's nothing to enroll into. Confirmed and dropped from scope per design discussion.
- **No "subgroup" creation**. The Firebase model created a per-family subgroup named `{orgName} - Family {familyId}`. The new model has no such concept. Activation codes map directly to existing groups, and the child joins the group the code resolves to.
- **`caretakerEmail` / `caretakerData` / `consentData` dropped** from the request body — the caretaker's identity comes from the auth token, parent updates go through `PATCH /v1/users/:id`, and consent is a separate domain.
- **Family-size cap of 12** is enforced both at the validation layer (`.max(FAMILY_SIZE_LIMIT)` on the children array) and at the service layer (`existing_members + children.length <= FAMILY_SIZE_LIMIT`). The constant is exported from the api-contract so both layers reference one source.
- **`generateAssessmentPid({ userId: email })`** is called per child to set `assessmentPid` server-side, matching the convention from POST /users and the POST /families caretaker creation.
- **`rosteringEnded` semantics are temporal** in the two checks this PR introduces (family active check, group active check). A non-null value blocks add-children only if it's `<= now()`; future-dated values are treated as "scheduled to end on X" and still permit adds. This matches the existing convention in `class.service.ts` and `school.service.ts` for parent-entity validity. The codebase is inconsistent on this — repository listing filters and `getActiveById` use strict `isNull(rosteringEnded)` — and that broader harmonization is a deliberate follow-up (draft at `.do-not-track/followup-rostering-ended-temporal.md`).

### Unit tests (`family.service.add-children.test.ts`)

- Happy path with N=2 children: returns ids, writes all four row types, writes 2 FGA tuples per child (family + group)
- 404 on missing family, 422 on rostered-ended family
- 403 on non-parent caller, super admin bypass
- 422 on invalid/expired activation code, 422 on rostered-out resolved group
- Code dedup verified — two children with the same code → one `findValidByCode` call
- 422 when existing + new exceeds the size cap; allows the request that exactly fills the cap
- 400 on case-insensitive duplicate email within the request
- 409 on email already in DB or Firebase Auth
- Cumulative Firebase rollback: when createUser for child N fails, UIDs for 1..N-1 are deleted
- 429 on `auth/too-many-requests`
- DB transaction failure → every Firebase account rolled back
- DB unique-violation race during INSERT → 409
- FGA write failure → tuple delete + DB rollback transaction + Firebase delete attempted before 500
- FGA compensation with DB rollback transaction failing → manual-cleanup warning logged (asserted)

### Controller tests (`families.controller.test.ts`)

- `addChildren` describe block uses `it.each` over the six error status codes (403, 404, 409, 422, 429, 500) to keep the file flat; also covers 201 happy path and non-ApiError re-throw

### Integration tests (`families.integration.test.ts`)

- New `POST /v1/families/:familyId/users` describe block, fixture-isolated from the existing `GET /:familyId/users` and `POST /v1/families` blocks
- Happy path asserts the full DB row set: users (`userType='student'`, `authProvider=['password']`, `assessmentPid` non-null), user_families (role='child'), user_groups (role='student'), rostering_provider_ids (`partnerId=familyId`, `providerType='dashboard'`)
- 401 unauthenticated, 403 non-parent
- 404 missing family, 422 unknown code, 422 expired code (`validTo` in the past)
- 400 empty children array, 400 case-insensitive duplicate email within the request
- 409 email already in the users table

## Out of scope

- Per-IP rate limiting — follow-up
- Captcha verification — follow-up

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist

### DIFF
--- a/apps/backend/src/repositories/invitation-code.repository.ts
+++ b/apps/backend/src/repositories/invitation-code.repository.ts
@@ -1,4 +1,4 @@
-import { and, desc, eq, gte, isNull, lte, or, sql } from 'drizzle-orm';
+import { and, desc, eq, gt, isNull, lte, or, sql } from 'drizzle-orm';
 import type { NodePgDatabase } from 'drizzle-orm/node-postgres';
 import { BaseRepository } from './base.repository';
 import type { InvitationCode } from '../db/schema';
@@ -24,7 +24,13 @@ export class InvitationCodeRepository extends BaseRepository<InvitationCode, typ
    * Filters by:
    * - `code` matches exactly (case-sensitive — codes are opaque)
    * - `validFrom <= NOW()`
-   * - `validTo >= NOW() OR validTo IS NULL`
+   * - `validTo > NOW() OR validTo IS NULL`
+   *
+   * The `validTo` check is strict (`>`, not `>=`) to match the convention in
+   * `isActiveRoster` and the rest of the codebase: a timestamp marks the
+   * *first instant* the resource is no longer valid, not the last instant it
+   * is. The boundary case is unobservably narrow in practice but consistency
+   * matters.
    *
    * Used by `POST /v1/families/:familyId/users` to resolve the caller-supplied
    * activation code to a group before enrolling a child. Returns `null` for
@@ -43,7 +49,7 @@ export class InvitationCodeRepository extends BaseRepository<InvitationCode, typ
         and(
           eq(invitationCodes.code, code),
           lte(invitationCodes.validFrom, now),
-          or(gte(invitationCodes.validTo, now), isNull(invitationCodes.validTo)),
+          or(gt(invitationCodes.validTo, now), isNull(invitationCodes.validTo)),
         ),
       )
       .limit(1);
@@ -57,9 +63,10 @@ export class InvitationCodeRepository extends BaseRepository<InvitationCode, typ
    * Filters by:
    * - groupId matches
    * - validFrom <= NOW()
-   * - validTo >= NOW() OR validTo IS NULL (no expiration)
+   * - validTo > NOW() OR validTo IS NULL (no expiration)
    *
-   * Orders by created_at DESC and returns the most recent one.
+   * Orders by created_at DESC and returns the most recent one. The `validTo`
+   * check is strict (`>`) — see `findValidByCode` for rationale.
    *
    * @param groupId - Group UUID
    * @returns The latest valid invitation code or null if none found
@@ -74,7 +81,7 @@ export class InvitationCodeRepository extends BaseRepository<InvitationCode, typ
         and(
           eq(invitationCodes.groupId, groupId),
           lte(invitationCodes.validFrom, now),
-          or(gte(invitationCodes.validTo, now), isNull(invitationCodes.validTo)),
+          or(gt(invitationCodes.validTo, now), isNull(invitationCodes.validTo)),
         ),
       )
       .orderBy(desc(invitationCodes.createdAt))
--- a/apps/backend/src/repositories/invitation-code.repository.ts
+++ b/apps/backend/src/repositories/invitation-code.repository.ts
@@ -19,6 +19,39 @@ export class InvitationCodeRepository extends BaseRepository<InvitationCode, typ
   }
 
   /**
+   * Look up an invitation code by its `code` value, only returning rows that are currently valid.
+   *
+   * Filters by:
+   * - `code` matches exactly (case-sensitive — codes are opaque)
+   * - `validFrom <= NOW()`
+   * - `validTo >= NOW() OR validTo IS NULL`
+   *
+   * Used by `POST /v1/families/:familyId/users` to resolve the caller-supplied
+   * activation code to a group before enrolling a child. Returns `null` for
+   * unknown or expired codes so the caller can produce a 422.
+   *
+   * @param code - The invitation code value (e.g. `'ABC123'`)
+   * @returns The invitation code row if valid, or null
+   */
+  async findValidByCode(code: string): Promise<InvitationCode | null> {
+    const now = sql`NOW()`;
+
+    const result = await this.db
+      .select()
+      .from(invitationCodes)
+      .where(
+        and(
+          eq(invitationCodes.code, code),
+          lte(invitationCodes.validFrom, now),
+          or(gte(invitationCodes.validTo, now), isNull(invitationCodes.validTo)),
+        ),
+      )
+      .limit(1);
+
+    return result[0] ?? null;
+  }
+
+  /**
    * Get the latest valid invitation code for a group.
    *
    * Filters by:
--- a/packages/api-contract/src/v1/families/schema.ts
+++ b/packages/api-contract/src/v1/families/schema.ts
@@ -1,12 +1,7 @@
 import { z } from 'zod';
-import { IDENTIFIER_WITH_SPACES } from '../common/regex';
 import { createPaginatedResponseSchema } from '../common/query';
-import {
-  UserSchema,
-  EnrolledUsersBaseQuerySchema,
-  UserGradeSchema,
-  FreeReducedLunchStatusSchema,
-} from '../common/user';
+import { UserSchema, EnrolledUsersBaseQuerySchema, UserGradeSchema } from '../common/user';
+import { CreateUserDemographicsSchema, CreateUserNameSchema } from '../users/schema';
 
 export const UserFamilyRoleSchema = z.enum(['parent', 'child']);
 
@@ -55,16 +50,6 @@ export const FamilyLocationSchema = z.object({
 export type FamilyLocation = z.infer<typeof FamilyLocationSchema>;
 
 /**
- * Name schema for the caretaker registering a new family. Reuses the same
- * regex as POST /users to keep validation consistent across signup paths.
- */
-const CreateFamilyCaretakerNameSchema = z.object({
-  first: z.string().regex(IDENTIFIER_WITH_SPACES),
-  middle: z.string().regex(IDENTIFIER_WITH_SPACES).optional(),
-  last: z.string().regex(IDENTIFIER_WITH_SPACES),
-});
-
-/**
  * Request body for POST /families.
  *
  * This endpoint registers a new caretaker (a `users` row with
@@ -72,6 +57,9 @@ const CreateFamilyCaretakerNameSchema = z.object({
  * family they're the parent of. The flat field shape mirrors POST /users
  * rather than wrapping caretaker fields in a `caretakerData` envelope — the
  * server already knows everything in the body belongs to the caretaker.
+ *
+ * `name` reuses the canonical `CreateUserNameSchema` from the users contract
+ * to keep validation rules consistent across signup paths.
  *
  * Excluded from this schema:
  * - userType / authProvider — server-set to caregiver / [password]
@@ -86,7 +74,7 @@ export const CreateFamilyRequestSchema = z
   .object({
     email: z.string().email().max(255),
     password: z.string().min(8),
-    name: CreateFamilyCaretakerNameSchema,
+    name: CreateUserNameSchema,
     location: FamilyLocationSchema.optional(),
   })
   .strict();
@@ -120,31 +108,10 @@ export type CreateFamilyResponse = z.infer<typeof CreateFamilyResponseSchema>;
 export const FAMILY_SIZE_LIMIT = 12;
 
 /**
- * Name schema for a child being added to a family. Same regex as the caretaker
- * (and POST /users) for consistency across signup paths.
- */
-const AddChildNameSchema = z.object({
-  first: z.string().regex(IDENTIFIER_WITH_SPACES),
-  middle: z.string().regex(IDENTIFIER_WITH_SPACES).optional(),
-  last: z.string().regex(IDENTIFIER_WITH_SPACES),
-});
-
-/**
- * Optional demographic fields for a child. Mirrors `CreateUserDemographicsSchema`
- * in the users contract; defined here to keep the families contract self-contained.
- */
-const AddChildDemographicsSchema = z.object({
-  gender: z.string().nullable().optional(),
-  race: z.string().nullable().optional(),
-  statusEll: z.string().nullable().optional(),
-  statusFrl: FreeReducedLunchStatusSchema.nullable().optional(),
-  statusIep: z.string().nullable().optional(),
-  hispanicEthnicity: z.boolean().nullable().optional(),
-  homeLanguage: z.string().nullable().optional(),
-});
-
-/**
  * Per-child input for `POST /v1/families/:familyId/users`.
+ *
+ * `name` and `demographics` reuse the canonical shapes from the users contract
+ * — every user-creation path on the platform validates these fields the same way.
  *
  * `activationCode` is required and must resolve to an active group via the
  * `invitation_codes` table at the service layer. `dob` must be a valid past
@@ -155,11 +122,11 @@ export const AddChildSchema = z
   .object({
     email: z.string().email().max(255),
     password: z.string().min(8),
-    name: AddChildNameSchema,
+    name: CreateUserNameSchema,
     dob: z.string().date(),
     grade: UserGradeSchema,
     activationCode: z.string().min(1),
-    demographics: AddChildDemographicsSchema.optional(),
+    demographics: CreateUserDemographicsSchema.optional(),
   })
   .strict();
 
--- a/packages/api-contract/src/v1/families/schema.ts
+++ b/packages/api-contract/src/v1/families/schema.ts
@@ -1,4 +1,5 @@
 import { z } from 'zod';
+import { IDENTIFIER_WITH_SPACES } from '../common/regex';
 import { createPaginatedResponseSchema } from '../common/query';
 import { UserSchema, EnrolledUsersBaseQuerySchema } from '../common/user';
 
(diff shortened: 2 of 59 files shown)


richford:
@claude, please review this PR. Here is the originating issue for reference:

## Summary

Adds an authenticated `POST /v1/families/:familyId/users` endpoint that lets a caretaker add one or more children to an existing family. Each child gets a Firebase Auth account and a `users` row, is linked to the family via `user_families` with `role='child'`, and is enrolled in the group corresponding to the activation code they were given.

This is the second step of the two-step ROAR@Home registration flow. Step 1 (caretaker + family creation) is in the previous PR in the chain; that PR is a prerequisite of this one.

The endpoint enforces a max family size of 12 (`existing user_families count + len(children) ≤ 12`) and a one-to-one activation-code → group mapping. There are no "subgroups" — a child enrolled with activation code `X` joins exactly the group that `X` resolves to in the `invitation_codes` table.

## Context

This endpoint replaces the `addStudentToFamily` function in roar-firekit. The Firebase implementation:

- Validated activation codes against the Firestore `orgs` collection
- Created or found a per-family "subgroup" named `{orgName} - Family {familyId}`
- Created student accounts in Firebase Auth and in two Firestore projects (admin + assessment)
- Updated parent org subgroup membership to trigger a downstream assignment sync
- Performed operations sequentially with limited atomicity

The Postgres implementation:

- Validates activation codes against the `invitation_codes` table directly (`validFrom ≤ now()` and `(validTo IS NULL OR validTo > now())`)
- Has **no subgroup concept** — the activation code resolves to a single existing group, and the child joins that group via `user_groups` with `role='student'`
- Has **no `family_groups` join table** — there is no such table in the new schema, and the original ticket's reference to `pg_family_groups` was hallucinated
- Has **no assessment-DB writes** — the assessment project is no longer the source of user identity for students added this way
- Has **no administration-assignment-sync enqueue** — assignment visibility is computed on-demand by the reporting endpoints
- Does **not** add children to `user_orgs` — groups are flat and don't have parent orgs, so org-tree membership doesn't apply here
- Runs all DB writes inside a single transaction; on transaction failure, all created Firebase Auth accounts are deleted via the same compensation pattern used by `POST /v1/users`

The order of operations:

1. Authorize: caller must be the authenticated caretaker of `:familyId` (a `user_families` row with `userId=req.user.userId`, `familyId=:familyId`, `role='parent'`) **or** a super admin.
2. Validate the family exists and is active (`rosteringEnded IS NULL`).
3. Enforce the family-size cap: `count(user_families WHERE familyId=:familyId AND leftOn IS NULL) + len(children) ≤ 12`. Return 422 otherwise.
4. Resolve each child's `activationCode` to a `(invitationCodeId, groupId)` pair. Return 422 if any code is invalid, expired, or not associated with an active group. **Deduplicate** activation codes within the request — multiple children using the same code share the same group enrollment row.
5. Pre-flight: check that every child email is unique against `users` and against Firebase Auth (best-effort race-safe, DB constraint is the true defense).
6. Create Firebase Auth accounts for all children. Collect UIDs. On any single failure, delete all UIDs created so far and abort with 409 (for `auth/email-already-exists` on a later child) or 500.
7. Open one DB transaction:
   - For each child: insert into `users` (`userType='student'`, `authProvider=['password']`, `authId=firebaseUid`, demographics from request), returning `childUserId`.
   - For each child: insert into `user_families` (`familyId`, `userId=childUserId`, `role='child'`, `joinedOn=now()`).
   - For each unique group from step 4: insert into `user_groups` for each child (`groupId`, `userId=childUserId`, `role='student'`, `enrollmentStart=now()`).
   - For each child: insert into `rostering_provider_ids` (`providerType='dashboard'`, `providerId=childUserId`, `partnerId=familyId`, `entityType='user'`, `entityId=childUserId`).
8. On transaction failure, delete every Firebase Auth account created in step 6.
9. Write FGA tuples for each child's `child` relation to the family and each child's group memberships. On FGA failure, delete tuples (best-effort), then DB rows (in a single rollback transaction with rostering_provider_ids first), then Firebase accounts.

## Endpoint

- **Method**: `POST`
- **Route**: `/v1/families/:familyId/users`

## Authentication

`AuthGuardMiddleware` required.

## Authorization

Service-layer access control:

- **Super admin** (`req.user.isSuperAdmin`): allowed, no further checks
- **Caretaker of the target family**: must have a `user_families` row with `userId=req.user.userId`, `familyId=:familyId`, `role='parent'`, `leftOn IS NULL`. FGA's `can_add_child` relation on `family:<familyId>` is the canonical check; the SQL fallback is documented for reference.
- **All other users**: 403 Forbidden

The original ticket included "District admins can add students to families within their organizations." That's dropped — district admins have no relationship to families, families are not part of the district org tree, and this endpoint is squarely a caretaker self-service operation.

## Request Specifications

### Request Body

```ts
{
  children: ChildInput[];
}

type ChildInput = {
  email: string;
  password: string;
  name: {
    first: string;
    middle?: string;
    last: string;
  };
  dob: string;          // ISO-8601 date (YYYY-MM-DD)
  grade: Grade;         // values from the `grade` enum
  activationCode: string;

  // optional demographics — match the existing `users` table columns
  gender?: string;
  statusEll?: string;
  statusIep?: string;
  statusFrl?: 'free' | 'reduced' | 'paid';   // free_reduced_lunch_status enum
  race?: string;
  hispanicEthnicity?: boolean;
  homeLanguage?: string;
};
```

The original ticket's `caretakerEmail`, `caretakerData`, and `consentData` fields are dropped:

- `caretakerEmail`: redundant — the caretaker's identity comes from `req.user`
- `caretakerData`: parent profile updates are a separate concern (use `PATCH /v1/users/:id`)
- `consentData`: consent is a separate domain not yet modelled in this PR; will be added via a dedicated endpoint when needed

### Required fields

- `children`: array, length 1 to 12 (the upper bound is the family-size cap; an empty array is a 400)
- `children[].email`: valid email, max 255 chars, unique
- `children[].password`: 8+ chars, at least one upper/lower/digit/special character
- `children[].name.first`, `children[].name.last`: 1-100 chars, regex `^[A-Za-z' -]+$`
- `children[].dob`: ISO-8601 date in the past
- `children[].grade`: a value from the `grade` enum
- `children[].activationCode`: non-empty string; resolved to a group via `invitation_codes`

### Optional fields

- `children[].name.middle`: same regex as first/last
- All demographics fields (`gender`, `statusEll`, `statusIep`, `statusFrl`, `race`, `hispanicEthnicity`, `homeLanguage`): optional, validated against the same schemas used by `POST /v1/users`

## Response Specification

### Success Response

**Status**: `201 Created`
**Body type**: `ApiResponse<{ ids: string[] }>`

```ts
{
  data: {
    ids: string[];  // userIds of newly created children, in request order
  }
}
```

The original ticket's `{ success: true }` payload is replaced with the new child ids — consistent with the rest of the v1 POST contract, useful to the frontend which otherwise has to issue a follow-up read to discover the new userIds.

### Error Responses

| Status Code | Reason |
| --- | --- |
| 400 | Request body fails Zod validation (missing required fields, invalid email, password too short, dob in future, grade not in enum, empty children array, more than 12 children in one request) |
| 401 | Unauthenticated (no JWT or invalid token) |
| 403 | Caller is not a parent of this family and is not a super admin |
| 404 | The `:familyId` does not exist |
| 409 | One or more child emails already exist (in `users` or Firebase Auth) |
| 422 | Validation that depends on resolving body-referenced data: invalid or expired activation code, family-size cap exceeded, family `rosteringEnded` is set |
| 429 | Firebase Auth rate-limited a `createUser` call |
| 500 | Internal server error (compensation paths run, no orphaned records) |

## Data Sources

### Firebase Auth

`FirebaseAuthClient.createUser` and `FirebaseAuthClient.getUserByEmail` from `apps/backend/src/clients/firebase-auth.clients.ts`.

### Tables (`core` database)

- `users`: one row per child (`userType='student'`)
- `user_families`: one row per child linking them to `:familyId` with `role='child'`
- `user_groups`: one row per (child, unique group) pair, with `role='student'` and `enrollmentStart=now()`
- `rostering_provider_ids`: one row per child, with `partnerId=familyId`
- `invitation_codes`: read-only — activation code lookup and validity check
- `groups`: read-only — confirming the resolved group is active (`rosteringEnded IS NULL`)
- `families`: read-only — confirming the family exists and is active

## Rules and Validation

### Family validation

- The `:familyId` URL parameter must be a valid UUID
- The family must exist in `families` and have `rosteringEnded IS NULL`
- The caller must be a `parent` of the family or a super admin (see Authorization)
- The size constraint: `count(user_families WHERE familyId=:familyId AND leftOn IS NULL) + len(children) ≤ 12`

### Activation code validation

- Each activation code must resolve to a row in `invitation_codes` (matched by `code`)
- The matched invitation code must satisfy `validFrom ≤ now()` and `(validTo IS NULL OR validTo > now())`
- The matched invitation code's `groupId` must reference a row in `groups` with `rosteringEnded IS NULL`
- Multiple children with the same code in one request resolve to the same group (deduplicated in service layer; no duplicate `user_groups` inserts per child)
- The Firebase-era "if `dateCreated` is older than 2 months, code is considered expired" heuristic is dropped — the new schema uses explicit `validFrom`/`validTo` columns

### Child validation

- Emails must be unique across all existing `users` rows (case-insensitive) and across all Firebase Auth accounts
- Emails must be unique within the request itself (a 400 if two children in the same request share an email)
- Password requirements identical to `POST /v1/users` and `POST /v1/families`
- `name.first` / `name.last` non-empty after trim, allowed character set `[A-Za-z' -]`
- `dob` must be a valid past date (matches the `users_dob_in_past` check constraint)

### Rostering provider IDs

One row per child:

- `providerType = 'dashboard'`
- `providerId = childUserId`
- `partnerId = familyId`
- `entityType = 'user'`
- `entityId = childUserId`

This matches the pattern used by `POST /v1/users` and `POST /v1/families` for users belonging to a family.

## Atomicity model

The endpoint is all-or-nothing across the entire children array:

- Step 6 (Firebase): if creating child N fails, delete UIDs for children 1..N-1 and abort
- Step 7 (DB transaction): wraps every insert for every child. If any insert fails, the whole transaction rolls back; delete every Firebase UID created in step 6.
- Step 9 (FGA): tuples written after the transaction commits. If FGA fails, attempt tuple delete, then DB row delete in a rollback transaction (rostering_provider_ids first, then user_groups, then user_families, then users), then Firebase delete.

Per-child partial success is not supported. A request to add three children either creates all three or creates none.

## Acceptance criteria

### Happy path

- [ ] Authenticated parent of `:familyId` adds N children (1 ≤ N ≤ 12 minus current family size)
- [ ] N Firebase Auth accounts are created
- [ ] N `users` rows are created with `userType='student'`, `authProvider=['password']`
- [ ] N `user_families` rows are created with `role='child'`
- [ ] M `user_groups` rows are created (M = N × unique activation codes in the request), each with `role='student'`
- [ ] N `rostering_provider_ids` rows are created with `partnerId=familyId`
- [ ] FGA tuples are written for child→family and child→group memberships
- [ ] Returns 201 with `{ data: { ids: [...] } }`

### Authorization

- [ ] Returns 401 if no `Authorization` header
- [ ] Returns 403 if the caller is authenticated but not a parent of `:familyId` and not a super admin
- [ ] Super admin can add children to any family

### Validation

- [ ] Returns 400 for empty `children` array, missing required fields, invalid email, password too short or missing required class, dob in the future, invalid grade, more than 12 children in one request
- [ ] Returns 404 if `:familyId` does not exist
- [ ] Returns 409 if any child email already exists (`users` or Firebase Auth)
- [ ] Returns 422 if an activation code is invalid or expired
- [ ] Returns 422 if the family has `rosteringEnded` set
- [ ] Returns 422 if adding the requested children would exceed the family-size cap
- [ ] Returns 429 if Firebase rate-limits

### Atomicity

- [ ] Firebase failure on child K leaves zero Firebase accounts and zero DB rows
- [ ] DB transaction failure leaves zero DB rows and zero Firebase accounts (compensation deletes them)
- [ ] FGA failure rolls back DB rows and Firebase accounts
- [ ] Multiple children with the same activation code share a single group enrollment row (one `user_groups` row per child-group pair, not per child-code pair)

### Tests

- [ ] Unit tests cover the service-layer authorization branches (super admin, parent of family, non-parent, unauthenticated), activation-code validation (valid, expired, not found, multiple codes deduplicated), family-size cap, atomicity (DB failure compensation, FGA failure compensation)
- [ ] Integration tests cover the full HTTP stack: end-to-end success with N children, each error branch, transaction rollback under simulated DB failure, Firebase cleanup verification
- [ ] Route integration tests cover the auth contract: 401 unauthenticated, 403 wrong caretaker, 200/201 super admin

## Out of scope

- `caretakerEmail`, `caretakerData`, `consentData` in the request body (dropped — see Request Specifications)
- Parent profile updates via `caretakerData` (use `PATCH /v1/users/:id`)
- Consent collection (separate domain, will land in its own endpoint)
- Frontend changes to the existing add-students UI (separate PR)

## Future hardening

- **Per-IP rate limiting** on this endpoint: tighter than the global WAF since it's a one-shot bot vector once a family exists. Same approach as the create-family endpoint.
- **Activation-code reuse audit**: today a code's `validTo` is the only expiry signal. A future hardening pass could add a per-code use cap (e.g. 1 family per code) to slow code-leak abuse.

richford:
@claude please re-review

richford:
@claude, please re-review

packages/api-contract/src/v1/families/schema.ts
```diff
@@ -101,3 +106,97 @@ export const CreateFamilyResponseSchema = z.object({
...
+ */
+const AddChildDemographicsSchema = z.object({
+  gender: z.string().nullable().optional(),
+  race: z.string().nullable().optional(),
+  statusEll: z.string().nullable().optional(),
+  statusFrl: FreeReducedLunchStatusSchema.nullable().optional(),
+  statusIep: z.string().nullable().optional(),
+  hispanicEthnicity: z.boolean().nullable().optional(),
+  homeLanguage: z.string().nullable().optional(),
+});
```
amyyeung17: I noticed `CreateUserDemographicSchema` in `users/schema.ts` is exactly the same. What are your thoughts on reusing it?

packages/api-contract/src/v1/families/schema.ts
```diff
@@ -101,3 +106,97 @@ export const CreateFamilyResponseSchema = z.object({
...
+ * Used both as an array-length cap on the request body (`children.length <= 12`)
+ * and as a runtime check at the service layer (`existing members + new children <= 12`).
+ */
+export const FAMILY_SIZE_LIMIT = 12;
+
+/**
+ * Name schema for a child being added to a family. Same regex as the caretaker
+ * (and POST /users) for consistency across signup paths.
+ */
+const AddChildNameSchema = z.object({
```
amyyeung17: See comment about `AddChildDemographicSchema`. The respective schema is `CreateUserNameSchema`. I already approved the base PR, but it could also apply to ` CreateFamilyCaretakerNameSchema`.

apps/backend/src/repositories/invitation-code.repository.ts
```diff
@@ -18,6 +18,39 @@ export class InvitationCodeRepository extends BaseRepository<InvitationCode, typ
     super(db, invitationCodes);
   }
 
+  /**
+   * Look up an invitation code by its `code` value, only returning rows that are currently valid.
+   *
+   * Filters by:
+   * - `code` matches exactly (case-sensitive — codes are opaque)
+   * - `validFrom <= NOW()`
+   * - `validTo >= NOW() OR validTo IS NULL`
```
amyyeung17: While this aligns with `getLatestValidByGroupId`, it contradicts the ticket: `(validFrom ≤ now() and (validTo IS NULL OR validTo > now()))`. Can we confirm which behavior is intended?

richford:
@claude, please re-review

amyyeung17 (dismissed):

richford:
@claude please re-review after the merge commit that resolved conflicts after the base branch changed.

kellyel (approved):

amyyeung17 (approved):

PR #1791, opened 2026-05-19
`ref` Create Shared Assessment Schema in Monorepo

ksmontville:
## Proposed changes

This PR introduces the `/assessment-schema` package which defines shared interfaces between assessments and the backend, particularly when reporting scores and subscores for reporting endpoints that feed into the frontend.

### Monorepo migration of ROAR-PA
 
  - apps/assessments/roar-pa/ — PA assessment migrated into the monorepo as a workspace package
  (@roar-dashboard/roar-pa), with its own build pipeline and Turborepo config
  - Assessment SDK firekit compat — roar-pa/src/index.js now uses startRun/finishRun/abortRun from
  @yeatmanlab/assessment-sdk/compat/firekit instead of the old Firekit directly; RoarPA.run() now correctly awaits
  jsPsych.run(timeline)
  - TaskPA.vue — new backend integration — calls GET /tasks/pa, GET /me, and GET /users/:id/administrations?embed=tasks
  on the new REST API to resolve the correct variantId and administrationId before calling initFirekitCompat()
  - Anonymous user endpoint (POST /v1/users/anonymous) — added body: c.noBody() to the contract and an integration test
  asserting no FGA tuples are written for anonymous users
  - Local dev infrastructure — Vite proxy (/v1 → https://localhost:4000) to avoid CORS, backend dev server runs with
  --use-system-ca for local TLS trust, and the CSP template dynamically includes the backend origin from
  VITE_ROAR_API_BASE_URL

### Completion semantics
RoarPA.run() now resolves by awaiting jsPsych.run(timeline) directly rather than polling  firekit.run.completed. This preserves the contract that dashboard callers (TaskPA.vue, PlayApp.vue, etc.) depend on — completeAssessment() and navigation still happen after the experiment ends. finishRun() is now fire-and-forget, but completeAssessment() operates on Firestore assignment state that doesn't depend on the run write having committed.

### Holding Pattern
The frontend integration is functional end-to-end but blocked on data alignment: the dashboard still resolves administrations from Firestore using document IDs, while the new backend uses Postgres UUIDs. Until the dashboard's administration queries migrate to the new REST API, the two sources won't share a common key to match on.

### Need Staging env-configs
Need to update `env-configs` repo with environment variable:
`VITE_ROAR_API_BASE_URL = https://roar-staging-api-884871331266.us-central1.run.app/v1`

## Types of changes

- [ ] Documentation Update
- [x] Repository Maintenance

## Checklist


ksmontville:
@claude Let's review this PR. We want to keep the review focus on how the new `assessment-schema` package interacts with the assessments and the backend. We want to also evaluate the usage of the `assessment-sdk` as a replacement for the deprecated Firekit as used within the `roar-pa` source code. 

We do not want to review the `roar-pa` source code itself, only how it now interacts with the rest of the monorepo. 

We also want to ensure that our new infrastucture for creating an anonymous user is sound, and then suggest some unit and integration test cases.

ksmontville:
@claude Feedback implemented, let's review again.

ksmontville:
@claude Feedback implemented, let's review again.

packages/assessment-schema/src/pa/config.ts
```diff
@@ -0,0 +1,26 @@
+export const PA_TASK_ID = "pa" as const;
+
+export const PA_DEV_PORT = 8000 as const;
+
+export const PA_SCORING_VERSION = {
+  FIXED: 3,
+  ADAPTIVE: 4,
+} as const;
```
amyyeung17: We’ll most likely have different scoring versions that are also adaptive. Just curious how we’d name those keys in the future. For example, swr has adaptive version 6 and 7.

ksmontville:
@claude Let's make sure that our rebase onto ref/1794/monorepo-migrate didn't introduce any regressions. We have intentionally removed all assessment deployment and versioning details, which will be addressed in a PR up the chain.

apps/backend/src/app.ts
```diff
@@ -8,6 +8,32 @@ import { ApiErrorCode } from './enums/api-error-code.enum';
...
+  }
+
+  if (req.method === 'OPTIONS') {
+    res.sendStatus(StatusCodes.NO_CONTENT);
+    return;
+  }
+
+  next();
+});
+
```
richford: The next PR in the chain (#1794) removes this entirely. Ideally, we wouldn't add this just to remove it later. Can we drop it from this PR and let #1794 introduce the proxy approach?

apps/assessments/roar-pa/serve/serve.js
```diff
@@ -40,61 +40,98 @@ const isAdaptive = urlParams.get('isAdaptive') === 'true';
...
+      if (!resolvedVariantId) {
+        // eslint-disable-next-line no-undef
+        const variantRes = await fetch(`${ROAR_API_URL}/v1/tasks/roar-pa/variants?perPage=1`, {
+          headers: { Authorization: `Bearer ${token}` },
+        });
+        const variantJson = await variantRes.json();
+        if (!variantRes.ok) {
+          console.error('Failed to fetch roar-pa task variants:', variantRes.status, variantJson);
+          return;
+        }
```
richford, line 78: The `token` here is the anonymous Firebase token from a few lines up. This fetch only works because `taskVariants.list`'s authorization is currently relaxed enough for anonymous callers. But I don't see a backend integration test confirming an anonymous bearer can actually reach `GET /v1/tasks/:taskId/variants` and that the response includes the variant id. Can you please add one?

Also, can you please file a tracked issue for the `TODO: Replace with a proper "default variant" concept` comment two lines above? The link points at the umbrella migration issue, not the variant-default work itself.

apps/assessments/roar-pa/src/index.js
```diff
@@ -1,37 +1,33 @@
...
     return buildExperiment(config);
   }
 
   async run() {
     const { jsPsych, timeline } = await this.init();
     this.jsPsych = jsPsych;
     this.jsPsych.run(timeline);
-
-    await waitFor(() => this.firekit.run.completed === true);
   }
```
richford, line 30: I get that you have to get rid of the firekit stuff and remove the `await waitFor(() => this.firekit.run.completed === true)` synchronization. But the new compat shim's `finishRun()` is fire-and-forget (it gets called from `config.js:on_finish` with `.catch(err => console.error(...))`), so `RoarPA.run()` now returns before the run reaches the backend. So if `apps/dashboard/src/components/tasks/TaskPA.vue` is awaiting `roarApp.run()` for completion semantics, it will now race the network write.

Can you check:

- Spot-check `TaskPA.vue` (and any other `roarApp.run()` caller) to verify they no longer depend on completion happening before the promise resolves.
- If completion semantics are genuinely no longer needed, call that out in the PR description.

apps/assessments/roar-pa/.env.example
```diff
@@ -10,3 +10,6 @@ FIREBASE_APP_STORAGE_BUCKET=
 FIREBASE_APP_MESSAGING_SENDER_ID=
 FIREBASE_APP_APP_ID=
 FIREBASE_APP_RECAPTCHA_SITE_KEY=
+
+# ROAR backend server root URL. The /v1 version prefix is appended automatically.
```
richford, line 14: This will only run locally if the user has already run `dev:setup:certs` to install local TLS certificates. I know that #1794 enforces that prerequisite but until that lands, let's add this comment.

```suggestion
# ROAR backend server root URL. The /v1 version prefix is appended automatically.
# Requires `dev:setup:certs` to be run first
```

packages/api-contract/src/v1/users/contract.ts
```diff
@@ -180,6 +181,28 @@ export const UsersContract = c.router(
     },
     // Nest guardian / longitudinal score report sub-router under /users
     scoreReports: GuardianStudentReportContract,
+    createAnonymous: {
+      method: 'POST',
+      path: '/anonymous',
+      body: z.object({}),
```
richford: The controller pulls nothing from the body, and `serve.js` posts `JSON.stringify({})`. As written, ts-rest will accept any extra fields silently. Can you either constrain with `.strict()` so unknown fields 400 explicitly, or drop the body schema entirely (and the matching `JSON.stringify({})` in `serve.js`) so the request shape is `POST /v1/users/anonymous` with no body?

apps/backend/src/middleware/anon-token/anon-token.middleware.ts
```diff
@@ -0,0 +1,89 @@
...
+        new ApiError(ApiErrorMessage.UNAUTHORIZED, {
+          statusCode: StatusCodes.UNAUTHORIZED,
+          code: ApiErrorCode.AUTH_REQUIRED,
+          context: { uid: decodedUser.uid },
+        }),
+      );
+    }
+
+    req.decodedAnonymousUser = decodedUser;
+    return next();
```
richford: The Firebase-claims narrowing happens twice: once inside `isAnonymousToken` (returns boolean), and a few lines later when pulling `signInProvider` out of `firebase.sign_in_provider` for the log line. Can you extract a single helper that returns something like `{ ok: boolean; signInProvider?: string }` so both branches share one type?

apps/backend/src/services/user/user.service.ts
```diff
@@ -1435,5 +1435,41 @@ export function UserService({
...
+   * consistent with how anonymous runs are handled in `run.service.ts` — the
+   * `isAnonymous` flag bypasses `can_create_run` checks there for the same reason.
+   * Authentication is enforced upstream by `AnonTokenMiddleware`, which verifies
+   * the Firebase anonymous token before this method is ever reached.
+   *
+   * @param authId - The Firebase UID of the anonymous user
+   * @returns The newly created user's ROAR UUID
+   * @throws {ApiError} If the database insert fails
+   */
+  async function createAnonymousUser(authId: string): Promise<{ id: string }> {
```
richford, line 1455: The JSDoc here correctly notes that `createAnonymousUser` skips FGA tuple seeding because anonymous users have no memberships. That's the right design, but the integration tests only asserts on `data.id` shape and idempotency.

Can you please add one assertion to the existing integration test? After `createAnonymousUser` returns, query the FGA store for tuples mentioning the new `authId` and assert the list is empty.

packages/assessment-schema/src/pa/config.ts
richford: Re-raising the taxonomy concern here. `FIXED` and `ADAPTIVE` don't scale well. As @amyyeung17 mentioned, SWR alone has versions 6 and 7, both adaptive.

Possible solutions:
- Rename to version-anchored keys: `V3_FIXED: 3`, `V4_ADAPTIVE: 4`. Same shape, scales to `V6_ADAPTIVE`, `V7_ADAPTIVE` for SWR, and so on.
- Or restructure to a record keyed by version number: `{ 3: { kind: 'fixed' }, 4: { kind: 'adaptive' } }`.

richford (changes requested):

apps/backend/src/services/scoring/scoring.config-schema.ts
```diff
@@ -142,6 +142,7 @@ const SubscoreFieldEntrySchema = z.object({
   correctName: z.string(),
   attemptedName: z.string(),
   percentCorrectName: z.string().optional(),
+  label: z.string().optional(),
```
richford, line 145: Adding `label` isn't validated against the existing JSON configs (`swr.json`, `sre.json`, etc.) that still don't carry labels. I won't make this blocking but I do want to ask two questions:

- Is `label` going to become the standard once other assessments migrate to TS configs? If yes, document that intent in a comment above the schema so future authors know labels aren't truly optional, but rather "optional for legacy configs only."
- If `label` is PA-specific, the field probably belongs on a PA-specific extension of the schema rather than the base. Keeping it on the base type means every other config has to think about whether to set it.

apps/assessments/roar-pa/src/index.js
richford:
```js
  document.querySelectorAll('audio').forEach((el) => el.pause());
  if (this.jsPsych) {
    this.jsPsych.endExperiment();
  }
  // Order so that UI teardown is synchronous; backend abort is best-effort,
  // but log on failure so we know the server's run state may be stale.
  abortRun().catch((err) => {
    console.warn('abortRun failed; backend run state may be stale', err);
  });
```

richford (changes requested):

apps/backend/src/services/scoring/scoring.config-schema.ts
ksmontville: I don't have enough knowledge of score reports to answer this question. We can do whichever the team thinks is best.

richford (dismissed):

apps/assessments/roar-pa/serve/serve.js
maximilianoertel, line 1: Thanks for adding the comment explaining why the raw fetches were added, but assessments should never call the backend API directly. This is the SDK's responsibility and all requests to the backend API should exclusively go through that. 

If this was implemented as raw fetches here because of the need to have a working proof-of-concept ready quickly, can we please add a ticket to the Backend Refactor project to add a `bootstrapAnonymousSession()` function (or similar) to the Assessment SDK that accepts a partial context, provisions the anonymous user, optionally resolves a default variant, and returns a fully-formed CommandContext? This would eliminate both raw fetches and ensure every assessment that needs standalone/anonymous mode goes through the SDK rather than reimplementing this bootstrap logic. (cc @richford)

packages/api-contract/package.json
```diff
@@ -5,7 +5,9 @@
   "exports": {
     ".": {
       "types": "./dist/index.d.ts",
-      "import": "./dist/index.mjs"
+      "import": "./dist/index.mjs",
+      "require": "./dist/index.mjs",
+      "default": "./dist/index.mjs"
```
maximilianoertel: This likely won't work because `require` should point to a CJS file, yet it's pointing at an ESM file. I don't think we need to support CJS? Technically, the `default` is, I believe, also redundant here


```suggestion
```

packages/assessment-schema/src/pa/config.ts
```diff
@@ -0,0 +1,26 @@
+export const PA_TASK_ID = "pa" as const;
+
+export const PA_DEV_PORT = 8000 as const;
```
maximilianoertel: What is this constant used for? I can't see it used anywhere in the code


```suggestion
```

packages/assessment-sdk/package.json
```diff
@@ -13,11 +13,15 @@
...
+      "import": "./dist/index.mjs",
+      "require": "./dist/index.mjs",
+      "default": "./dist/index.mjs"
     },
     "./compat/firekit": {
       "types": "./dist/compat/firekit.d.ts",
-      "import": "./dist/compat/firekit.mjs"
+      "import": "./dist/compat/firekit.mjs",
+      "require": "./dist/compat/firekit.mjs",
+      "default": "./dist/compat/firekit.mjs"
```
maximilianoertel, line 22: Same as https://github.com/yeatmanlab/roar-dashboard/pull/1791/changes#r3322667592

apps/dashboard/vite.config.js
maximilianoertel, line 1: The dev proxy and CSP connect-src injection are workarounds for the backend not having CORS configured yet. The backend currently has no CORS middleware and returns no Access-Control-* headers. 

The proxy also hardcodes https://localhost:4000 but if a dev runs the backend on a different port by specifying `PORT`, this will fail.

The proper fix is to add CORS to the backend with an allowlist of dashboard origins. That would eliminate both of these workarounds: the dashboard can call the API directly at its real URL without a proxy, and the CSP connect-src hack becomes unnecessary.

I will work on https://github.com/yeatmanlab/roar-project-management/issues/1835 shortly and open a PR so that the below workaround can be removed

maximilianoertel (changes requested):

packages/api-contract/package.json
ksmontville: We can get rid of the "require" key, but without the "default" key we keep getting lint errors like this:


Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in /home/runner/work/roar-dashboard/roar-dashboard/node_modules/@roar-dashboard/api-contract/package.json
Occurred while linting /home/runner/work/roar-dashboard/roar-dashboard/apps/dashboard/src/components/tasks/TaskPA.vue:9


I don't know the underlying cause of this or if it's worth addressing in this PR.

ksmontville:
@claude Let's check for regressions after a considerable large rebase onto project/backend-refactor branch.

apps/assessments/roar-pa/serve/serve.js
richford: Agreed. Thanks for catching @maximilianoertel

apps/dashboard/src/components/tasks/TaskPA.vue
```diff
@@ -117,7 +119,76 @@ async function startTask(selectedAdmin) {
...
+    // TODO: resolve the participant's Postgres UUID for the proxy-launch path (launchId set).
+    //
+    // NOTE: Until the dashboard migrates its administration queries to the new REST API,
+    // selectedAdmin.value.id is a Firestore document ID and will not match any administration
+    // in the new backend. The fallback (matching by PA task UUID within embedded tasks)
+    // is used until that migration is complete.
+    const roarApiClient = getRoarApiClient();
+
+    const [taskRes, meRes] = await Promise.all([
+      roarApiClient.tasks.get({ params: { taskId: 'pa' } }),
```
richford: Shouldn't we use `PA_TASK_ID` here as well from the `@roar-dashboard/assessment-schema` package?

apps/dashboard/src/components/tasks/TaskPA.vue
```diff
@@ -117,7 +119,76 @@ async function startTask(selectedAdmin) {
...
+    if (adminsRes.status !== 200) {
+      throw new Error(`Failed to fetch administrations from the ROAR backend (status ${adminsRes.status}).`);
+    }
+
+    const paTaskUuid = taskRes.body.data.id;
+    const backendAdmins = adminsRes.body.data.items;
+
+    const matchedAdmin =
+      backendAdmins.find((a) => a.id === selectedAdmin.value.id) ??
+      backendAdmins.find((a) => (a.tasks ?? []).some((t) => t.taskId === paTaskUuid));
```
richford, line 171: This fallback pattern is a holding pattern artifact, right? It's already documented above in the `NOTE` comment. But can you add a TODO here to remove this fallback?

apps/backend/src/services/scoring/scoring.config-schema.ts
richford: I'm fine with leaving this as is and deferring this decision to when we migrate future assessments into the monorepo.

packages/assessment-schema/src/pa/subscores.test.ts
richford, line 1: This test (plus the `satisfies Record<PaSubtaskKey, PaSubscoreDef>` typing) covers the schema's *internal* consistency, which is good, and sufficient for this PR. It can't catch producer <-> schema drift, though: `scores.js` is a separate package that doesn't import the schema, and the score-persistence path doesn't exist yet. The SDK's `writeTrial` does `void computedScoreCallback;` ("not yet implemented"), so there's no live producer to assert against here. Binding the producer to the schema, so an unregistered `run_scores.name` (like #1808's `composite_foundational`) can't be emitted. This is tracked separately in [issue 1838](https://github.com/yeatmanlab/roar-project-management/issues/1838), which has the full plan (a `computedScores → ScoreEntry[]` mapping sourced from `PA_SCORE_NAMES`/`PA_SUBSCORE_DEFS`, consumed by the SDK once it implements score-writing). No action needed here beyond keeping the schema as the canonical source of names. I just thought I'd flag it here. No action needed.

richford (changes requested):

apps/dashboard/src/components/tasks/TaskPA.vue
ksmontville: We can do this:

```ts
const props = defineProps({
  taskId: { type: String, default: PA_TASK_ID },
  language: { type: String, default: 'en' },
  launchId: { type: String, default: null },
});
```

```ts
    const [taskRes, meRes] = await Promise.all([
      roarApiClient.tasks.get({ params: { taskId: props.taskId } }),
      roarApiClient.me.get(),
    ]);
```

apps/dashboard/src/components/tasks/TaskPA.vue
ksmontville: Issue created and noted.

packages/assessment-schema/src/pa/subscores.test.ts
ksmontville: Noted.

richford (approved):

PR #1794, opened 2026-05-19
`ref` Introduce Assessment Development Environment

ksmontville:
## Summary

This PR introduces an assessment development environment for ROAR Phonological Awareness (roar-pa) that spins up a fully local, self-contained stack — Postgres, backend API, Firebase Auth emulator, and assessment dev server — with a single command. It also wires up CI to run assessment E2E tests against that same stack when assessment code changes are detected.

### Docker environment (`docker-compose.assessment.yml` + `assessment.Dockerfile`)

Four services, started in dependency order:

- **`assessment-db`** — Postgres 17 container that creates `roar_core` and `roar_assessment` databases on first run. Data persists to a named Docker volume (`pgdata`) across restarts.
- **`assessment-db-migrate`** — One-shot container that runs: FDW setup → all Drizzle migrations (core + assessment) → PA task seed. Exits 0 on success; the backend waits for it before starting.
- **`firebase-auth-emulator`** — Firebase Auth emulator on port 9099. Pure Node.js (no Java required). The backend and assessment connect to it via `FIREBASE_AUTH_EMULATOR_HOST`.
- **`backend`** — ROAR API server on port 4000, running in production mode (HTTP, no TLS). FGA is configured with placeholder values — anonymous assessment runs never trigger FGA calls.

`assessment.Dockerfile` is a dual-target image: `assessment-runtime` serves as both the migration runner and the API server; `firebase-emulator` runs the Auth emulator.

### PA seed script (`apps/backend/seeds/pa.seed.ts`)

An idempotent seed script that inserts the roar-pa task and two default variants:

| Variant          | Language | Mode              |
|------------------|----------|-------------------|
| English Fixed    | en       | Fixed-form        |
| English Adaptive | en       | IRT/adaptive (mfi)|

Safe to re-run — existing records are left unchanged.

### Assessment environment scripts

**`scripts/assessment-env-up.sh`** — called by `npm run assessment-environment:up`:

1. Checks that Docker Compose v2 is available
2. Checks that port 5432 is free (gives a clear error if a local Postgres is running)
3. Removes any orphaned containers from a previous run
4. Runs `docker compose up -d --wait` — starts all four services and waits until every health check passes
5. Starts the webpack assessment dev server on port 8000 with `FIREBASE_AUTH_EMULATOR_HOST` and `BACKEND_URL` injected

**Ctrl+C** stops only the webpack dev server. Docker services keep running with data intact. Run `assessment-environment:up` again to reattach.

**`npm run assessment-environment:down`** — stops all Docker services and removes the `pgdata` volume (permanently deletes all data).

### CI integration (`.github/workflows/ci.yml`)

A new `e2e-assessments` job runs Cypress tests for any assessment that has changed (detected by `detect-assessment-changes`). It uses the same `docker-compose.assessment.yml` to bring up the full stack, then starts the assessment webpack server via a new `dev:ci` npm script (identical to the local start, minus `--open`).

## Usage

From `apps/assessments/roar-pa`:

```bash
npm run assessment-environment:up
```

Wait for "All services healthy" — first run takes a few minutes while Docker builds the images and runs migrations. Subsequent runs are fast (layer cache).

| Service | URL |
|---------|-----|
| Assessment app | http://localhost:8000 |
| Backend API | http://localhost:4000 |
| Firebase emulator | http://localhost:9099 |
| Postgres | localhost:5432 |

Query data with any Postgres client:

```bash
pgweb --url "postgres://postgres@localhost:5432/roar_core?sslmode=disable"
# or
psql "postgres://postgres@localhost:5432/roar_assessment?sslmode=disable"
```

## Tear down

```bash
npm run assessment-environment:down
```

## Data model

All runs are anonymous. The guest user created during a session has no org, class, group, or family memberships — this mirrors real-world standalone mode where assessments run outside the ROAR dashboard. All data is wiped on `assessment-environment:down`.

## Types of changes

- [x] Documentation Update
- [ ] Repository Maintenance

## Checklist


ksmontville:
@claude Let's review this PR, focusing on the setup and usage of the researcher development environment enabled with `npm run researcher-environment:up`. We should look for ways to improve the Docker compose set up, reduce redundancies, and simplify the npm scripts if we can since currently they are quite verbose.

ksmontville:
@claude Fedback implemented, let's check again.

The researcher-env:down script was never intended to be in the root package.json, so we removed it. Those scripts are all scoped to the assessment.

ksmontville:
@claude Feedback implemented, let's review.

ksmontville:
@claude Feedback implement, let's review again.

The dev:pa script was never supposed to be accessible at the monorepo root (likely a leftover during initial development). All assessment scripts are scoped to their directory by choice; users are expected to CD into that assessment and go from there.

The race condition is a non-issue and will not be addressed.

ksmontville:
@claude We have removed the `optimization-js` package as it seems largely unused. Let's do another review.

ksmontville:
@claude Let's make sure that our rebase onto ref/1794/monorepo-assessment-schema didn't introduce any regressions. We have intentionally removed all assessment deployment and versioning details, which will be addressed in a PR up the chain. We have also adjusted how we handle CORS when the ephemeral database attempts to write to it, now preferring a Webpack proxy.

ksmontville:
@claude Feedback addressed, let's re-review.

ksmontville:
@claude Feedback addressed, let's done another review.

.github/workflows/deploy-platform-development.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       image-version: ${{ steps.docker-build.outputs.image-version }}
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@v4
```
richford: Is this an accidental downgrade or do we really need `v4` for some reason? If possible, can you please revert to `v6`?

.github/workflows/deploy-platform-development.yml
```diff
@@ -58,7 +58,7 @@ jobs:
     timeout-minutes: 10
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@v4
```
richford: Same here about the action version downgrade.

.github/workflows/deploy-platform-development.yml
```diff
@@ -87,7 +87,7 @@ jobs:
...
```
richford: Same here about the action version downgrade.

.github/workflows/deploy-platform-development.yml
```diff
@@ -116,7 +116,7 @@ jobs:
       DOTENV_PRIVATE_KEY_DEVELOPMENT: ${{ secrets.DOTENV_PRIVATE_KEY }}
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@v4
```
richford: Same here about the action version downgrade.

.github/workflows/deploy-platform-production.yml
```diff
@@ -35,7 +35,7 @@ jobs:
     steps:
       - name: Checkout repository
         if: github.event_name == 'workflow_call'
-        uses: actions/checkout@v6
+        uses: actions/checkout@v4
```
richford: Same here about the action version downgrade.

.github/workflows/deploy-platform-production.yml
```diff
@@ -96,7 +96,7 @@ jobs:
       image-version: ${{ steps.docker-build.outputs.image-version }}
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@v4
```
richford: Same here about the action version downgrade.

.github/workflows/deploy-platform-production.yml
```diff
@@ -123,7 +123,7 @@ jobs:
     timeout-minutes: 10
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@v4
```
richford: Same here about the action version downgrade.

.github/workflows/deploy-platform-production.yml
```diff
@@ -152,7 +152,7 @@ jobs:
...
```
richford: Same here about the action version downgrade.

.github/workflows/deploy-platform-production.yml
```diff
@@ -181,7 +181,7 @@ jobs:
       DOTENV_PRIVATE_KEY_PRODUCTION: ${{ secrets.DOTENV_PRIVATE_KEY }}
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@v4
```
richford: Same here about the action version downgrade.

.github/workflows/deploy-platform-staging.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       image-version: ${{ steps.docker-build.outputs.image-version }}
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@v4
```
richford: Same here about the action version downgrade.

.github/workflows/deploy-platform-staging.yml
```diff
@@ -57,7 +57,7 @@ jobs:
     timeout-minutes: 10
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@v4
```
richford: Same here about the action version downgrade.

.github/workflows/deploy-platform-staging.yml
```diff
@@ -86,7 +86,7 @@ jobs:
...
```
richford: Same here about the action version downgrade.

.github/workflows/deploy-platform-staging.yml
```diff
@@ -115,7 +115,7 @@ jobs:
       DOTENV_PRIVATE_KEY_STAGING: ${{ secrets.DOTENV_PRIVATE_KEY }}
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@v4
```
richford: Same here about the action version downgrade.

scripts/researcher-env-up.sh
```diff
@@ -0,0 +1,52 @@
...
+  --prefix-colors cyan,yellow,green \
+  "npx firebase emulators:start --only auth --project demo-roar --config $FIREBASE_CONFIG" \
+  "cd $REPO_ROOT/apps/backend && NODE_ENV=development \
+   GOOGLE_CLOUD_PROJECT=demo-roar \
+   CORE_DATABASE_URL=postgresql://postgres@localhost:5432/roar_core \
+   ASSESSMENT_DATABASE_URL=postgresql://postgres@localhost:5432/roar_assessment \
+   FIREBASE_AUTH_EMULATOR_HOST=127.0.0.1:9099 \
+   npm run dev" \
+  "docker compose -f $COMPOSE_FILE run --rm -T researcher-db-migrate \
+   && cd $REPO_ROOT/apps/assessments/roar-pa && FIREBASE_AUTH_EMULATOR_HOST=127.0.0.1:9099 npm run dev"
```
richford: With
```bash
npx concurrently \
  --kill-others-on-fail \
  ...
  "docker compose -f $COMPOSE_FILE run --rm -T researcher-db-migrate \
   && cd $REPO_ROOT/apps/assessments/roar-pa && FIREBASE_AUTH_EMULATOR_HOST=127.0.0.1:9099 npm run dev"
```
The command only waits for `researcher-db-migrate` to exit before running `npm run dev`, but PA's first network call goes to the `backend` command's HTTPS server and the migrate container's completion says nothing about whether the backend has bound port 4000 yet. If a researcher's machine is slow enough that backend hasn't finished booting when PA tries to fetch the first variant, then `--kill-others-on-fail` tears down the whole stack on `ECONNREFUSED`.

Can you please insert a `wait-on` between the migrate and the dev server:

```bash
"docker compose -f $COMPOSE_FILE run --rm -T researcher-db-migrate \
 && npx --yes wait-on https://localhost:4000/health --timeout 60000 \
 && cd $REPO_ROOT/apps/assessments/roar-pa && FIREBASE_AUTH_EMULATOR_HOST=127.0.0.1:9099 npm run dev"
```

(`wait-on` should accept the self-signed cert because backend HTTPS uses the same `roar-local.crt` you're checking exists at line 19; if it doesn't, add `--httpStatusOk 200`.)

migrate.Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+COPY packages/assessment-schema/package.json      packages/assessment-schema/
+COPY packages/assessment-sdk/package.json         packages/assessment-sdk/
+COPY packages/authz/package.json                  packages/authz/
+COPY packages/config-eslint/package.json          packages/config-eslint/
+COPY packages/config-prettier/package.json        packages/config-prettier/
+COPY packages/config-typescript/package.json      packages/config-typescript/
+
+# Install dependencies. --ignore-scripts skips native addon compilation (e.g. canvas)
+# which fails on Node 24's ABI and is not needed for migrations.
+RUN npm ci --ignore-scripts
```
richford: The comment cites "Node 24's ABI" but the base on line 1 is `FROM node:22`.

- If we actually want Node 24 can you switch line 1 to `FROM node:24`?
- If we actually want Node 22, can you update the comment to drop the version-specific claim and just say we don't need native addon compilation for migrations?

apps/backend/src/clients/firebase-core.client.ts
```diff
@@ -52,6 +52,16 @@ export class FirebaseCoreClient {
       return this.appInstance;
     }
 
+    // When the Auth emulator is active, the Admin SDK routes token verification to localhost
+    // and does not validate credentials. applicationDefault() returns a lazy credential object
+    // that defers ADC lookup until getAccessToken() is called — which the emulator never needs.
+    // This allows initialization to succeed even without gcloud credentials (e.g., under sudo).
+    if (process.env.FIREBASE_AUTH_EMULATOR_HOST) {
+      const projectId = process.env.GCLOUD_PROJECT ?? process.env.GOOGLE_CLOUD_PROJECT ?? 'demo-roar';
+      this.appInstance = initializeApp({ credential: applicationDefault(), projectId });
```
richford: Can you drop `credential: applicationDefault()`? For emulator-only use [the docs](https://firebase.google.com/docs/emulator-suite/connect_auth#admin_sdks) recommend `initializeApp({ projectId })` with no credential. `applicationDefault()` is a lazy credential, but it still creates the credential object and registers it on the App. If we drop the ADC, then I think it's cleaner and we don't have any risk of ADC drift if a future Firebase SDK version changes when the lazy resolution fires.

Does that work, or was there some reason you needed to use the ADC here instead of the demo project ID?

apps/assessments/shared/firebaseConfig.js
```diff
@@ -1,27 +1,19 @@
...
-      authDomain: process.env.FIREBASE_APP_AUTH_DOMAIN,
-      projectId: process.env.FIREBASE_APP_PROJECT_ID,
-      storageBucket: process.env.FIREBASE_APP_STORAGE_BUCKET,
-      messagingSenderId: process.env.FIREBASE_APP_MESSAGING_SENDER_ID,
-      appId: process.env.FIREBASE_APP_APP_ID,
-    };
+  if (process.env.FIREBASE_AUTH_EMULATOR_HOST) {
+    // The client SDK requires apiKey to be non-empty before calling getAuth(), but the
+    // emulator never validates it. A placeholder satisfies the check without real credentials.
+    return { projectId: 'demo-roar', apiKey: 'emulator-api-key' };
```
richford: The string `'demo-roar'` is hard-coded in at least four places:

- here on this line
- `scripts/researcher-env-up.sh:45`
- `scripts/researcher-env-up.sh:47`
- `apps/backend/src/clients/firebase-core.client.ts:60`

Can you central this constant? Maybe in a `RESEARCHER_LOCAL_FIREBASE_PROJECT_ID` (or some such) constant in `@roar-dashboard/assessment-schema` (alongside `PA_TASK_ID`).

apps/dashboard/vite.config.js
richford: Between this line and `apps/assessments/roar-pa/webpack.config.cjs:145-152`, both proxies hard-code the target: 'https://localhost:4000'. The backend reads `BACKEND_PORT` and `PORT` from turbo.json's `globalPassThroughEnv`, so a researcher who customizes the port will have the dashboard and assessment silently miss the proxy.

Can we hoist the URL into a `.env.example`-documented variable and read it in both configs. E.g.,

```js
// vite.config.js
const backendUrl = process.env.BACKEND_URL ?? 'https://localhost:4000';
// ...
proxy: { '/v1': { target: backendUrl, secure: false, changeOrigin: true } },
```

And the same in webpack.config.cjs. And then also document `BACKEND_URL` in the `.env.example` files. I doubt a researcher will ever use it but it's probably good to surface the assumption.

apps/backend/scripts/seed-pa.ts
```diff
@@ -0,0 +1,128 @@
...
+  {
+    name: 'English Adaptive',
+    description: 'English adaptive (IRT-based) assessment',
+    params: {
+      language: 'en',
+      isAdaptive: true,
+      itemSelect: 'mfi',
+    },
+  },
+] as const;
```
richford: Is it possible to tie this stuff to the schema package so that it doesn't drift? For example, `PA_SCORING_VERSION` is already there. Is there a way to build the default variants dynamically from the possible values in the schema package?

I don't know exactly how this would work having the variants hard-coded here seems like another opportunity for drift. And one of the major motivations for the monorepo transition was to prevent that.

apps/backend/seeds/pa.seed.ts
```diff
@@ -0,0 +1,128 @@
...
+    console.log(`Inserted PA task: ${task.id}`);
+  } else {
+    console.log(`PA task already exists (${task.id}), skipping insert.`);
+  }
+
+  for (const { name, description, params } of DEFAULT_VARIANTS) {
+    // Query-first: the unique index on task_variants is a functional partial index
+    // (lower(name) WHERE name IS NOT NULL), which Drizzle cannot target in
+    // onConflictDoNothing — so we check existence explicitly rather than relying
+    // on a bare conflict clause that would silently swallow unexpected errors.
```
richford, line 75:
```suggestion
    // Query-first: the unique index on task_variants is a functional partial index
    // (lower(name) WHERE name IS NOT NULL), which Drizzle cannot target in
    // onConflictDoNothing — so we check existence explicitly rather than relying
    // on a bare conflict clause that would silently swallow unexpected errors.
    // Safe in this script because researcher-db-migrate runs as a one-shot via docker compose; concurrent invocation would race each other.
```

richford (changes requested):

apps/backend/scripts/seed-pa.ts
richford: Thinking more about this, it seems natural for the assessment schema package to own this. I guess the big question for you is "what's the long-term home for variant catalogs across assessments?" Even if your answer is "we'll figure it out when we add the third assessment," that's fine. I just want to make sure we are intentional about the placement.

richford: One suggestion: move the runtime-meaningful dimensions to `@roar-dashboard/assessment-schema/pa` and have `seed-pa.ts` derive `DEFAULT_VARIANTS` mechanically:

In `packages/assessment-schema/src/pa/variants.ts`
```ts
import { PA_SCORING_VERSION, PA_SCORE_KIND } from './config';

/**
 * Canonical PA variant kinds. Each entry must be representable in
 * `task_variants.params` and consumed by the PA assessment runtime.
 * Add an entry here when introducing a new scoring approach.
 */
export const PA_VARIANT_KINDS = {
  FIXED: {
    scoringVersion: PA_SCORING_VERSION.V3_FIXED,
    scoreKind: PA_SCORE_KIND.FIXED,
    isAdaptive: false,
    itemSelect: 'fixed',
    label: 'Fixed',
    description: 'Fixed-form assessment',
  },
  ADAPTIVE: {
    scoringVersion: PA_SCORING_VERSION.V4_ADAPTIVE,
    scoreKind: PA_SCORE_KIND.ADAPTIVE,
    isAdaptive: true,
    itemSelect: 'mfi',
    label: 'Adaptive',
    description: 'Adaptive (IRT-based) assessment',
  },
} as const;

/**
 * Languages PA ships variants for. The seed script produces a row per
 * (language, variant kind). Add an entry here when PA's content is
 * localized into a new language.
 */
export const PA_LANGUAGES = {
  en: { code: 'en', label: 'English' },
} as const satisfies Record<string, { code: string; label: string }>;
```

Then in `seed-pa.ts`:
```ts
import { PA_VARIANT_KINDS, PA_LANGUAGES } from '@roar-dashboard/assessment-schema/pa';

const DEFAULT_VARIANTS = Object.values(PA_LANGUAGES).flatMap((lang) =>
  Object.values(PA_VARIANT_KINDS).map((kind) => ({
    name: `${lang.label} ${kind.label}`,
    description: `${kind.description} (${lang.label})`,
    params: {
      language: lang.code,
      isAdaptive: kind.isAdaptive,
      itemSelect: kind.itemSelect,
    },
  })),
);
```

After this, adding a scoring version, a language, or renaming "Adaptive" only requires editing the schema package. The seed picks it up automatically.

scripts/researcher-env-up.sh
ksmontville: The LLM suggests that waiting on `tcp:localhost:4000` is sufficient and doesn't require worrying about self-signed certs.

apps/backend/scripts/seed-pa.ts
ksmontville: It does make sense for the assessment-schema to own this. 

My only issue with the seeding scripts is that changing them requires a PR, and adding new variants/tasks to seed is more cumbersome for the researchers than was simply entering URL parameters. They are also free to insert tasks/variants using PgWeb and such, but that seems more complicated still.

richford (approved):

monorepo-example.md
maximilianoertel, line 1: This looks like it was committed accidentally?

versioning-strategy.md
maximilianoertel, line 1: Same as https://github.com/yeatmanlab/roar-dashboard/pull/1794/changes#r3324688379 ?

packages/api-contract/src/v1/families/post-families-ticket.md
maximilianoertel, line 1: Same as https://github.com/yeatmanlab/roar-dashboard/pull/1794/changes#r3324688379 ?

docker-compose.yml
maximilianoertel, line 1: The researcher environment services shouldn't live in the same `docker-compose.yml` as the OpenFGA dev services. These are two different audiences with conflicting needs: 
- Developers running `docker compose up` now get a Postgres container on port 5432 they don't need, which conflicts with any local Postgres. The OpenFGA services connect to host.docker.internal:5432 expecting a host-side Postgres but now there's a Docker-managed one competing for the same port. 
- Researchers shouldn't need OpenFGA at all, but docker compose up would start it. 

Let's please separate this into its own `docker-compose.researcher.yml` that owns the full researcher stack with proper dependency ordering: researcher-db (healthy) → researcher-db-migrate (completed) → backend → assessment dev server 

This also fixes the race condition in the current `researcher-env-up.sh`, where the backend starts concurrently with migrations. With compose-level `depends_on` and `service_completed_successfully`, the backend only starts after the schema exists. This also eliminates the need for the`migrate.Dockerfile`. The main `docker-compose.yml` stays single-purpose for OpenFGA dev infrastructure.


This would also let us eliminate most of the shell scripts: 
- **researcher-env-up.sh**: the concurrently orchestration, port-conflict checks, cert validation, and Java detection all become unnecessary. Compose handles service ordering natively, and dependency checks become container health conditions. Reduces to `docker compose -f docker-compose.researcher.yml up`. 
- **researcher-env-down.sh**: becomes `docker compose -f docker-compose.researcher.yml down -v`. 
- **wait-for-port.sh**: replaced by compose `depends_on: condition: service_healthy` with a healthcheck on the backend service. 
- **init-researcher-db.sh**: this one stays as standard Docker Postgres entrypoint for creating multiple databases, but it's 3 lines of SQL so it's fine. 

The Firebase Auth emulator could also be a compose service instead of requiring Java on the host? The only thing that should stay on the host is the assessment dev server itself to keep hot-reload against the local files, but that's just one npm run dev command, not a complex script.

apps/backend/seeds/pa.seed.ts
maximilianoertel, line 1: Let's please move this into a dedicated seeds directory `apps/backend/seeds/pa.seed.ts` to establish a convention as more assessments get ported into the monorepo. Each assessment will presumably need its own seed so having a `seeds/` directory with a predictable naming pattern is clearer than accumulating files in scripts/.

packages/assessment-schema/src/researcher-environment.ts
maximilianoertel, line 1: Firebase emulator constants living in the `assessment-schema` seems like a misplacement? 

As far as I see it, the `assessment-schema` package is exclusively for task/scoring domain types shared between the assessment runtime and the backend. Emulator configuration is more of a dev-tooling concern and it belongs in the researcher env configuration. If the researcher environment actually requires environment variables, a dedicated dotenv file would be more appropriate.

apps/assessments/roar-pa/package.json
```diff
@@ -11,6 +11,8 @@
     "build": "echo 'roar-pa: skipped (Vite uses src directly)'",
     "predev": "fkill :8000 -f || true",
```
maximilianoertel: This kills anything on port 8000 before dev starts and could kill completely unrelated processes. A warning + exit would be safer?

apps/assessments/roar-pa/package.json
```diff
@@ -56,6 +58,7 @@
     "@types/node": "^20.9.0",
     "babel-loader": "^8.2.5",
     "concurrently": "^9.1.0",
+    "firebase-tools": "^15.18.0",
```
maximilianoertel: `firebase-tools` isn't actually a devDependency of `roar-pa`. It's currently used by the researcher-env-up script, not by roar-pa's build or dev workflow. But if we move to a compose-based approach as [recommended here](https://github.com/yeatmanlab/roar-dashboard/pull/1794/changes#r3324782523), this dependency becomes redundant anyway.

apps/assessments/shared/firebaseConfig.js
maximilianoertel, line 1: `getFirebaseConfig()` currently handles two scenarios: 
1. **Researcher env** when FIREBASE_AUTH_EMULATOR_HOST is set → returns hardcoded demo-roar config 
2. **Firebase Hosting**  fetches `/__/firebase/init.json` which only works when served by Firebase Hosting or its emulator. 

As far as I can tell, we're missing is the third scenario: developer running npm run dev against a real Firebase project without the emulator? Given we are planing on deprecating the remote dev environment and switching to the Auth Emulator locally as well, this is okay but just flagging here to make sure.

Additionally, `RESEARCHER_LOCAL_FIREBASE_PROJECT_ID` / `RESEARCHER_LOCAL_FIREBASE_API_KEY` make this sound researcher-specific, but it's really just "emulator mode." A developer debugging auth against the emulator hits the same branch. Can we please rename these to`FIREBASE_EMULATOR_PROJECT_ID` and `FIREBASE_EMULATOR_API_KEY`?

maximilianoertel (changes requested):

ksmontville:
@claude Let's check for regressions after rebasing onto ref/1794/monorepo-assessment-schema.

packages/assessment-schema/src/researcher-environment.ts
ksmontville: It's not a perfectly semantic placement, but these constants are used across the /assessments directory  and the /backend directory, which is the exact cross of domains that the /assessment-schema package addresses.

I'll rename them to be more emulator focused rather than researcher focused, however.

apps/assessments/shared/firebaseConfig.js
ksmontville: Check out the script I just added `dev-platform.sh` and the corresponding scripts in `roar-pa/package.json`. This script removes the need for the dev Firebase, in favor of always using the emulator when the `FIREBASE_AUTH_EMULATOR_HOST` variable is set in the backend's environment.

ksmontville:
@claude Let's review the new additions to this PR, which include some variable renaming and introducing a new script to allow assessments to run against a local Postgres instance.

ksmontville:
@claude Feedback implemented, let's review again.

Regarding the heaviness of the dev script, the issue is that there is no real lightweight dev environment anymore. Th Firebase dev project is being phased out in favor of purely local development, either with a local Postgres instance of the ephemeral researcher env. This is a necessary and intentional change for the platform to scale. 

Let's continue to try to find ways to reduce development friction, but there simply will be more intitial setup now than there used to be by nature of the refactor.

ksmontville:
@claude Minor feedback addressed, let's check again.

docker-compose.assessment.yml
```diff
@@ -0,0 +1,158 @@
...
+      FGA_MODEL_ID: "00000000000000000000000000"
+    ports:
+      - "4000:4000"
+    depends_on:
+      researcher-db-migrate:
+        condition: service_completed_successfully
+      firebase-auth-emulator:
+        condition: service_healthy
+    healthcheck:
+      test: ["CMD-SHELL", "curl -sf http://localhost:4000/health"]
```
richford: #1801 removes `GET /health` in favor of `/health/{live,startup,ready}`, so this healthcheck never goes healthy and `docker compose up --wait` fails. Switch to `/health/live`. Note `/health/startup` won't work here because it checks OpenFGA and this backend runs placeholder FGA (`FGA_API_URL: http://placeholder:5050`). Coordinate with #1801.


```suggestion
      test: ["CMD-SHELL", "curl -sf http://localhost:4000/health/live"]
```

scripts/dev-platform.sh
```diff
@@ -0,0 +1,75 @@
...
+  echo "Error: FIREBASE_AUTH_EMULATOR_HOST is not set in apps/backend/.env." >&2
+  echo "  Add this line, then restart the backend before running this script:" >&2
+  echo "" >&2
+  echo "    FIREBASE_AUTH_EMULATOR_HOST=$EMULATOR_HOST" >&2
+  echo "" >&2
+  exit 1
+fi
+
+# The backend must be running before the assessment dev server starts.
+if ! curl -sfk "https://localhost:4000/health" >/dev/null 2>&1; then
```
richford: same `/health` removal issue as the compose healthcheck. Update to `/health/live` once the #1801 endpoint name is settled.


```suggestion
if ! curl -sfk "https://localhost:4000/health/live" >/dev/null 2>&1; then
```

docker-compose.assessment.yml
```diff
@@ -0,0 +1,158 @@
...
+    build:
+      context: .
+      dockerfile: researcher.Dockerfile
+      target: researcher-runtime
+    environment:
+      NODE_ENV: production
+      PORT: "4000"
+      CORE_DATABASE_URL: postgres://postgres@researcher-db:5432/roar_core
+      ASSESSMENT_DATABASE_URL: postgres://postgres@researcher-db:5432/roar_assessment
+      GOOGLE_CLOUD_PROJECT: demo-roar
```
richford, line 137: `GOOGLE_CLOUD_PROJECT: demo-roar` is inline even though the emulator project ID was centralized as `FIREBASE_EMULATOR_PROJECT_ID`. Two sources of the same value can drift. Reference the constant if possible, or add a comment explaining why the compose env is intentionally separate.

richford (changes requested):

docker-compose.assessment.yml
ksmontville: It's a .yml file so we can't import the constant directly, but I'll add anote to the .yml file that it's the canonical value.

ksmontville:
@claude Let's check this PR for regressions after rebasing onto its base branch project/backend-refactor.

apps/backend/scripts/seed-pa.ts
maximilianoertel: @ksmontville I'd say that it's completely okay that changing seeds require a PR. If a researcher develops a new variant, just like with any other development workflow, the changes can be done manually by the researcher directly on the database. Then, once happy with the new variant, it gets committed and updated via a PR.  Especially with the change proposed by @richford where the seeds are automatically generated from the assessment-schema package!

richford (approved):

PR #1801, opened 2026-05-22
Add health check endpoints for liveness, startup, and readiness probes

maximilianoertel:
## Summary

This PR adds dedicated health check endpoints for Cloud Run's liveness, startup, and readiness probes, replacing the previous unconditional `GET /health` endpoint with three purpose-built routes under `/health/*`. Startup and liveness probes return unconditional 200s and only the readiness probe checks external dependencies.

## Endpoints

| Endpoint | Probe | Behavior |
|----------|-------|---------|
| `GET /health/live` | Liveness | Unconditional 200 — confirms the process is running |
| `GET /health/startup` | Startup | Unconditional 200 — confirms the process has started |
| `GET /health/ready` | Readiness | Checks Postgres + OpenFGA, returns 200 or 503 with per-dependency status |

## Why startup doesn't check dependencies

The previous design had the startup probe checking Postgres and OpenFGA. This is architecturally wrong:

- **Startup probe failure kills and restarts the container.** If FGA or Postgres has a transient outage, every new container fails startup, blocking deploys and causing restart loops on scale-up.
- **Readiness probe failure drains traffic and self-heals.** The container stays alive; when the dependency recovers, traffic resumes automatically.

Neither DB nor FGA actually blocks server startup — `initializeDatabasePools()` creates Pool objects lazily, `FgaClient.initialize()` only sets up OIDC auth when configured, and Express binds to a port regardless of either.

## Key design decisions

- **Plain Express routes** — not ts-rest contract routes. These are infrastructure endpoints that don't need Zod validation, auth middleware, or typed responses.
- **`withTimeout` helper** — wraps `Promise.race` with `clearTimeout` in a `finally` block to prevent timer leaks under frequent probing.
- **`HealthCheckTimeoutError` sentinel class** — reliable timeout detection via `instanceof` instead of fragile string-matching.
- **`Promise.allSettled`** — ensures one dependency failure doesn't prevent reporting the other's status.
- **Only successful results are cached** (3s TTL, below the default 10s probe interval) — failed checks are never cached so probes detect recovery immediately.
- **SIGTERM-aware readiness** — `setShuttingDown()` is called before `server.close()`, so the readiness probe returns 503 while connections drain. Liveness and startup remain 200 during shutdown to prevent Kubernetes from killing the draining container.
- **Shutdown grace period** — 10s force-exit safety net via `setTimeout(...).unref()`.

Resolves yeatmanlab/roar-project-management/issues/1819


apps/backend/src/db/clients.ts
```diff
@@ -49,44 +49,72 @@ let corePool: Pool | null = null;
...
  *
  * @returns The initialized assessment database client.
- * @throws {Error} If {@link initializeDatabasePools} has not been called.
+ * @throws {Error} If database pools are not yet initialized (initialization is triggered automatically).
  */
 export function getAssessmentDbClient(): AssessmentDbClientType {
   if (!AssessmentDbClient) {
-    throw new Error('Database not initialized. Call initializeDatabasePools() first.');
+    triggerLazyInit();
+    throw new Error('Database is initializing. The request will succeed once initialization completes.');
```
richford:
```suggestion
    throw new Error('Database initialization in progress. Retry this request once initialization completes.');
```

apps/backend/src/db/clients.ts
```diff
@@ -49,44 +49,72 @@ let corePool: Pool | null = null;
...
  *
  * @returns The initialized core database client.
- * @throws {Error} If {@link initializeDatabasePools} has not been called.
+ * @throws {Error} If database pools are not yet initialized (initialization is triggered automatically).
  */
 export function getCoreDbClient(): CoreDbClientType {
   if (!CoreDbClient) {
-    throw new Error('Database not initialized. Call initializeDatabasePools() first.');
+    triggerLazyInit();
+    throw new Error('Database is initializing. The request will succeed once initialization completes.');
```
richford:
```suggestion
    throw new Error('Database initialization in progress. Retry this request once initialization completes.');
```

apps/backend/src/db/clients.ts
```diff
@@ -49,44 +49,72 @@ let corePool: Pool | null = null;
...
 }
 
+/**
+ * Triggers lazy database initialization if not already initialized or in progress.
+ *
+ * Called automatically by {@link getCoreDbClient} and {@link getAssessmentDbClient}
+ * on first access. Stores the promise to prevent duplicate concurrent initialization.
+ * Logs fatal and exits the process if initialization fails (so Cloud Run can restart).
+ */
+function triggerLazyInit(): void {
```
richford: Can we tests for `triggerLazyInit()`?

- A test that verifies a request arriving before initialization triggers `initializeDatabasePools()` exactly once across N concurrent calls.
- A test that verifies `initPromise` is cleared on failure so the next request triggers a fresh init attempt.

apps/backend/src/db/clients.ts
```diff
@@ -49,44 +49,72 @@ let corePool: Pool | null = null;
...
+
+  initPromise = initializeDatabasePools()
+    .then(() => {
+      logger.info('Database pools initialized (lazy)');
+    })
+    .catch((err) => {
+      logger.fatal({ err }, 'Database initialization failed — shutting down');
+      initPromise = null;
+      process.exit(1);
+    });
```
richford:
```suggestion
    .then(() => {
      logger.info('Database pools initialized (lazy)');
      initPromise = null; // clear once we have the cached clients
    })
    .catch((err) => {
      logger.fatal({ err }, 'Database initialization failed');
      initPromise = null;
      // Defer to the next event-loop tick so the readiness probe has time to flip.
      // SIGTERM lets the existing graceful-shutdown handler drain in-flight reqs.
      setImmediate(() => process.kill(process.pid, 'SIGTERM'));
    });
```

apps/backend/src/server.ts
```diff
@@ -126,6 +116,16 @@ async function startServer(): Promise<void> {
...
   process.on('SIGTERM', () => shutdown('SIGTERM'));
   process.on('SIGINT', () => shutdown('SIGINT'));
+
+  // FGA OIDC initialization requires an async network call to Google APIs.
+  // Fire-and-forget at startup so the OIDC client is ready by the time
+  // the first FGA-dependent request arrives. Non-OIDC environments skip this.
+  FgaClient.initialize().catch((err) => {
+    logger.warn(
+      { err },
+      'FGA client initialization failed — FGA will be unavailable until the client is lazily initialized',
```
richford: As Claude suggested earlier.
```suggestion
      'FGA-dependent endpoints will return 500 until the server restarts with OpenFGA reachable.',
```

richford (changes requested):
Thanks Max for delving into this. Some suggestions. Feel free to push back:

- Use Cloud Run's startup probe to gate initial routing. Add a `/startup` endpoint that returns 503 until `initialized === true` and (when `FGA_OIDC_AUDIENCE` is set) the FGA OIDC axios instance is attached. Wire it to `spec.template.spec.containers[].startupProbe` in the Cloud Run service YAML; the existing unconditional `/health` stays wired to `livenessProbe`. The startup probe only runs at boot, so this addresses the cold-start gap but not mid-life degradation. Mid-life degradation has to go through SIGTERM (see inline suggestion below).
- Make `triggerLazyInit` retryable with backoff. Two attempts with exponential backoff before exiting would absorb most transient cases.
- Add an integration test that boots the server with `CORE_DATABASE_URL` pointed at a closed port, then hits `/v1/me`, and asserts the request returns a 503 or 500 with structured logs. This would verify the lazy-init contract works without a real Cloud Run deployment.

Also note:

- The lazy-init pattern conflicts subtly with @ksmontville's monorepo work in #1794. That introduces `researcher-env-up.sh` which uses `--kill-others-on-fail` to tear down on race conditions. Combined with lazy DB init the backend would "start" without a DB, the researcher env would think it's healthy, and the first assessment fetch would 500.
- The `migrate.Dockerfile` (also in @ksmontville's #1794) is also relevant. Do I understand correctly that migrations no longer run at boot, so they have to run via this separate Dockerfile in deployment. If this PR lands before #1794, prod migrations won't run on next deploy. If I'm right, then we should make #1794 a hard prerequisite for merging this.

maximilianoertel:
@claude review

maximilianoertel:
@claude feedback implemented, please re-review

apps/backend/src/app.ts
```diff
@@ -5,15 +5,14 @@ import { registerAllRoutes } from './routes';
...
 
 const app = express();
 
 app.use(requestLogger);
 app.use(express.json());
 
-app.get('/health', (_req: Request, res: Response) => {
-  res.status(StatusCodes.OK).json({ status: 'ok' });
-});
+app.use(healthRouter);
```
richford, line 15: Heads up that removing `GET /health` will break #1794's researcher health checks. If you want to prevent that, we could keep a `/health` alias for `/health/live`. Otherwise, we just need to coordinate so that if this get's merged first, then we switch over #1794 / #1806 accordingly.

Tagging @ksmontville

apps/backend/src/health/health-routes.ts
```diff
@@ -0,0 +1,55 @@
...
+ * @param res - Express response object
+ */
+async function respondWithHealthCheck(res: Response): Promise<void> {
+  try {
+    const result = await runHealthChecks();
+    const status = result.status === 'ok' ? StatusCodes.OK : StatusCodes.SERVICE_UNAVAILABLE;
+    res.status(status).json(result);
+  } catch (err) {
+    logger.error({ err }, 'Unexpected error from runHealthChecks');
+    res.status(StatusCodes.SERVICE_UNAVAILABLE).json({ status: 'error' });
```
richford: The success path returns `{ status, checks: { postgres, openfga } }` but this catch-all returns `{ status: 'error' }` with no `checks`. Suggest returning a `checks` object (either empty or all-`error`) here for shape stability. Nit pick.

apps/backend/src/health/health-routes.ts
```diff
@@ -0,0 +1,55 @@
...
+ * stops routing new requests to this instance.
+ */
+healthRouter.get('/health/ready', async (_req, res) => {
+  if (isShuttingDown()) {
+    res.status(StatusCodes.SERVICE_UNAVAILABLE).json({ status: 'shutting_down' });
+    return;
+  }
+
+  await respondWithHealthCheck(res);
+});
```
richford, line 60: Correction to an earlier note of mine: Cloud Run **does** support an HTTP readiness probe (`--readiness-probe httpGet.path="/health/ready"`) ([docs](https://docs.cloud.google.com/run/docs/configuring/healthchecks)). So `/health/ready` has a natural consumer and should be wired it to the readiness probe. Sorry for my previous error there.

apps/backend/src/health/health-checks.ts
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  // checkPostgres and checkOpenFga catch all errors internally and never reject.
+  // Promise.allSettled is a defensive guard in case that contract changes.
+  const [postgresResult, openfgaResult] = await Promise.allSettled([checkPostgres(), checkOpenFga()]);
+
+  const postgres = postgresResult.status === 'fulfilled' ? postgresResult.value : 'error';
+  const openfga = openfgaResult.status === 'fulfilled' ? openfgaResult.value : 'error';
+
+  const result: HealthCheckResult = {
```
richford: The strict verdict here (`postgres === 'ok' && openfga === 'ok'`) is a good check. The question I have is about which probe consumes it. The dependency is soft at the process level: `Promise.allSettled` (line 111) only stops one check from masking the other, and the 200/503 comes from this `&&`, not the combinator.

In `health-routes.ts`, you wire this up to `/health/startup` and `/health/ready`. But I wonder if we should only wire this strict check to Cloud Run's **readiness** probe via `/health/ready` (not startup, not liveness). A readiness failure drains traffic from the instance and self-heals when FGA recovers, without blocking deploys or restarting. Could we keep `/health/live` on liveness (unconditional) and `startup` light (Postgres-only)?

Here's my reasoning: startup failure kills and retries the instance so an FGA outage would block deploys (the new revision never passes startup) and trigger cold-start restart loops on any scale-up, whereas readiness just drains and self-heals.

What do you think?

richford (changes requested):

apps/backend/src/app.ts
ksmontville: It's fine to merge this first, and then update this line in `docker-compose.researcher.yml` according to the new route:

```ts
healthcheck:
      test: ["CMD-SHELL", "curl -sf http://localhost:4000/health"]
```

ksmontville (dismissed):

richford (approved):

ksmontville (approved):

PR #1811, opened 2026-05-29
Add CORS and security headers middleware

maximilianoertel:
## Summary

This PR adds CORS and HTTP security headers to the backend API via two new middleware modules (`cors/` and `security-headers/`), each following the existing one-concern-per-directory middleware convention.

**CORS** (`middleware/cors/`): Configured from the `ALLOWED_ORIGINS` env var (comma-separated, exact match only). Enables credentials, 24-hour preflight cache. Allows `sentry-trace` and `baggage` request headers for Sentry distributed tracing ([docs](https://docs.sentry.io/platforms/javascript/tracing/distributed-tracing/dealing-with-cors-issues/)).

**Security headers** (`middleware/security-headers/`): Sets `Content-Security-Policy: default-src 'none'`, `X-Frame-Options: DENY`, HSTS, `nosniff`, `Referrer-Policy: no-referrer`, and other Helmet defaults. Disables COEP, COOP, CORP, and `Origin-Agent-Cluster` since they address document/frame concerns not applicable to a JSON API.

Resolves yeatmanlab/roar-project-management#1835


apps/backend/src/middleware/cors/parse-allowed-origins.ts
```diff
@@ -0,0 +1,30 @@
...
+        .split(',')
+        .map((s) => s.trim())
+        .filter(Boolean),
+    ),
+  ];
+
+  if (origins.length === 0) {
+    logger.warn({ defaultOrigin: DEFAULT_ORIGIN }, 'ALLOWED_ORIGINS is not set or empty, falling back to default');
+    return [DEFAULT_ORIGIN];
+  }
```
richford, line 35: In production, an unset/empty `ALLOWED_ORIGINS` falls back to `https://localhost:5173` with only a `warn`. That's harmless to security (CORS is browser-only and AuthGuard is the real control), but it converts a config error into an easily-missed warning and silently ships a dev allowlist. Consider throwing when `NODE_ENV === 'production'` and `ALLOWED_ORIGINS` is missing, so a broken deploy fails loudly.

apps/backend/src/app.ts
```diff
@@ -3,11 +3,16 @@ import type { Request, Response } from 'express';
...
+import { securityHeadersMiddleware } from './middleware/security-headers/security-headers.middleware';
 import { requestLogger } from './middleware/request-logger/request-logger.middleware';
 import { ApiErrorCode } from './enums/api-error-code.enum';
+import { ApiErrorMessage } from './enums/api-error-message.enum';
 
 const app = express();
 
+app.use(securityHeadersMiddleware);
+app.use(corsMiddleware);
 app.use(requestLogger);
```
richford: Should `requestLogger` come before `corsMiddleware`? Right now, preflight stuff will be invisible when debugging a blocked origin. But I don't know what's standard here or if there any any drawbacks to putting `corsMiddleware` after `requestLogger`.

apps/backend/src/middleware/cors/cors.middleware.ts
```diff
@@ -0,0 +1,24 @@
+import cors from 'cors';
+import { parseAllowedOrigins } from './parse-allowed-origins';
+
+const allowedOrigins = parseAllowedOrigins(process.env.ALLOWED_ORIGINS);
```
richford, line 7: `parseAllowedOrigins(process.env.ALLOWED_ORIGINS)` is evaluated at module import, so origins are frozen at first import and an env change needs a restart. This is totally fine but I think it's worth a brief comment, since it differs from the lazy env-read convention in `fga.client.ts`.

apps/backend/src/middleware/cors/cors.middleware.ts
```diff
@@ -0,0 +1,24 @@
...
+ * - 24-hour preflight cache (browsers may cap lower: Chrome 2h, Firefox 24h)
+ *
+ * NOTE: CORS is a browser-side enforcement mechanism only. Non-browser clients
+ * (curl, server-to-server) receive full responses regardless of origin.
+ * AuthGuardMiddleware is the actual server-side access control.
+ */
+export const corsMiddleware = cors({
+  origin: allowedOrigins,
+  methods: ['GET', 'POST', 'PUT', 'PATCH', 'DELETE', 'OPTIONS'],
+  allowedHeaders: ['Content-Type', 'Authorization', 'X-Request-ID', 'sentry-trace', 'baggage'],
```
richford, line 23: Minor: the dashboard's SDK context doesn't pass `requestId` yet, so the dashboard path doesn't emit it today. You could wire it up here if you want dashboard request tracing.

richford (changes requested):

apps/backend/.env.example
```diff
@@ -15,6 +15,10 @@ ASSESSMENT_DATABASE_USER=
 # - Google Cloud Project ID
 GOOGLE_CLOUD_PROJECT=
 
+# CORS
+# Comma-separated list of allowed origins. Defaults to https://localhost:5173 if not set.
+ALLOWED_ORIGINS=https://localhost:5173
```
ksmontville, line 23: The assessments run on `https://localhost:8000`, so we should probably include it here.

Looking forward at the assessment monorepo migration, the plan for each of the assessments in the monorepo was to run on a specific port counting up from 8000.

So for example:
roar-pa on localhost:8000
roar-swr on localhost:8001
roar-sre on localhost:8002
and so on...

This is to allow a researcher to run multiple assessments without port conflicts. However, this means that the allowed origins perpetually grows as assessments are migrated to the monorepo, and it might just be better to enforce a single port for all assessment webpack servers. I'm not sure how often it is the case that researchers need to run multiple assessments in parallel, in any case.

Alternatively, we just continue to use the webpack proxy at `/v1` (implemented in #1794 ) and not list any assessment dev servers under allowed origins.

Just thinking out loud here.

ksmontville (changes requested):

richford:
@claude please review this PR

richford:
@claude, please re-review. Investigate whether we have resolved previous review findings for this PR.

richford (dismissed):

package-lock.json
maximilianoertel, line 1: @richford I'm curious why this change was necessary? As far as I know, `npm ci` works just fine on a local macOS machine and this is only needed when running CI on macOS runners (which we don't). Did you run into a specific issue that triggered this commit?

apps/backend/src/middleware/cors/cors.middleware.ts
maximilianoertel: Let's keep the requestId wiring for later as there is more work to get this to work properly

apps/backend/.env.example
maximilianoertel: Good points. A few thoughts: 

1. Port 8000 is too common. Many default HTTP servers and other tools default to that port. We should pick a range that is less likely to collide. I'd suggest something along the lines of 8100+ (`roar-pa = 8100, roar-swr = 8101, etc.). This range is clearly distinct from backend (4000), dashboard (5173), OpenFGA (5050) and the usual 8000/8080 cluster. 

2. Let's go with dedicated ports per assessment, not a shared one. Beyond researchers, other engineers work on these too, and forcing everyone onto a single port makes DX significantly worse. 

3. I'd like to get rid of the /v1 proxy mostly just to avoid maintaining this additional configuration.

4. Not every port needs to be baked into `.env.example`. I'd argue that setup instructions for researchers should include instructions to add their assessment ports when they create their `.env` from the example

package-lock.json
richford: I was having trouble running `npm ci` on my mac. But also, I think the lockfile regressed. `main` has a multi-platform lockfile (darwin + win32 + linux `@rollup`/`@esbuild` native nodes). PR #1794 (36b1bf4e9) regenerated it down to 0 darwin / 0 win32 / 4 linux. I think it's likely that someone did an `npm install` on a Linux machine.

Test it out: take a complete rollup lockfile, delete the current platform's `@rollup/rollup-<platform>` node, then `rm -rf node_modules && npm ci`. On my machine, the binary is not installed and rollup throws `Cannot find module @rollup/rollup-<platform>`. For me, with no `darwin-arm64` node, that's what happened with `npm ci`: no darwin rollup/esbuild binary and then the build/tests crash.

So you're right it's not about macOS runners. It's the local npm ci on macOS against this lockfile. Having said that, I'm happy to drop this commit if you think it belongs in another issue/PR. It is your PR, after all.

richford (approved):

ksmontville (approved):

PR #1814, opened 2026-06-02
Persist PA computed scores through the new pipeline

Emily-ejag:
## Proposed changes

Implemented the PA computed-score pipeline producer side.

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist


Emily-ejag:
@claude please review

packages/assessment-sdk/src/compat/firekit.ts
```diff
@@ -669,11 +694,38 @@ export async function writeTrial(
 
   const cmd = new WriteTrialCommand(api, ctx.participant.participantId);
 
+  // Invoke computed score callback if provided and map to ScoreEntry[] for persistence
+  let scores: ScoreEntry[] | undefined;
+  if (computedScoreCallback) {
+    try {
+      const rawScores = facade._getRawScores?.();
```
richford: This is the line that makes the whole feature a no-op: `const rawScores = facade._getRawScores?.()` and `_getRawScores()` are stubs that return `undefined`. So is `_getScoreAdapter()` (line 236). Nothing overrides either and `toPaScoreEntries` is never imported outside its own file/test. So `rawScores` is always `undefined`, then the `if (rawScores)` block never runs, so the callback isn't invoked and `scores` stays `undefined`. 

So I think this is causing computed scores not to persist end-to-end. To make it work, roar-pa (or `initFirekitCompat`) must override `_getRawScores()` to return the accumulated raw scores and `_getScoreAdapter()` to return `toPaScoreEntries`, and add a test asserting `writeTrial(data, cb)` produces a `scores` array on the emitted event.

packages/assessment-sdk/src/compat/firekit.ts
```diff
@@ -218,6 +218,33 @@ export class FirekitFacade {
...
+  }
+
+  /**
+   * Internal getter for score adapter function.
+   * Returns undefined by default; can be overridden by assessment-specific implementations.
+   * @internal
+   */
+  _getScoreAdapter(): ((scores: ComputedScores) => ScoreEntry[]) | undefined {
+    return undefined;
+  }
```
richford, line 238: `_getRawScores()` and `_getScoreAdapter()` return `undefined` by default and are never overridden, so the `writeTrial` score path above is dead. If the intent is for assessments to override these, ship the PA override in this PR (or via `initFirekitCompat`). Otherwise the SDK plumbing has no producer and the "scores persisted" claim doesn't hold.

packages/assessment-schema/src/roar-pa/score-entries.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  // Process composite and composite_foundational groups
+  for (const groupKey of ['composite', 'composite_foundational']) {
+    const groupScores = computed[groupKey];
+    if (groupScores) {
+      for (const summaryName of SUMMARY_NAMES) {
+        add(summaryName, groupScores[summaryName]);
+      }
+    }
+  }
```
richford: The `add` helper hard-codes `domain: PA_TASK_ID` (`'pa'`, line 88) for **every** entry. But in `run_scores`, `domain` is the *measured construct*, not the task slug. The schema doc says "Assessment domain being measured (e.g. phonics, vocabulary)," and the backend's best-run recompute reads composite scores by `domain = SCORE_DOMAIN.COMPOSITE` (`'composite'`) with names `thetaSE`/`numAttempted` (`apps/backend/src/repositories/run.repository.ts:169-172`). So there are two problems:

1. `composite` and `composite_foundational` are written under the same `(type='computed', domain='pa', name, stage=null)` key and collide, overwriting each other. The second silently overwrites the first, clobbering the student's composite; and
2. More fundamentally, writing `domain='pa'` at all means the recompute (which queries `domain='composite'`) never finds these scores. Fix: set `domain` to the construct — `composite` → `SCORE_DOMAIN.COMPOSITE` (`'composite'`), `composite_foundational` → a new `'composite_foundational'` value added to the backend `SCORE_DOMAIN` constants; don't keep `'pa'`. (`domain: groupKey` happens to produce exactly those values.)

Do **not** disambiguate by name instead — the backend keys composite lookups on `domain`, not name, so renaming would just hide the scores a different way. Also note the recompute consumes `thetaSE`/`numAttempted` under `domain='composite'`, which this mapping doesn't emit. Confirm the full set the backend reads.

Can you add a test asserting both groups land under distinct domains and survive the upsert. (`scores.js` also spreads `...composite_foundational` into `composite` upstream, so settle the semantics together.)

packages/assessment-schema/src/pa/score-entries.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  // Iterate over subtasks (FSM, LSM, DEL) in canonical order
+  for (const subtaskKey of PA_SUBTASK_KEYS) {
+    const subtaskKeyLower = subtaskKey.toLowerCase();
+    const subtaskScores = computed[subtaskKeyLower];
+
+    if (subtaskScores) {
+      const def = PA_SUBSCORE_DEFS[subtaskKey];
+      add(def.correctName, subtaskScores.numCorrect);
+      add(def.attemptedName, subtaskScores.numAttempted);
```
richford: but `scores.js` stores only `numCorrect`/`percentCorrect` per subtask, not `numAttempted`, so `add` skips it (`value == null`). `fsmAttempted`/`lsmAttempted`/`delAttempted` are in `PA_SUBSCORE_DEFS` but will never be written. Either emit `numAttempted` from the callback or drop the `*Attempted` names from the schema.

packages/assessment-schema/src/roar-pa/score-entries.ts
```diff
@@ -0,0 +1,132 @@
...
+ * //   { type: 'computed', domain: 'pa', name: 'roarScore', value: '25' },
+ * //   { type: 'computed', domain: 'pa', name: 'percentile', value: '60' },
+ * //   { type: 'computed', domain: 'pa', name: 'standardScore', value: '105' },
+ * //   ...
+ * // ]
+ * ```
+ */
+export function toPaScoreEntries(
+  computed: Record<string, any>,
+  { strict = false } = {},
```
richford, line 133: `strict` defaults to `false`, and the (un-wired) SDK adapter path wouldn't pass it. So the no-drift guard only runs in the unit test. When the adapter is wired, call `toPaScoreEntries(computed, { strict: true })` in dev/CI so an unregistered name fails loudly rather than being silently dropped.

packages/assessment-schema/src/roar-pa/score-entries.ts
```diff
@@ -0,0 +1,132 @@
...
+ * Mirrors the shape of the api-contract ScoreEntry type but kept local
+ * to avoid coupling assessment-schema to api-contract.
+ */
+export interface ComputedScoreEntry {
+  type: 'computed';
+  domain: string;
+  name: PaScoreName;
+  value: string;
+  assessmentStage?: 'practice' | 'test';
+}
```
richford, line 33: `ComputedScoreEntry` is hand-duplicated from the contract's `ScoreEntry` ("kept local to avoid coupling"). Reasonable, but nothing ties it to `ScoreEntrySchema`, so the two can drift. Since the SDK already depends on `@roar-dashboard/api-contract`, add a `satisfies ScoreEntry[]` (or equivalent) where the adapter is wired so a contract change surfaces at compile time.

packages/assessment-sdk/src/compat/firekit.ts
richford: Stale JSDoc: "Supports optional computed score callback (not yet implemented)". This PR is implementing it. Can you update?

richford: Another stale JSDoc: "(not yet implemented)".

richford (changes requested):
Thanks @Emily-ejag! The `toPaScoreEntries` mapping is well-designed and unit-tested. Two blockers before it does what the description claims, plus the two CI failures both trace to the regenerated lockfile (not the feature code):

1. **The score path isn't wired**. `writeTrial` gates the callback behind `facade._getRawScores()` and maps via `facade._getScoreAdapter()`, but both are stubs returning `undefined` and nothing overrides them, and `toPaScoreEntries` is never imported by the SDK or roar-pa. So the callback never runs and no `scores` are emitted. Computed scores are still dropped. The unit tests pass because they call `toPaScoreEntries` directly.
2. **`composite` and `composite_foundational` collide** on the backend's natural key (same `type`/`domain`/`name`/stage), so one silently overwrites the other.

For the CI failures, we have

- `strip-literal` (assessment-schema vitest run): the regenerated `package-lock.json` is inconsistent and mixes vitest majors — `^2.1.8` (assessment-sdk) / `^2.0.5` (dashboard) vs `^3.2.4` (schema/api-contract/backend), resolving `@vitest/runner` at both 2.1.9 and 3.2.4. The v2 runner `import`s `strip-literal`, but the lockfile has no `node_modules/strip-literal` install node, so `npm ci` never installs it. Align vitest to `^3.2.4` everywhere, delete `node_modules` + `package-lock.json`, and `npm install` for a coherent lockfile (v3's runner doesn't need `strip-literal`).

- `vitest/no-conditional-expect` (a bunch of places): the rule flags `expect()` reached only inside a conditional (`if`/`try`/`catch`/ternary), since a non-taken branch silently skips the assertion (false green flag). It's firing broadly now because the lockfile regen moved `@vitest/eslint-plugin` (declared `^1.3.4`) to **1.6.19**, whose recommended set (used via `config-eslint/index.js`) flags pre-existing patterns. Fix via the clean lockfile regen (pin/align the plugin) and/or rewrite genuine offenders (`try { … } catch (e) { expect(e)… }` → `await expect(fn()).rejects.toThrow(…)`).

richford (changes requested):
One more request for this PR:

Can you add `RoarScores` scoring tests to close the coverage gap. This PR tests `toPaScoreEntries` (the mapping) and the SDK, but not `RoarScores.computedScoreCallback` (`apps/assessments/roar-pa/src/experiment/scores.js`), the scoring math itself, which is untested on the base branch too and is the highest-risk tier. Since this is the PR that makes those scores reach the backend, please add `RoarScores` unit tests here for:

- adaptive (v4/v5) vs fixed (v3) selection by `scoringVersion`, incl. the new non-finite `throw`,
- the `taskId !== PA_TASK_ID` guard returning `null`
- the `composite_foundational` theta/SE mapping and its spread into `composite` (≈ `scores.js:222`). Here, assert the resulting `composite`, since that merge can move a student's score
- the grade `<6` vs `≥6` lookup branch. (Mock `store2` session config + `thetas`/`thetaSEs`, `papaparse`'s `initTable` download, and `getGrade`.)
- Pair it with the end-to-end `writeTrial`→`scores` test from the wiring comment.

With those, we'll actually have some test coverage for the scoring.

Emily-ejag:
@claude can you please make sure I have addressed all the comments and did not introduce new errors?

Emily-ejag:
@claude can you re-review making sure I have addressed all the comments?

Emily-ejag:
@claude how about now?

Emily-ejag:
@claude what about now :) ?

Emily-ejag:
@claude please re-review

Emily-ejag:
@claude please re-review

Emily-ejag:
@claude and now?

Emily-ejag:
@claude and now?

Emily-ejag:
@claude I chose to have subtasks and the domains to be the subtasks

richford (changes requested):
I think you responded to all of my previous comments. But I think in order to get the CI passing again, you need to do

```
rm -f package-lock.json
npm install
```

`npm ci` won't fix it.

Emily-ejag:
@claude please review I haven't introduced any error after recent rebasing

Emily-ejag:
@claude what about now?

richford:
@claude, please re-review.

richford:
@claude, please re-review

richford (approved):

PR #1816, opened 2026-06-03
Update Security Policies

kellyel:
## Proposed changes
This PR updates our headers in response to a review of our security. For Content-Security-Policy, we will now be enforcing the header in production rather than just reporting. To reflect this, we have added URLs to the policy that have been reported in the past several months. There should be no changes to the end user.

Additionally, we have added the Cross-Origin-Resource-Policy and Cross-Origin-Embedding-Policy. These should also not change the user experience. 

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist

### DIFF
--- a/apps/dashboard/firebase/admin/csp.template.json
+++ b/apps/dashboard/firebase/admin/csp.template.json
@@ -100,7 +100,6 @@
   ],
   "connect-src": [
     "'self'",
-    "data:",
     "https://*.analytics.google.com",
     "https://*.google-analytics.com",
     "https://*.googleapis.com",
--- a/apps/dashboard/firebase/admin/csp.template.json
+++ b/apps/dashboard/firebase/admin/csp.template.json
@@ -15,20 +15,24 @@
     "https://www.gstatic.com/recaptcha/",
     "blob:"
   ],
+  "script-src-elem": ["'self'", "'unsafe-inline'", "https://cdn.jsdelivr.net"],
   "style-src": [
     "'self'",
     "'unsafe-inline'",
     "https://cdn.jsdelivr.net/npm/katex@0.16.8/",
     "https://fonts.googleapis.com",
     "https://googletagmanager.com",
+    "https://www.gstatic.com",
     "https://tagmanager.google.com"
   ],
+  "style-src-elem": ["'self'", "'unsafe-inline'", "https://cdn.jsdelivr.net", "https://www.gstatic.com"],
   "img-src": [
     "'self'",
     "data:",
     "https://*.google-analytics.com",
     "https://*.googletagmanager.com",
     "https://googletagmanager.com",
+    "https://eyetrackingdata.blob.core.windows.net",
     "https://raw.githubusercontent.com/yeatmanlab/",
     "https://ssl.gstatic.com",
     "https://storage.googleapis.com/road-dashboard/",
@@ -56,9 +60,12 @@
     "https://storage.googleapis.com/roav-readaloud/",
     "https://storage.googleapis.com/trog/",
     "https://storage.googleapis.com/roav-mp/",
-    "https://www.gstatic.com"
+    "https://www.gstatic.com",
+    "https://fonts.gstatic.com",
+    "https://maps.gstatic.com"
   ],
   "media-src": [
+    "blob:",
     "https://storage.googleapis.com/road-dashboard/",
     "https://storage.googleapis.com/roam-apps/",
     "https://storage.googleapis.com/roar-ak/",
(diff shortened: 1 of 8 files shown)


richford:

kellyel:
I have updated this PR to remove `data:` from `connect-src` as I agree that it is non-standard. I have also changed the COEP header to be report-only for a short period. We want to enforce `require-corp`, but we should ensure that it does not impact the end user before enforcing. Similarly, since we made changes to the CSP, we will be keeping it as report-only for a short time to ensure there are no remaining domains that need to be added to the whitelist. 

With these decisions in mind, @claude please review this PR again.

Emily-ejag (changes requested):
I believe @amyyeung17 and I need these extra ones for running `read-aloud` and `ran`

```
script-src-elem:
  'self'
  'unsafe-inline'
  https://cdn.jsdelivr.net

style-src:
  https://www.gstatic.com

style-src-elem:
  'self'
  'unsafe-inline'
  https://cdn.jsdelivr.net
  https://www.gstatic.com

img-src:
  https://storage.googleapis.com/roar-vocab/
  https://fonts.gstatic.com
  https://maps.gstatic.com

media-src:
  https://storage.googleapis.com/roar-vocab/

font-src:
  https://storage.googleapis.com/roar-ak/font/primer_print_bold.ttf

connect-src:
  https://cdn.jsdelivr.net

child-src:
  blob:
```

kellyel:
@Emily-ejag, thanks for taking a look at these. I believe all of these items are included in this PR, or were preexisting. If there are any that are not here, please let me know so I can get them added.

apps/dashboard/firebase/admin/csp.template.json
```diff
@@ -15,20 +15,24 @@
     "https://www.gstatic.com/recaptcha/",
     "blob:"
   ],
+  "script-src-elem": ["'self'", "'unsafe-inline'", "https://cdn.jsdelivr.net"],
```
maximilianoertel: If I recall correctly, when `script-src-elem` is set, browsers use it instead of `script-src` for every `<script>` element, not in _addition_ to it. So this change would block GTM, the Sentry CDNs, reCAPTCHA, *.googleapis.com

apps/dashboard/firebase/admin/csp.template.json
```diff
@@ -15,20 +15,24 @@
...
   "style-src": [
     "'self'",
     "'unsafe-inline'",
     "https://cdn.jsdelivr.net/npm/katex@0.16.8/",
     "https://fonts.googleapis.com",
     "https://googletagmanager.com",
+    "https://www.gstatic.com",
     "https://tagmanager.google.com"
   ],
+  "style-src-elem": ["'self'", "'unsafe-inline'", "https://cdn.jsdelivr.net", "https://www.gstatic.com"],
```
maximilianoertel: Same mechanics as `script-src-elem` above

apps/dashboard/firebase/admin/csp.template.json
```diff
@@ -92,20 +99,28 @@
...
+    "https://cdn.jsdelivr.net",
     "https://firestore.googleapis.com",
     "https://identitytoolkit.googleapis.com",
     "https://raw.githubusercontent.com/yeatmanlab/",
     "https://us-central1-__ADMIN_PROJECT_ID__.cloudfunctions.net/",
-    "https://us-central1-__ASSESSMENT_PROJECT_ID__.cloudfunctions.net/"
+    "https://us-central1-__ASSESSMENT_PROJECT_ID__.cloudfunctions.net/",
+    "https://eyetrackingdata.blob.core.windows.net"
   ],
+  "child-src": ["blob:"],
```
maximilianoertel: Could you please explain why we're adding this one?

apps/dashboard/firebase/admin/csp.template.json
```diff
@@ -92,20 +99,28 @@
...
+    "data:"
+  ],
   "connect-src": [
     "'self'",
     "https://*.analytics.google.com",
     "https://*.google-analytics.com",
     "https://*.googleapis.com",
     "https://*.googletagmanager.com",
     "https://*.sentry.io",
+    "https://cdn.jsdelivr.net",
```
maximilianoertel: What actually fetches from `jsdelivr` via XHR? `connect-src` is for XHR/WebSockets/etc., not for loading scripts or styles – those are covered by `script-src`/`style-src`

apps/dashboard/firebase/admin/csp.template.json
maximilianoertel: Separately, on the jsdelivr entry itself: whole-host https://cdn.jsdelivr.net whitelisting is a no-go. 

With this change, you'd be whitelisting script execution for anything anyone has ever published – an attacker who finds an injection point doesn't need to compromise a CDN, they just npm publish their payload and load it from a legit, allowlisted host.

We already do this correctly for katex on line 22 (`/npm/katex@0.16.8/`) – let's pin whatever package this is for the same way, including the version

apps/dashboard/vite.config.js
```diff
@@ -136,6 +136,12 @@ const buildFirebaseConfig = (mode = 'development') => {
   if (!header) throw new Error('CSP header not found');
   header.value = cspPolicy;
 
+  // Enforce strict CSP in staging and production, Report-Only elsewhere
+  // Uncomment to enforce strict CSP on staging and production
+  // if (['staging', 'production'].includes(mode)) {
+  //   header.key = 'Content-Security-Policy';
+  // }
+
```
maximilianoertel, line 144: What's this block for, given we want to enforce across all environments? If that's the goal, there's nothing to branch on – we can rename the key directly in firebase.template.json to Content-Security-Policy?

maximilianoertel (changes requested):
Thanks for taking care of this. The PR description says this is "in response to a review of our security" but no ticket is linked. Could you please link this PR to a ticket with the corresponding detailed review findings? Thanks

Emily-ejag (approved):
Rechecking this, it has the link we need for read-aloud, thanks

apps/dashboard/vite.config.js
kellyel: This logic was added to that we can continue to use 'Report-Only' in the dev environment, so that we can accumulate Sentry reports in dev. This won't be relevant post-BERF though, so I could see the argument for removing this block.

apps/dashboard/firebase/admin/csp.template.json
kellyel: All URLs added by this PR are a result of reports being made to Sentry for the dashboard. I've reviewed the dashboard for possible reasons that `blob:` was reported here, and my best guess is that this report occurred during some short lived work with web workers via the dashboard. I think the best approach is removing it here and monitor for further reports before including it again.

Emily-ejag (approved):

PR #1836, opened 2026-06-11
Bump Node to 24 LTS, regenerate lockfile, and fix vitest conditional-expect violations

richford:
## Summary

This PR brings the toolchain back to health in three coupled moves: bumps the runtime off EOL Node, makes the lockfile reproducible (regenerated once, under the new Node's npm), and fixes the lint violations the refreshed plugin versions surfaced.

**The runtime problem.** `.nvmrc` pinned `lts/iron` (Node 20), which reached end-of-life on 2026-04-30 — no more security patches. Node 22 (Jod) is already in maintenance (EOL 2027-04-30), so this PR jumps to **Node 24 "Krypton"**, the current active LTS (maintenance from 2026-10, EOL 2028-04-30), avoiding a repeat bump in ten months.

Changes: `.nvmrc` → `lts/krypton` (all CI jobs follow automatically via the shared `setup-node-environment` action's `node-version-file`); `apps/backend/Dockerfile` build stages `node:20-slim` → `node:24-slim` and the runtime stage `gcr.io/distroless/nodejs20-debian12:nonroot` → `gcr.io/distroless/nodejs24-debian12:nonroot` (the Node 20 distroless image is also EOL); `assessment.Dockerfile` `node:22*` → `node:24*` for consistency; `package.json` `engines.node` `>=20` → `>=22` (floor excludes EOL versions while still admitting maintained Jod). Note Node 24 bundles npm 11, which writes the same `lockfileVersion: 3` — the `packageManager: npm@10.8.2` pin is only enforced by corepack and can be bumped separately if/when corepack is adopted.

**The lockfile problem.** `package-lock.json` on `project/backend-refactor` was missing `resolved`/`integrity` metadata on many entries (e.g. `@vitest/eslint-plugin` pinned `1.3.20` with no registry URL). npm treats such entries as incomplete: any fresh `npm install` re-resolves the affected semver ranges to their *latest* matching versions and rewrites the lockfile. In other words, fresh installs were not reproducing CI's dependency tree — every new clone/worktree silently drifted. This PR commits one deliberate regeneration (npm 10, full `resolved`/`integrity` on every entry), so subsequent `npm ci` / `npm install` runs are deterministic.

**The lint fallout.** The regeneration legitimately bumps in-range dev dependencies; notably `@vitest/eslint-plugin` 1.3.20 → 1.6.20, whose `vitest/no-conditional-expect` rule now flags `expect` calls inside conditional branches. That found **188 pre-existing violations**: 11 in the assessment SDK's receiver tests, 154 across 17 backend test files, and 23 across 6 dashboard test files. These are real test smells, not rule noise — a conditionally-executed assertion can silently never run, and several genuinely were dead (see below).

### `package-lock.json`

Regenerated once with npm 10 (`packageManager: npm@10.8.2`). Large diff (+27k/−18k lines) but mechanical: `resolved`/`integrity` fields added throughout, in-range version refreshes. No `package.json` dependency declarations changed anywhere in the monorepo.

### `packages/assessment-sdk/src/receiver/roar-api.test.ts` (2 violations)

The constructor-throws test used `expect` inside `try/catch`. Now captures the thrown error and asserts unconditionally — if the constructor stops throwing, `thrownError` stays `undefined` and `toBeInstanceOf` fails (previously `expect.fail` covered this, but the instance/code assertions only ran inside `catch`).

### `packages/assessment-sdk/src/receiver/roar-api.integration.test.ts` (9 violations)

The ts-rest envelope-narrowing pattern `if (status === X && 'data' in body) { expect(...) }` made assertions skippable: a wrong response shape would pass the earlier status assertion's sibling checks but silently skip the body assertions. Each is converted to the guard-throw pattern the file already used elsewhere:

```ts
expect(response.status).toBe(201);
if (response.status !== 201 || !('data' in response.body)) {
  throw new Error('Expected a 201 response with a data envelope');
}
expect(response.body.data?.id).toBeDefined(); // now unconditional
```

TypeScript narrowing is preserved (the guard performs the same discriminated-union narrowing the `if` did), and assertions now always execute.

### `apps/backend` test files (154 violations, 17 files)

Controllers (`administrations`, `groups`, `task-bundles`, `task-variants`, `tasks`), services (`authorization`, `district`, `school`, `report`, `run-event`, `scoring`, `validations.utils`, `extract-fga-object-id.helper`), and integration suites (`users`, `reports`, `task` routes; `district.repository`). Four rewrite patterns, all semantics-preserving:

1. **ts-rest envelope narrowing** → assert-then-guard-throw (same as the SDK fixes above).
2. **`try { … expect.fail() } catch (e) { expect(e)… }`** → capture the thrown error, assert unconditionally afterward. Used heavily in the authorization service tests; assertions are strictly stronger (a non-throw now fails `toBeInstanceOf` instead of relying on `expect.fail`).
3. **Expects inside `if` over collection items** → select the data first (`filter`/`find`), assert the selection is non-empty/defined, then assert unconditionally.
4. **Parameterized tests branching on a should-throw flag** → split/restructure so each branch's assertions are unconditional.

**Deliberately strengthened (previously silently-skippable or dead) assertions** — these now require the seeded fixtures to actually contain the data the tests claim to verify, which is the point:

- `users.integration.test.ts`: row-shape / no-embed-default / `embed=stats` / `embed=tasks` assertions were skipped entirely when `items` was empty or `find()` missed; now gated by `toBeGreaterThan(0)` / `toBeDefined()`.
- `reports.integration.test.ts`: row-shape and per-task-shape blocks likewise now assert non-empty result sets.
- `run-event.service.test.ts` ("includes metadata in error context"): was doubly dead — passed if nothing threw or if a non-`ApiError` threw; now asserts the throw and its context unconditionally.
- `district.repository.integration.test.ts` ("applies sorting"): the pairwise-order loop was skipped if fewer than 2 rows came back; now asserts the count first.

If any of these surface a fixture gap in CI, that's the rule doing its job — the prior code would have passed vacuously.

### `apps/dashboard` test files (23 violations, 6 files)

- **Five mutation tests** (`useAddTaskMutation`, `useDeleteAdministrationMutation`, `useUpdateTaskMutation`, `useUpdateUserMutation`, `useUpsertAdministrationMutation`) shared an identical `try { await mutateAsync(...) } catch (error) { expect(...) }` block — silently green if the mutation unexpectedly succeeded. Converted to the capture pattern; the failure-path assertions (error identity, `isSuccess`/`isError` flags, no query invalidation) now run unconditionally.
- **`PdfExport.service.test.js`** had three vacuous expects inside `.catch()` callbacks of `Promise.race` calls where both timeout and rejection were documented as acceptable. Replaced with explicit swallows plus a real unconditional assertion (`expect(promise).toBeInstanceOf(Promise)`) so the "handles gracefully" tests still assert something (`vitest/expect-expect` enforces this).

### Type-check fallout from the refreshed dependency tree (15 backend files + root `package.json`)

The dependency refresh also moved `@types/express` (→ 5.0.6), TypeScript (→ 5.9.3), and axios (→ 1.17.0), with two `check-types` consequences:

- **45 `@ts-expect-error - Express v4/v5 types mismatch` directives became unused** (TS2578) and are removed. Note this is a *partial* cleanup by design: the route files carried ~71 of these directives, and the other ~26 still suppress a live mismatch between `@ts-rest/express@3.52.1` (typed against Express 4; `peerDependencies: express ^4.0.0`) and the repo's `@types/express@5` — those stay, exactly as the self-cleaning `@ts-expect-error` convention intends. They'll fall away when ts-rest ships Express 5 typings.
- **`fga.client.ts` hit a dual-axios type identity error**: root axios resolved to 1.17.0 while `@openfga/sdk` pins exactly 1.16.0, giving npm no choice but to nest a second copy. Fixed with a root `overrides: { "axios": "1.16.0" }`, which forces a single tree-wide copy at the SDK's pin — eliminating both the type error and the subtler runtime hazard of two axios module instances feeding the FGA authorization client. Revisit when `@openfga/sdk` relaxes its pin.

## Test plan

- [x] `eslint` over all five lint workspaces (backend, dashboard incl. cypress, roar-pa, api-contract, assessment-sdk), run per-directory with directly captured exit codes, against the regenerated lockfile's plugin versions — 0 errors (pre-existing warnings only, e.g. `vitest/no-disabled-tests` in `users.integration.test.ts:596`, out of scope)
- [x] `vitest run --project unit` for apps/backend — **2081 passed, 0 failures**
- [x] `vitest run` for apps/dashboard — **993 passed, 0 failures** (full suite at this branch's state)
- [x] `vitest run` for assessment-sdk — 99 passed, integration suite skipped without `RUN_INTEGRATION_TESTS` (as designed)
- [x] `tsc --noEmit` for apps/backend against the refreshed tree — clean except the dual-axios error, which resolves once the `overrides` entry is installed (`npm install` after this change updates the lockfile and dedupes axios)
- [x] `tsc --noEmit` for assessment-sdk — clean (the guard-throws preserve narrowing)
- [x] `prettier --check` on all 25 changed test files — clean
- [x] On a dev machine: `nvm install lts/krypton && nvm use` (Node 24 + stock npm 11), regenerate the lockfile with `rm -rf node_modules && npm install`, then `npm run lint && npm run check-types && npm run test` at repo root
- [ ] Backend + dashboard preview deploys build and boot on the Node 24 images
- [ ] CI: backend integration tests (real Postgres) cover the rewritten integration assertions, incl. the deliberately strengthened ones; the `e2e-tests` job covers the SDK integration suite with `RUN_INTEGRATION_TESTS=true`

References https://github.com/yeatmanlab/roar-project-management/issues/1881.

## Types of changes

- [ ] Documentation Update
- [x] Repository Maintenance


richford:
@claude, please review this PR

richford:
@claude, please re-review

richford:
@claude, please re-review

ksmontville (approved):

apps/dashboard/package.json
```diff
@@ -5,7 +5,7 @@
   "type": "module",
   "scripts": {
     "dev": "vite --host",
-    "build": "vite build --mode ${MODE:-development}",
+    "build": "NODE_OPTIONS=--max-old-space-size=8192 vite build --mode ${MODE:-development}",
```
maximilianoertel, line 8: Was this added to fix a specific OOM failure? What was the exact error, and was it in CI or local?

maximilianoertel (approved):

apps/dashboard/package.json
richford: It was on CI. I'm flying now but can try to look it up when I'm back home.

PR #1838, opened 2026-06-11
Migrate task read composables to the backend API

richford:
## Summary

This PR is the first in a four-PR chain that migrates the tasks-and-variants domain from `roarfirekit` / Firestore REST to the ts-rest backend client (see [issue 1881](https://github.com/yeatmanlab/roar-project-management/issues/1881)). It cuts the task **read** path over: `useTasksQuery` and `useTasksDictionaryQuery` now call `client.tasks.list()` via `getRoarApiClient()`, and every consumer that read `task.publicName` / `task.technicalName` from a backend-derived tasks dictionary now reads `task.nameSimple` / `task.nameTechnical` to match the new contract shape.

Mutations and the bundle / variant read flows remain on their legacy paths for now; they migrate in PRs B–D.

## Scoping decisions

- **`useTaskBundlesQuery` is descoped** (stays on firekit, deferred to PR D). `GET /v1/task-bundles/` and the cross-task `GET /v1/task-variants/` are still **super-admin-only** on the current base. Migrating bundles now would 403 CreateAdministration for platform/district admins. The relaxation policy (any-authenticated vs. platform-admins-only) is still TBD.
- **Variant reads are deferred to PR C** (matching the old shipped PR A, not the handoff summary): ManageVariants' update flow clones the variant read shape directly into the firekit mutation payload, so flipping reads to the new flat shape while mutations stay on firekit would corrupt updates. Reads and mutations for variants move together in PR C.

### Composable migrations (3 files)

- **`useTasksQuery.js`** — calls `client.tasks.list()` at the contract's max page size and **follows the response pagination**, aggregating all pages so a catalog that outgrows 100 tasks is never silently truncated. Gated on `authStore.accessToken` via `computeQueryOverrides`, unwraps the success envelope, throws on non-200 so TanStack routes the error through `error`. Uses `StatusCodes` constants. The legacy positional args (`registeredTasksOnly`, `taskIds`) are gone — the contract has no `?registered` filter (concept retired) and no `?ids=` filter (consumers filter client-side via the new `helpers/taskIdentifiers.js`).
- **`useTasksDictionaryQuery.js`** — thin selector over `useTasksQuery`. Keys each task by **both** `id` (UUID) and `slug` (e.g. `'phonics'`) so legacy slug-shaped lookups keep working alongside UUID lookups.
[...]

PR #1839, opened 2026-06-11
Migrate task mutations and rewrite task forms

richford:
## Summary

This PR is the second in the four-PR chain migrating the tasks-and-variants domain to the ts-rest backend client. PR #1838 migrated the **read** path; this PR migrates the task **write** path: `useAddTaskMutation` and `useUpdateTaskMutation` now call `client.tasks.create` / `client.tasks.update`, and `CreateTaskForm.vue` / `UpdateTaskForm.vue` are rewritten against the contract's `CreateTaskRequestBodySchema` / `UpdateTaskRequestBodySchema`.

This also closes PR #1838's biggest interim degradation: the **"Update Task" submit button, disabled in #1838** (because the form read the new backend shape while the mutation still wrote legacy Firestore documents), is **re-enabled** — reads and writes now speak the same contract.

Variant mutations + the ManageVariants rewrite come in PR C; rules/Cypress/bundles/cleanup in PR D.

### Mutations (2 files + 2 test files)

- **`useAddTaskMutation.js`** — `client.tasks.create({ body })`, unwraps the 201 envelope (resolves to `{ id }`), throws a structured error (`status` + `body`) on non-201, invalidates `TASKS_QUERY_KEY` on success. No more firekit/auth-store dependency.
- **`useUpdateTaskMutation.js`** — `client.tasks.update({ params: { taskId }, body })`, 200 unwrap, same throw/invalidate pattern. **Signature change:** payload is now `{ taskId, body }` instead of the legacy flat task object (the two forms are the only callers; both updated).
- Both test files rewritten against a mocked `getRoarApiClient`: request shape, envelope unwrap, invalidation on success, structured throw + no-invalidation on failure (capture pattern, assertions unconditional).

### Shared constants (`constants/tasks.js`)

`TASK_SLUG_REGEX`, `TASK_SLUG_MAX_LENGTH`, `TASK_NAME_REGEX`, `TASK_NAME_MAX_LENGTH` mirror the contract's schema constraints so both forms validate inline (vuelidate `helpers.regex`) instead of surfacing backend 400s.

### `CreateTaskForm.vue` — full rewrite

- New required fields: `slug` (regex-validated, replaces `taskId`), `name`, `nameSimple`, `nameTechnical`.
- Optional: `description`, `image` (was `coverImage`), `tutorialVideo` — omitted from the body when empty (the strict schema rejects empty strings).
- One `TaskParametersConfigurator` editing the flat `taskConfig` (the "Task Configuration" vs "Game Parameters" split is gone).
- Dropped: `registered` / `testData` / `demoData` checkboxes (concepts retired) and the `external` / `taskURL` / URL-params flow (`TODO(external-tasks)` comment marks the decision point).
- 409 conflicts get a specific "slug already exists" toast — the one failure admins can self-serve.

### `UpdateTaskForm.vue` — full rewrite

- Task selector unchanged (dropdown over the PR #1838 backend catalog).
- `slug` shown disabled (immutable per contract); editable fields match CreateTaskForm; one `taskConfig` editor seeded from the task's `taskConfig` object.
- Submit builds a **diffed PATCH body** — only changed fields are sent; cleared nullable fields (`description`, `image`, `tutorialVideo`) are sent as `null` to clear. No-op submissions short-circuit with a "Nothing to update" toast (the contract rejects empty bodies).
- The firekit `restConfig?.()` init guards are gone — `useTasksQuery` self-gates on the auth token.
- Submit re-enabled behind `userCan(Permissions.Tasks.UPDATE)`.

### taskConfig integrity (post-review rework)

An independent review caught that the original form implementation reused the legacy `convertParamArrayToObject`, whose `camelCase()` silently rewrote any non-camelCase `taskConfig` key (e.g. `max_attempts` → `maxAttempts`) on every save — and defeated the diff guard, so even no-op submits corrupted config. The forms now use a new **key-preserving** helper module (`helpers/taskConfig.js`, unit-tested incl. a snake_case round-trip no-op):

- `splitTaskConfig` separates scalar entries (editable rows) from `null`/array/object entries, which the contract explicitly allows but the row editor can't represent. **Passthrough entries are preserved verbatim** and listed read-only in the UI; previously a single `null` value made the form unsubmittable with no visible error. A `taskConfig` that isn't a plain object disables the editor entirely instead of corrupting on save.
- `buildTaskConfigFromRows` uses row names verbatim as keys. New-row names are format-validated (`IDENTIFIER_WITH_UNDERSCORES` mirror) so junk keys (e.g. `camelCase('!!!') === ''`) can't be created; existing backend keys are exempt (their name inputs were already disabled in edit mode).
- The update diff compares `taskConfig` with deep equality (`lodash/isEqual`), not `JSON.stringify`, so key order can't produce false-positive diffs.
- The form-seeding watch is keyed on `selectedTaskId` rather than the computed task object, so background catalog refetches (window-focus, or a create in the adjacent tab invalidating `['tasks']`) no longer wipe in-progress edits.
- Also from review: `description` gets `maxLength(1024)` + trimming (whitespace-only treated as blank/cleared); the update failure toast surfaces the backend's error message when present; `mutate` calls no longer misleadingly `await`ed; Create's config editor starts with zero rows so an empty `taskConfig` doesn't require deleting a seed row.

## UX changes

- Create form: slug + three name fields; single config editor; legacy checkboxes and the external-task flow are gone. Slug validation is stricter than legacy `taskId` (lowercase-hyphenated, ≤32 chars) — matching the backend's persistence model.
- Update form: works again (was disabled in PR #1838); now edits `nameSimple` / `nameTechnical` / `tutorialVideo`; `gameConfig` nesting is gone. Existing tasks' legacy `gameConfig` keys are not auto-merged into `taskConfig` — admins editing pre-migration data re-add keys if needed.
- Both endpoints are super-admin-only on the backend; the permission-gated buttons mirror that.

## Test plan

- [x] Dashboard suite: **114 files, 1016 passed, 0 failures** (incl. 6 rewritten mutation tests + 9 `taskConfig` helper tests)
- [x] eslint / prettier clean on all touched files
- [x] Repo-root `npm run lint && npm run check-types && npm run test` on a dev machine
- [ ] Manual smoke (preview deploy): create task with slug `pr-b-smoke-test` (expect 201 + appears in dropdowns without reload); edit its `nameSimple` + a `taskConfig` key (expect 200 + persists); resubmit unchanged (expect "Nothing to update", no request); create with a duplicate slug (expect the 409 conflict toast). **Critically, also edit a task whose `taskConfig` has snake_case keys and/or `null`/array values** (expect: keys unchanged after save, non-scalar entries listed as preserved, no-op resubmit short-circuits).

## Chained PRs

| # | Status | Scope |
|---|--------|-------|
| maint | in review | Node 24 + lockfile + lint/test health |
| A | in review | Task read composables + field renames |
| **B** | **this PR** | Task mutations + Create/Update form rewrites |
| C | next | Variant mutations + ManageVariants rewrite (re-enables Create Variant) |
| D | last | AI rules + Cypress + bundles (needs backend authz relaxation) + retire `helpers/query/tasks.js` |

References https://github.com/yeatmanlab/roar-project-management/issues/1881; `Resolves` is reserved for PR D.


richford:

richford:
@claude, please re-review

apps/dashboard/src/containers/ManageTasks/components/TaskParametersConfiguratorRow.vue
```diff
@@ -131,6 +131,14 @@ const rules = {
...
+      // taskConfig keys (no camelCasing), so junk names must be rejected up front.
+      // Existing backend keys are unconstrained server-side and stay editable as-is
+      // (their name inputs are disabled anyway in edit mode).
+      nameFormat: helpers.withMessage(
+        'Must start with a letter and contain only letters, numbers, and underscores',
+        (value) => !row.value.isNew || !value || TASK_PARAMETER_NAME_REGEX.test(value),
+      ),
     },
     type: { required },
     value: { required },
```
Emily-ejag: I wonder if this required validator is a little too strict here.

Because `taskConfig` is arbitrary JSON on the backend, an existing task could validly have a value like `{ note: '' }`. In that case, Vuelidate would mark the row as invalid and block saving the whole form, even for unrelated edits.

Could we allow empty strings here and only treat null / undefined as missing?

I think is a mismatch:

Backend says: empty string is allowed.
Frontend says: empty string is invalid.
Result: valid backend data breaks the edit form.

apps/dashboard/src/constants/tasks.js
```diff
@@ -9,3 +9,21 @@ export const TASK_PARAMETER_DEFAULT_SHAPE = {
...
+export const TASK_SLUG_MAX_LENGTH = 32;
+export const TASK_NAME_REGEX = /^[a-zA-Z][a-zA-Z0-9_\- ]*$/;
+export const TASK_NAME_MAX_LENGTH = 255;
+export const TASK_DESCRIPTION_MAX_LENGTH = 1024;
+
+/**
+ * Format for newly created taskConfig parameter names, mirroring the contract's
+ * IDENTIFIER_WITH_UNDERSCORES. Only applied to NEW rows — existing backend keys
+ * are unconstrained server-side and must remain loadable/editable as-is.
+ */
```
Emily-ejag, line 33: This says the regex mirrors `IDENTIFIER_WITH_UNDERSCORES`, but I think that contract applies to task variant parameter names, not arbitrary task `taskConfig` keys. Since taskConfig keys are accepted by the backend more broadly, this UI validation may reject valid keys like dotted or hyphenated ones.

Could we either make this a clearly intentional UI-only constraint, or relax the regex to match what the backend allows?

apps/dashboard/src/containers/ManageTasks/components/UpdateTaskForm.vue
```diff
@@ -13,38 +13,94 @@
...
         <div>
-          <legend class="text-lg font-medium mb-0">Game Parameters</legend>
-          <p class="text-md text-gray-500 mt-2">Create the game parameters for variants of this task.</p>
+          <legend class="text-lg font-medium mb-0">Task Configuration</legend>
+          <p class="text-md text-gray-500 mt-2">Adjust the configuration for this task.</p>
         </div>
 
-        <TaskParametersConfigurator v-model="taskGameParameters" edit-mode :validation-key-blacklist="ignoreFields" />
+        <template v-if="canEditTaskConfig">
+          <TaskParametersConfigurator v-model="taskConfigModel" edit-mode />
```
Emily-ejag: I think we may need to pass the passthrough keys as a blacklist here.

Right now, the UI shows the passthrough keys as preserved, but it does not prevent a user from adding a new scalar row with the same name.

Could we pass :validation-key-blacklist="taskConfigPassthroughKeys" to TaskParametersConfigurator in edit mode so those preserved keys cannot be reused?

```suggestion
          <TaskParametersConfigurator
  v-model="taskConfigModel"
  edit-mode
  :validation-key-blacklist="taskConfigPassthroughKeys"
/>
```

apps/dashboard/src/containers/ManageTasks/components/UpdateTaskForm.vue
```diff
@@ -131,40 +159,133 @@ const formattedTasks = computed(() => {
...
+ * @returns {Object} The PATCH body containing only changed fields.
+ */
+function buildDiffedBody(task) {
+  const body = {};
+
+  for (const field of ['name', 'nameSimple', 'nameTechnical']) {
+    if (formModel[field] !== (task[field] ?? '')) {
+      body[field] = formModel[field];
+    }
+  }
```
Emily-ejag, line 264:
```suggestion
for (const field of ['name', 'nameSimple', 'nameTechnical']) {
  const trimmed = formModel[field].trim();

  if (trimmed !== (task[field] ?? '')) {
    body[field] = trimmed;
  }
}
```

Emily-ejag (changes requested):

ksmontville (dismissed):

richford:
@claude, please re-review

richford:
@claude, please re-review

Emily-ejag (approved):

PR #1843, opened 2026-06-11
Refresh tasks-domain AI rules and add backend-only Cypress smoke coverage

richford:
## Summary

This PR is the fourth dashboard PR in the chain for [issue 1881](https://github.com/yeatmanlab/roar-project-management/issues/1881), slimmed to documentation, e2e, and test hardening:

- **AI rules refresh** — rewritten `.ai/rules/frontend-composable-patterns.md` (real T11 excerpts: `useTaskVariantsByTaskQuery` for queries, `useUpdateTaskVariantMutation` for 204 mutations; token gating, envelope unwrap, structured throws, retry placement after `...options`, prefix invalidation, test pattern), new `.ai/rules/frontend-backend-api-client.md` (`getRoarApiClient()` lazy init, 401 refresh retry, response/error shape table, `embed` wire format, when NOT to use the client, the `process.env.CYPRESS` `manualChunks` guard), and the `README.md` index update.
- **Cypress** — new backend-only smoke spec `cypress/e2e/smoke/tasks.cy.js` per `frontend-e2e-testing-pattern.md`: emulator sign-in, envelope/flat-shape/published-only assertions on `/v1/tasks` and `/v1/tasks/:taskId/variants` as a teacher, 403s on `/v1/task-variants` + `/v1/task-bundles` for teacher and district-admin tiers, 401s unauthenticated. **Gap:** `CYPRESS_FIXTURE_USER_KEYS` seeds no super/platform admin, so the catalog 200 path isn't asserted yet — documented in the spec.
- **Retry-branch tests** — `useTaskVariantsByTaskQuery.test.js` gains both retry branches (terminal-error short-circuit; transient retries < `MAX_RETRIES`), which it previously didn't cover.
- **Comment cleanup** — `HomeParticipant.vue`: the `TODO(1881)` above the GameTabs `external`/`taskURL`/`meta` mapping becomes a permanent comment explaining the defensive `taskConfig` mapping.

## Deferred: TaskPicker catalog cutover

An earlier draft of this PR also migrated `useTaskBundlesQuery` and the cross-task `useTaskVariantsQuery` to the ts-rest client and rewired the `CreateAdministration` → `TaskPicker` → `TaskBundleCard`/`VariantCard`/`EditVariantDialog` surface to the flat contract shapes. That cutover is **deliberately deferred**:

[...]

PR #1840, opened 2026-06-11
Migrate variant mutations and rewrite ManageVariants

richford:
## Summary

Third PR in the [issue 1881 chain](https://github.com/yeatmanlab/roar-project-management/issues/1881):

- PR #1838 migrated task reads
- PR #1839 migrated task writes
- This PR migrates the **variant** domain: both variant mutations move to the ts-rest client, a new per-task variants query replaces the cross-task firekit read for the admin surface, and `ManageVariants.vue` is fully rewritten around the task-first UX — pick a task, filter its variants by publication status, then create or edit against the flat contract shape.

This closes PR #1838's remaining interim degradations on this page: the **"Create Variant" submit, disabled since #1838**, is re-enabled, and the "Show Only Registered Variants" checkbox is replaced by the status model (`draft | published | deprecated`).

### Composables (3 new/rewritten + 3 test files)

- **`useTaskVariantsByTaskQuery.js` (new)** — `GET /tasks/:taskId/variants` with optional status filter; paginated aggregation, token+taskId enablement gating, canonical retry policy. Query key: `[TASK_VARIANTS_QUERY_KEY, taskId, status]`.
- **`useAddTaskVariantMutation.js`** — `POST /tasks/:taskId/variants`; payload `{ taskId, body }` (taskId must be the UUID per the contract's create path param); 201 unwrap to `{ id }`.
- **`useUpdateTaskVariantMutation.js`** — `PATCH /tasks/:taskId/variants/:variantId`; payload `{ taskId, variantId, body }`; **treats 204 No Content as success** (resolves to `undefined`; callers rely on invalidation refetch).
- Both mutations invalidate `[TASK_VARIANTS_QUERY_KEY]` — the prefix covers the new per-task query *and* the legacy cross-task query still feeding CreateAdministration. The legacy `TASKS_QUERY_KEY` invalidation is dropped: variant writes no longer touch task documents.
- 10 new tests across the three composables (key composition, status param pass/omit, pagination, 201/204 handling, structured throws, invalidation).

### `ManageVariants.vue` — full rewrite (task-first)

- **Create Variant**: task dropdown (UUID-keyed — the slug interim hack from PR A is gone since the mutation now wants UUIDs) → optional `name` (contract regex/length mirrors) + optional `description` + required `status` (default `draft`) + parameters editor. The legacy demo/test/registered checkboxes, duplicate-name client check, and gameConfig prefill are gone — replaced by the contract model.
[...]

PR #1846, opened 2026-06-11
Relocate TaskParametersConfigurator to shared components and split ManageVariants into child forms

richford:
## Summary

This PR completes two structural refactors that reviews of both PR #1840 and PR #1843 requested but that were deferred to keep those diffs focused:

1. **Relocate `TaskParametersConfigurator`** out of `containers/ManageTasks/components/` into the shared `src/components/TaskParametersConfigurator/` directory. The configurator stopped being a ManageTasks-private component the moment `ManageVariants` (PR #1840) started importing it via `@/containers/ManageTasks/components/...` — a cross-container reach that violates the container/presentational conventions in `frontend-layer-architecture.md`. The component, its row child, and both colocated Cypress component specs move verbatim (content unchanged; the only import that needed adjusting was nothing — all internal imports were already relative or aliased). `CreateTaskForm`, `UpdateTaskForm`, and the new variant form components now import from the shared path.
2. **Split `ManageVariants.vue`** into the same parent/child shape PR #1839 established for `ManageTasks`: a thin parent owning only the `PvSelectButton` view toggle plus two new child components, `components/tasks/components/CreateVariantForm.vue` and `components/tasks/components/UpdateVariantForm.vue`. Each child owns its complete form — its own `useVuelidate` instance, queries (`useTasksQuery`, and for the update form `useTaskVariantsByTaskQuery`), mutation, models, and submit handler. Code moved verbatim wherever possible; `PvToast` moves into each child, mirroring `CreateTaskForm`/`UpdateTaskForm`.

## Why the split also deletes a workaround

The single-file version carried a documented vuelidate caveat: both forms lived in one component, so vuelidate registered the configurator rows' nested validators with **every** `useVuelidate` instance in that component, and the inactive form's invalid rows could block the visible form's submit. The file worked around it by toggling forms with `v-if` (not `v-show`) and explaining why in a comment.

With each form in its own component, the two `useVuelidate` instances live in separate component scopes and row registration is naturally isolated — the workaround comment and the cross-form caveats are gone. The parent still uses `v-if` (no reason to keep the inactive form mounted), but `v-if` vs `v-show` no longer has validation implications.

## Shared task selection preserved

The old single file shared `selectedTaskId` between the create and update views, so picking a task in one view pre-selected it in the other. That UX is preserved by lifting `selectedTaskId` into the thin parent and passing it to each child via `v-model:selected-task-id` (`defineModel` in the children). Everything else is per-form state, as before.

## Behavior preserved (verbatim moves)

- Update view's status filter with the `'all'` string sentinel (PvSelect null-model quirk; sentinel maps to omitting the status query param)
- Passthrough parameter blacklist on the update configurator (prevents duplicate-name 409s) and the "preserved exactly as-is" notice
- `updateNameValidator` changed-name exemption for legacy variant names
- `buildVariantPatchBody` diffing with the "Nothing to update" short-circuit
- Create form's `paramsModel` clearing on task switch
- `userCan(Permissions.Tasks.CREATE/UPDATE)` submit gates with tooltips
- All toasts (warning/success/error/info) including backend-message passthrough
- The unscoped `.submit-button` / `.select-button` styles stay in the parent (they're global, exactly as before)

## Files

**Moved (old files listed for `git rm`):**

| From `containers/ManageTasks/components/` | To `components/TaskParametersConfigurator/` |
| --- | --- |
| `TaskParametersConfigurator.vue` | `TaskParametersConfigurator.vue` |
| `TaskParametersConfiguratorRow.vue` | `TaskParametersConfiguratorRow.vue` |
| `TaskParametersConfigurator.cy.js` | `TaskParametersConfigurator.cy.js` |
| `TaskParametersConfiguratorRow.cy.js` | `TaskParametersConfiguratorRow.cy.js` |

**New:** `components/tasks/components/CreateVariantForm.vue`, `components/tasks/components/UpdateVariantForm.vue`

**Changed:** `components/tasks/ManageVariants.vue` (now a thin parent), `containers/ManageTasks/components/CreateTaskForm.vue` + `UpdateTaskForm.vue` (import path only)

## Test plan

- [x] `eslint` on every new/changed file — exit 0
- [x] `prettier --check` on every new/changed file — exit 0, all clean
- [x] `npx vitest run` (apps/dashboard, 4 shards) — 117 files, 1047 passed, 2 skipped, 0 failed
- [x] Grep: zero imports of the old `containers/ManageTasks/components/TaskParametersConfigurator` path remain (outside the to-be-removed files themselves)
- [x] Repo-root `npm run lint` / `format:check` / `check-types` / `test` on a dev machine
- [ ] CI Cypress component tests pick up the relocated `.cy.js` specs at the new path

## Chained PRs

| #       | Status      | Scope                                                                                                   |
| ------- | ----------- | ------------------------------------------------------------------------------------------------------- |
| maint   | in review   | Node 24 + lockfile + lint/test health                                                                   |
| A       | in review   | Task read composables + field renames + dead-code prune                                                 |
| B       | in review   | Task mutations + Create/UpdateTaskForm rewrites                                                         |
| C       | in review   | Variant mutations + per-task variants query + ManageVariants rewrite                                    |
| backend | in review   | Relax `/v1/task-variants` + `/v1/task-bundles` list authz to super admin OR platform admin              |
| D       | in review   | AI rules refresh + backend-only Cypress smoke + retry-branch tests + comment cleanup                    |
| **E**   | **this PR** | Relocate `TaskParametersConfigurator` to shared components + split `ManageVariants` into child forms    |

References https://github.com/yeatmanlab/roar-project-management/issues/1881 — the issue stays open; the final `Resolves` belongs to the TaskPicker catalog cutover follow-up issue.

## Independent review fixes

Two adjustments from an independent review of this PR (superseding the `v-if`/per-child-toast details above):

- **`v-show` instead of `v-if` for the child forms.** Keeping both forms mounted preserves in-progress work (drafts, the update view's status filter, variant selection) across view toggles — restoring pre-split behavior. This is safe now precisely because of the split: vuelidate registers configurator rows only within each child's own component tree, so the old single-component cross-registration hazard no longer applies.
- **Single `<PvToast />` outlet in the parent.** The per-child `<PvToast />` elements are removed (with both children now always mounted under `v-show`, duplicate outlets would render every toast twice); the children keep `useToast()`, which posts to the parent's outlet.


richford:
@claude please review this PR

ksmontville (dismissed):

richford:
@claude, please re-review

richford:
@claude, please re-review

ksmontville (approved):

PR #1854, opened 2026-06-16
`ref` Configure Assessment Scoring in Monorepo

ksmontville:
### Normed SWR English, grade 7, age 12
image

### Non-normed SWR Italian, grade 2, age 5
image

### Normed PA, Adaptive v5
image

## Summary
This PR configues the individual assessments to write scores to the backend database (local, ephemeral, or persistent).

For each assessment, the `assessment-schema` package is expanded to describe the score names and score entries that assessment trials write out. A firekit-facade-adapter is configured for each assessment which handles these scoring details in a single module. These facades export a function `wireScoreAdapter` which is used in `config.js` to configure scoring for the assessment.

Some shared constants in `assessment-schema` are also introduced and used throughout the package and in the individual assessments where universal and/or repeated string literals are used frequently; for example, "test" and "practice" assessment stages, "raw" and "computed" score types, etc.

Score reporting services were expanded to become "domain-aware", to facilitate assessment scores which re-use score names across domains (e.g. domain "FSM" with score name "numCorrect"). These changes are backwards compatible with the pre-existing "flat" score structure.

## Types of changes

- [ ] Documentation Update
- [ ] Repository Maintenance

## Checklist


ksmontville:
@claude Let's review this PR. In particular, look for opportunities to simplify scoring configs or plumbing which can be refactored into shared utilities or other improvements.

ksmontville:
@claude Feedback implemented, let's review.

Regarding item 4, the score type mismatch between roar-pa and roar-swr is know and intentional until we can align the research team on a singular type; accept for now.

Also leaving item 5 in place with a comment explaining why,

ksmontville:
@claude Feedback addressed, ready for another review.

ksmontville:
@claude Done, let's review again.

Emily-ejag (dismissed):
LGTM

packages/assessment-schema/src/roar-pa/score-names.ts
```diff
@@ -1,44 +1,68 @@
...
+  THETA_SE_RAW: 'thetaSERaw',
 } as const;
 
 export type PaScoreName = (typeof PA_SCORE_NAMES)[keyof typeof PA_SCORE_NAMES];
+
+/**
+ * Score names that map to type='raw' entries — live state captured directly
+ * from trial accumulation (counts, unscaled theta internals).
+ * thetaEstimate and thetaSE are type='computed': they are IRT-derived ability
+ * estimates, not raw trial observations.
```
richford: See my other comment on the `score-entries.ts` file as well.

These comments imply that IRT-derived ability estimates are computed but also note that the unscaled thetas are raw. I think the real distinction here is that "raw" values are in a "native scale" that the assessment owns (including IRT scoring on the native scale) and the computed scores are in a "shared scale" that the assessment shared with other assessments or with representative norms. So,

- **`thetaEstimateRaw` / `thetaSERaw` → RAW**: the IRT ability estimate (and its SE) in the assessment's **native theta scale**. This is the CAT engine's direct output (`clowder.theta.composite`).
- **`thetaEstimate` / `thetaSE` → COMPUTED**: that same estimate linearly transformed (`raw × transformation.scale + transformation.shift`, from the IRT hyperparameter table) onto a **scale shared across assessments** for reporting.

```suggestion
 * Scores fall into two types, "raw" and "computed," depending on whether they align with the assessment's "native" scale or a "shared" scale with other assessments on representative norms.
 * `type='raw'` = values in the assessment's native measurement space: trial counts, plus the native-scale IRT estimate (`thetaEstimateRaw`/`thetaSERaw`).
 * `type='computed'` = values derived for reporting: the shared-scale ability estimate (`thetaEstimate`/`thetaSE`, comparable across assessments) and normed lookups (percentile, standardScore).
```

You'll probably have to do some formatting for line length if you accept that suggestion.

packages/assessment-schema/src/roar-pa/score-entries.ts
```diff
@@ -1,217 +1,163 @@
...
+ * Score entry for PA scores written to run_scores.
+ * Mirrors the api-contract ScoreEntry discriminated union without importing it
+ * directly, keeping assessment-schema decoupled from api-contract at build time.
  *
- * Compile-time type check: If api-contract adds a required field to ComputedScoreEntry,
- * the _typeCheck assignment below will fail with a TypeScript error, alerting us to update
- * this interface. This ensures contract changes surface immediately at compile time.
- * @see packages/api-contract/src/v1/runs/schema.ts
+ * - type=RAW: live state captured from trial accumulation (counts, theta). assessmentStage required.
+ * - type=COMPUTED: derived values (percentile, standardScore, percentCorrect). assessmentStage required.
```
richford: See my comment in score-names about the distinction between raw and computed here.


```suggestion
 * - `type='raw'` = values in the assessment's native measurement space: trial counts, plus the native-scale IRT estimate (`thetaEstimateRaw`/`thetaSERaw`).
 * - `type='computed'` = values derived for reporting: the shared-scale ability estimate (`thetaEstimate`/`thetaSE`, comparable across assessments) and normed lookups (percentile, standardScore).
  ```

File: packages/assessment-schema/src/roar-pa/score-entries.ts [hunk 3]
```diff
@@ -1,217 +1,163 @@
...
- *
- * @param computed - Nested computed scores object from RoarScores.computedScoreCallback
- * @param options - Configuration options
- * @param options.strict - If true, throw on unrecognized input group keys; if false, silently skip them
- * @returns Array of ScoreEntry objects ready for backend upsert
- * @throws {Error} If strict=true and an unrecognized input group key is encountered
+ * Output: Array of PaScoreEntry objects where:
+ * - Subtask entries carry the uppercase domain (FSM, LSM, DEL) — never lowercase
+ * - Composite entries carry 'composite' or 'composite_foundational'
+ * - type is RAW for counts/theta, COMPUTED for normed/derived scores
```
richford:
```suggestion
 * - type is RAW for counts and unscaled theta, COMPUTED for scaled theta and normed/derived scores
```

packages/assessment-schema/src/roar-swr/score-entries.ts
```diff
@@ -0,0 +1,83 @@
...
+import { ScoreType } from '../enums/score-type.enum.js';
+import { SWR_SCORE_DOMAINS, SWR_SCORE_NAMES, SWR_RAW_SCORE_NAMES, type SwrScoreName } from './score-names.js';
+
+/**
+ * Score entry shape for SWR scores written to run_scores.
+ * Mirrors the api-contract ScoreEntry discriminated union without importing it
+ * directly, keeping assessment-schema decoupled from api-contract at build time.
+ *
+ * - type=RAW: live state captured per trial (theta, counts). assessmentStage required.
+ * - type=COMPUTED: normed/derived values (percentile, standardScore). assessmentStage
```
richford: I'm anchoring this comment here but it actually applies across `roar-swr/score-names.ts`, `roar-swr/score-entries.ts`, and `swr-firekit-facade.js`.

Some context on the theta scale for Word. ROAR Word actually defines the shared scale for the other assessments here. So the "native" scale **IS** the "shared" scale. That means that the `thetaEstimate` should be written as both a raw score and a computed score.

It looks like SWR writes the current theta only once as `thetaEstimate` with the `COMPUTED` type. `SWR_SCORE_NAMES` has only `THETA_ESTIMATE` (no `THETA_ESTIMATE_RAW`), and `SWR_RAW_SCORE_NAMES` = `{numAttempted, numCorrect, numIncorrect, percentCorrect}`. So `toSwrScoreEntries` emits a single computed `thetaEstimate` (= the native `cat.theta` from `swr-firekit-facade.js:62`). No raw theta entry is written.

To honor the model and match PA (which writes both `thetaEstimateRaw` (native/raw) and `thetaEstimate` (shared/computed)), SWR should write the same value as both:

1. Add `THETA_ESTIMATE_RAW: 'thetaEstimateRaw'` to `SWR_SCORE_NAMES`, and add it to `SWR_RAW_SCORE_NAMES` (→ `type=raw`).
2. Have the facade emit `thetaEstimateRaw: theta` alongside `thetaEstimate: theta` (same native value), so `toSwrScoreEntries` produces a raw entry (native scale) and a computed entry (shared scale) with equal values.
3. Update the doc comments accordingly: for SWR the native scale **is** the shared reporting scale, so `thetaEstimateRaw` (raw) and `thetaEstimate` (computed) carry the same value. Both are written so that every IRT scored assessment uniformly exposes a native-scale and a shared-scale theta. This also replaces the misleading "IRT-derived … not raw trial observations" rationale on `score-names.ts:37–39` and the `score-entries.ts:11` doc.

(Net effect: PA and SWR end up with the same `run_scores` shape for theta: `thetaEstimateRaw` as `raw`, `thetaEstimate` as `computed`. They just happen to be equal for SWR because SWR is the reference scale.)

packages/assessment-schema/src/score-utils.ts
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Score names shared across all ROAR assessments.
+ * Assessment-specific score names (e.g. PA_SCORE_NAMES, SWR_SCORE_NAMES) extend
+ * these with fields that only apply to their own scoring schema.
+ */
+export const COMMON_SCORE_NAMES = {
+  NUM_CORRECT: 'numCorrect',
+  NUM_ATTEMPTED: 'numAttempted',
+  NUM_INCORRECT: 'numIncorrect',
+} as const;
```
richford: `COMMON_SCORE_NAMES` is introduced as the shared base for `numCorrect`/`numAttempted`/`numIncorrect`, but `PA_SCORE_NAMES` (roar-pa/score-names.ts:36–38) and `SWR_SCORE_NAMES` (roar-swr/score-names.ts:28–30) re-declare the same literals rather than composing it. Consider spreading `COMMON_SCORE_NAMES` into the per-assessment maps so there's a single source of truth.

packages/assessment-schema/src/roar-pa/score-names.ts
```diff
@@ -1,44 +1,68 @@
+/**
+ * Canonical run_scores.domain strings for PA score entries.
+ * Subtask domains use uppercase to match the BigQuery schema convention.
+ */
+export const PA_SCORE_DOMAINS = {
```
richford, line 7: I get that the casing is intentional given your comment in the JSDoc string. But can you make it even more explicit. Like, "N.B., the casing inconsistency in deliberate." or something like that.

richford (changes requested):

apps/assessments/roar-swr/src/experiment/scores.js
richford, line 162: The schema (`SWR_SCORE_NAMES` + `SWR_RAW_SCORE_NAMES`), the facade, and the docs were all updated to write the raw theta as `thetaEstimateRaw` (type=raw) alongside `thetaEstimate` (type=computed) with equal values. But here at runtime SWR writes **only** `thetaEstimate`. Running the real `RoarScores.computedScoreCallback` and the real `toSwrScoreEntries` on the exact object the facade emits for a test trial:

- `computedScoreCallback` output
```
.composite = {"thetaEstimate":0.42}      // no thetaEstimateRaw key
```
- `toSwrScoreEntries(computed) =`
```
[{"name":"thetaEstimate","type":"computed"}]   // raw entry absent
```
- Expected `toSwrScoreEntries({thetaEstimateRaw:0.42, thetaEstimate:0.42}) =`
```
[
  {"name":"thetaEstimateRaw","type":"raw","value":"0.42"},
  {"name":"thetaEstimate","type":"computed","value":"0.42"}
]
```

Root cause: the SDK only persists `scoreAdapter(computedScores)`, the output of `computedScoreCallback`. SWR's `scores.js` builds `computedScore = { thetaEstimate: score, ...normed }` and never copies `thetaEstimateRaw`, so the `thetaEstimateRaw: theta` the facade puts in the raw input is consumed and discarded. (PA works because PA's `scores.js` sets `thetaEstimateRaw` in its composite output; SWR's wasn't given the equivalent line.)

Suggested fix:

```js
let computedScore = {
  thetaEstimateRaw: subtaskScores.test?.thetaEstimateRaw ?? score, // native === shared for SWR
  thetaEstimate: score,
};
```

And after that, can you add a test that asserts the callback output (not a hand-built `computed`) contains `thetaEstimateRaw`. The current unit tests feed `thetaEstimateRaw` straight into `toSwrScoreEntries`, so they pass while the real facade→`scores.js`→adapter flow drops it.

richford (changes requested):
Thanks @ksmontville! I have an inline change request below. And also this more general comment about how this interacts with the `recomputeUseForReporting` method:

`recomputeUseForReporting` (run.repository.ts:159) runs in the writeTrial transaction and re-ranks the user's runs. Within the two "incomplete" tiers the **primary** sort is lowest `thetaSE`, then highest `numAttempted`. It joins `run_scores` on:

```
type='raw', domain='composite', assessment_stage='test', name='thetaSE'   (SCORE_TYPE.RAW, SCORE_NAME.THETA_SE)
```

But under the new naming in this PR, **nothing writes `(type='raw', name='thetaSE')`**:

- PA writes the raw SE as `thetaSERaw` (type=raw) and `thetaSE` as type=computed.
- SWR writes no SE at all (`THETA_SE` isn't in `SWR_SCORE_NAMES`; per "don't report thetaSE in roar-swr computed scores").

So the `theta_se` join is NULL (`NULLS LAST`) → the primary SE tiebreaker is inert. For PA it silently degrades to the `numAttempted` secondary (which does match because PA writes `(raw, composite, numAttempted)`), then `created_at`. For normed (en/es) SWR test runs there's neither a raw SE nor a raw composite count, so incomplete runs fall through to `created_at`.

The recompute constants/query are unchanged by this PR. It introduced the `thetaSE`(computed)/`thetaSERaw`(raw) split, so producer and recompute drifted apart. The documented precedent in `run-scores.ts` (names mirror `runData.scores.raw.composite.test.*`) is that the **raw** SE was named `thetaSE`, which the new convention renames to `thetaSERaw`.

I can see two coherent fixes going forward. Please pick one of them and keep the producer and recompute in agreement:

1. **Point the recompute at the new raw name** (consistent with the "distinct names" decision): add `SCORE_NAME.THETA_SE_RAW = 'thetaSERaw'` and use it in the `theta_se` join, keeping `type='raw'`. Then PA's SE tiebreaker works again, and any IRT-scored assessment that wants SE-based best-run selection (including SWR) must emit `(raw, composite, test, thetaSERaw)`.
2. **Read the computed SE** instead (`type='computed', name='thetaSE'`). The ranking is identical since scaled SE is monotonic in raw SE. But this changes the "raw" semantics the constant documents.

Either way, SWR needs to emit *some* composite SE if its incomplete runs should rank by precision rather than recency. This compounds with the `thetaEstimateRaw` drop above: fixing that alone won't make the SE tiebreaker work, because the recompute keys on `thetaSE`, not `thetaEstimateRaw`.

richford:
Following up on my review above, I think this is also a change from how legacy scores were written. I think they were written as `thetaEstimate` in both the raw and the computed scores for Word. Can you verify, please?

Emily-ejag:
@ksmontville had asked me about `thetaSE` for SWR, and when I reviewed the legacy behavior, I thought it was intentional that SWR was not writing `thetaSE`, since I understood `thetaSE` as mainly being used to validate the quality/precision of the score.

Because of that, I had asked @ksmontville to keep the SWR scores as they currently are.

After chatting with @richford yesterday, we realized that SWR does need to write a `thetaSE`, and that this has actually been a bug. We need `thetaSE` for two things:

1. Selecting the best run for reporting.
2. The new equation for calculating the composite scores.

I can add this to legacy. @ksmontville, if you already made the changes in the monorepo, I can also add them there.

apps/assessments/roar-swr/src/sdk/swr-firekit-facade.js
```diff
@@ -0,0 +1,119 @@
...
+      const { cat } = experimentModule;
+      const theta = cat?.theta ?? null;
+      return {
+        composite: {
+          test: {
+            // SWR defines the shared IRT scale, so the native theta IS the shared
+            // theta. Write both so run_scores has thetaEstimateRaw (type=raw) and
+            // thetaEstimate (type=computed) with equal values, matching the PA shape.
+            // thetaSERaw (type=raw) is the native-scale SE used by recomputeUseForReporting;
+            // no computed thetaSE is written because SWR has no cross-scale SE transform.
```
richford: This says "no computed thetaSE is written because SWR has no cross-scale SE transform," but line 81 writes `thetaSE`, and `SWR_SCORE_NAMES` (in score-names.ts) includes `THETA_SE`. So a computed `thetaSE` is emitted (equal to `thetaSERaw`). Can you reconcile the comment with something like "SWR writes a computed `thetaSE` mirroring `thetaSERaw`?"

richford (dismissed):
Amazing! I have only one minor nit below about a stale comment. But it's non-blocking so I'll approve here.

ksmontville:
@claude Let's check this PR for regressions that may have been introduced in recent merge commits, or in roar-swr `serve.js` file being writed to use the new sdk session bootstrap.

richford (approved):

PR #1856, opened 2026-06-16
Return isSuperAdmin from the /me endpoint

richford:
_First in a stacked series of 4 — this PR targets `project/backend-refactor`. Merge the series bottom-up (1887 → 1888 → 1889 → 1891)._

## Summary

This PR adds `isSuperAdmin` to the `/v1/me` response so an authenticated client can determine its own platform-wide super-admin status without a separate lookup. The flag comes from the verified `AuthContext` (`req.user.isSuperAdmin`), which `AuthGuardMiddleware` already populates from the user record — `/me` always describes the caller, so the auth context is the canonical source.

## Contract change

`MeSchema` gains a required `isSuperAdmin: z.boolean()`. This touches the shared `api-contract` package, so the backend and any SDK consumers pick up the field. It is additive — consumers that ignore it are unaffected.

## Tests

Controller, route, and integration tests assert `isSuperAdmin` is `false` for non-super-admin tiers and `true` for the super-admin tier.

Resolves yeatmanlab/roar-project-management#1887


richford:

## Background

The `/v1/me` endpoint is the authenticated user's canonical self-description, but it currently omits whether the caller is a platform-wide super admin. Clients that need this (for example, the dashboard deciding what a user may see) have to infer it from a separate source such as Firestore custom claims. The backend already has the verified flag on the request auth context (`req.user.isSuperAdmin`, populated by `AuthGuardMiddleware` from the user record), so `/me` can report it directly.

## Proposal

Add a required `isSuperAdmin: boolean` field to the `/me` response, sourced from `AuthContext`. Because `/me` always describes the caller, the verified auth-context flag is the canonical source — there is no risk of disclosing another user's status.

## Acceptance criteria

- [ ] `MeSchema` in `packages/api-contract` includes `isSuperAdmin: boolean`.
- [ ] The controller returns `authContext.isSuperAdmin`.
- [ ] Controller, route, and integration tests assert `false` for non-super-admin tiers and `true` for the super-admin tier.
- [ ] The change is additive — existing consumers that ignore the field are unaffected.

## Notes

- Touches the shared `api-contract` package, which affects both the backend and SDK consumers.

amyyeung17 (dismissed):

amyyeung17 (approved):

PR #1888, opened 2026-06-22
Fix auth emulator for non-super-admins

richford:
## Summary

This PR fixes two bugs in the local Firebase Auth emulator dev flow (`npm run dev:local`): non-super-admins were routed to an empty participant page, and logging out then back in as a different user hung until a manual refresh. Both fixes are gated on `VITE_FIREBASE_EMULATOR_ENABLED` and leave the deployed production auth paths unchanged.

## Route non-super-admins to the admin dashboard

In emulator mode the Firestore `userClaims` document isn't available, so `resolveUserClaims` derives claims from the backend `/me` response. It previously derived only `super_admin`, so every non-super-admin fell through to PARTICIPANT in `useUserType` and rendered the (locally-empty) assignments page.

`resolveUserClaims` now derives admin routing claims from `/me` `userType` via a new `deriveEmulatorClaims` helper:

- `super_admin` ← `/me` `isSuperAdmin`
- `userType` `admin` / `educator` → `{ admin: true, role: UserRoles.ADMIN }` so `useUserType` resolves to ADMIN and they route to `HomeAdministrator`
- `student` / `caregiver` → no admin claims → `HomeParticipant`

The synthesized claims drive only client-side routing/UI; the backend still scopes all data via FGA, so they cannot widen access. The derivation fails closed when the `/me` payload is missing or unreachable. Covered by `resolveUserClaims.test.js`.

## Avoid the re-login hang on emulator sign-out

On sign-out the dashboard re-initializes Firekit client-side. In emulator mode roar-firekit calls `connectAuthEmulator` during `init()`, which Firebase only permits before the Auth instance is used; the second call leaves Auth broken until a full page reload (the reason a manual refresh recovers it).

In emulator mode, `useSignOutMutation` now performs a full reload to `/signin` (`window.location.assign`) after clearing client state, instead of the client-side re-init + SPA redirect. This gives the next sign-in a clean Firebase/Firekit init — exactly what the manual refresh does. The production path (re-init + `router.push`) is unchanged.

## Verification

- Independent auth-focused review: approved, no blockers. Production paths confirmed unchanged and emulator-gated; routing traced correct for all five user tiers against the `/me` `userType` enum (`['student','educator','caregiver','admin']`); list-query enablement confirmed independent of `minimalAdminOrgs`.
- `resolveUserClaims.test.js` covers the mapping (admin/educator/student/caregiver × super/non-super, plus fail-closed on empty/missing `/me`).
- Issue 2's reload fix is reasoned from the `connectAuthEmulator` call-once constraint and the "refresh fixes it" symptom — worth a quick runtime confirmation in the local stack (log out → log back in as a different user).

## Scope

Local-dev/emulator only; both paths gated on `VITE_FIREBASE_EMULATOR_ENABLED`.

Resolves [roar-project-management/issues/1921](https://github.com/yeatmanlab/roar-project-management/issues/1921)

### DIFF
--- a/apps/dashboard/src/composables/mutations/useSignOutMutation.js
+++ b/apps/dashboard/src/composables/mutations/useSignOutMutation.js
@@ -4,6 +4,7 @@ import * as Sentry from '@sentry/vue';
 import { useAuthStore } from '@/store/auth';
 import { SIGN_OUT_MUTATION_KEY } from '@/constants/mutationKeys';
 import { APP_ROUTES } from '@/constants/routes';
+import { IS_FIREBASE_EMULATOR_ENABLED } from '@/constants/firebase';
 
 /**
  * Sign-Out mutation.
@@ -37,6 +38,19 @@ const useSignOutMutation = () => {
       // Clear the query client to remove all cached data — this also drops
       // any cached `/me` payload.
       queryClient.clear();
+
+      // Local emulator: re-initializing Firekit client-side re-runs
+      // connectAuthEmulator on the already-used Auth instance, which Firebase only
+      // permits before the instance is used — the second call leaves auth broken
+      // until a manual page refresh. That is the cause of the local "hang when
+      // logging out and back in as a different user". A full reload to the sign-in
+      // page gives the next sign-in a clean Firebase/Firekit init, exactly like the
+      // manual refresh that currently works around it. Gated on the emulator flag,
+      // so deployed builds keep the client-side re-init + SPA redirect below.
+      if (IS_FIREBASE_EMULATOR_ENABLED) {
+        window.location.assign(APP_ROUTES.SIGN_IN);
+        return;
+      }
 
       // Re-initialize Firekit. This is necessary to ensure that Firekit is properly reset after
       // sign-out in order to allow a new user to sign in.
--- a/apps/dashboard/src/composables/mutations/useSignOutMutation.js
+++ b/apps/dashboard/src/composables/mutations/useSignOutMutation.js
@@ -45,10 +45,11 @@ const useSignOutMutation = () => {
       // until a manual page refresh. That is the cause of the local "hang when
       // logging out and back in as a different user". A full reload to the sign-in
       // page gives the next sign-in a clean Firebase/Firekit init, exactly like the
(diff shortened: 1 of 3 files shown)


richford:
@claude, please review this PR. Here is the originating issue:

Two bugs in the local Firebase Auth emulator dev flow (`npm run dev:local`). Both are inert in deployed builds (gated on `VITE_FIREBASE_EMULATOR_ENABLED`).

## Non-super-admins land on an empty participant page

### Symptom

Signing into the local stack as any non-super-admin (district/school admin, teacher) shows the participant view: "No assignments available at the moment. If you have just registered, please wait a few minutes and refresh the page."

### Cause

In emulator mode the Firestore `userClaims` document isn't available, so `resolveUserClaims` derives claims from the backend `/me` response — but it only derived `super_admin`. `useUserType` (which drives `HomeSelector` routing) returns ADMIN only when `claims.role` is an admin role or `claims.minimalAdminOrgs` is non-empty. With neither set, every non-super-admin fell through to PARTICIPANT and rendered the (Firestore-backed, locally-empty) assignments page.

### Expected

Admins and teachers route to the admin dashboard; students/caregivers to the participant view. The backend already scopes data via FGA, so the client only needs to route correctly — no admin-org list is required.

## Logging out and back in as a different user hangs

### Symptom

After signing out and signing back in as a different user, the page hangs. A manual browser refresh fixes it and lets the user in.

### Cause

On sign-out the dashboard re-initializes Firekit client-side (`authStore.initFirekit()`). In emulator mode roar-firekit calls `connectAuthEmulator` during `init()`. Firebase only permits that call before the Auth instance is used; the second call (after the first session used the instance) leaves Auth in a broken state until a full page reload — which is exactly why a manual refresh recovers it.

### Expected

A fresh sign-in works without a manual refresh.

## Scope

Local-dev/emulator only. Both fixes are gated on `VITE_FIREBASE_EMULATOR_ENABLED` and leave the deployed production paths (Firestore claims read; client-side Firekit re-init + SPA redirect) unchanged.

richford:
@claude, please re-review

kellyel (approved):

PR #1890, opened 2026-06-22
Refactor local development and CI infrastructure

maximilianoertel:
## Summary

This PR overhauls the local development environment to be fully Docker-based, self-contained, and reproducible. It eliminates `server-test.ts`, the `TestAuthProvider`, shell-based OpenFGA setup scripts, and the `dev-local-stack.sh` orchestrator — replacing them with a single `docker compose up` stack, a unified seed script, and Firebase Auth emulator-based authentication everywhere (dev, CI, and SDK integration tests).

## Motivation

The previous local dev setup required Java on the host (for the Firebase Auth emulator), manual shell scripts to initialize OpenFGA, a special `server-test.ts` entry point with `TestAuthProvider` for Cypress, and a 230-line bash orchestrator (`dev-local-stack.sh`). Each of these was a source of friction and divergence between local dev, CI, and production.

### Docker infrastructure

- **Consolidated `docker-compose.yml`**: Postgres (with automatic database creation via init script), OpenFGA (port 4010, no TLS), and a containerized Firebase Auth emulator (no Java required on host). Removed `docker-compose.local-dev.yml` and the `openfga-init` service.
- **Postgres init script** (`docker/postgres/init-dev-databases.sh`): Creates all four databases (core, assessment, core_test, assessment_test) and provisions FDW prerequisites automatically.
- **Firebase Auth emulator Dockerfile** (`docker/firebase-auth-emulator/`): Runs the Auth emulator inside Docker so developers don't need Java or firebase-tools installed.
- **OpenFGA moved to port 4010** to avoid conflicts with common web services on 8080.

### Seed script

- **New unified seed script** (`apps/backend/seeds/index.ts`): Handles database migrations, fixture seeding, FGA store creation/model deployment/tuple sync, Auth emulator user creation, and fixture file writing — all in one entry point.
- **Deterministic dev fixture** (`apps/backend/tooling/dev-fixture.ts`): Hardcoded UUIDs for orgs, users, administrations, and tasks so data is stable across runs and matches Cypress expectations.
- **FGA IDs written to `.env`** by the seed script, eliminating the need for the `openfga-init` Docker container and shell scripts.
- **`npm run dev:seed` and `npm run dev:reset`** convenience scripts.

### Authentication

- **Eliminated `TestAuthProvider`** and `BUILD_TEST_SERVER` flag. All environments now use the Firebase Auth emulator via `FIREBASE_AUTH_EMULATOR_HOST`.
- **Eliminated `server-test.ts`** (721 lines). The production `server.ts` entry point handles all environments; the seed script handles data setup.

### CI

- **Docker Compose for CI test infrastructure**: The CI test job uses the same `docker-compose.yml` as local dev (Postgres + OpenFGA), removing divergence.
- **Consolidated `.env.test.example`**: All CI env vars defined in one place, loaded via `dotenvx`.
- **Removed stale scripts**: `scripts/dev-local-stack.sh`, `scripts/init-openfga-store.sh`, `scripts/setup-openfga-local.sh`, `apps/backend/scripts/create-test-databases.ts`.
- **Gated Cypress jobs on build** to avoid wasted CI time when the build step fails.

### SDK integration tests

- **Split into unit and integration Vitest projects** with separate configs.
- **Graceful skip when infrastructure is unavailable**: Probes OpenFGA and Auth emulator before seeding; provides `integrationReady: false` via Vitest's `provide`/`inject` so integration tests skip cleanly (exit 0) instead of crashing unit tests.
- **Fixed teardown**: The previous teardown was a named export alongside a default export — Vitest requires the teardown to be returned from the default export. The backend process was never killed, leaving stale processes across runs.
- **Port-in-use detection** before spawning the backend server.
- **Early process exit detection** in health polling.
- **SIGTERM → SIGKILL escalation** and FGA store cleanup in teardown.

## Local dev workflow (after this PR)

```bash
docker compose up -d --wait   # Start Postgres, OpenFGA, Firebase Auth emulator
npm run dev:seed              # Migrate databases, seed fixture data
npm run dev                   # Start dashboard + backend via turbo
```

Resolves https://github.com/yeatmanlab/roar-project-management/issues/1935


apps/backend/package.json
```diff
@@ -24,19 +24,19 @@
...
     "db:seed:roar-pa": "tsx seeds/roar-pa.seed.ts",
     "db:seed:roar-swr": "tsx seeds/roar-swr.seed.ts",
     "db:seed:roar-sre": "tsx seeds/roar-sre.seed.ts",
     "db:seed:roar-letter": "tsx seeds/roar-letter.seed.ts",
     "db:seed:roar-multichoice": "tsx seeds/roar-multichoice.seed.ts",
     "db:seed:roar-survey": "tsx seeds/roar-survey.seed.ts",
-    "db:seed:roar-levante-tasks": "tsx seeds/roar-levante-tasks.seed.ts"
+    "db:seed:roar-levante-tasks": "tsx seeds/roar-levante-tasks.seed.ts",
+    "dev:seed": "tsx ./seeds/index.ts",
+    "dev:reset": "npm run db:migrate && npm run dev:seed"
```
mikekirsten: `dev:reset` runs both `db:migrate` and `dev:seed`, but based on what I can see, `dev:seed` appears to run migrations anyway, so this may be redundant?

apps/backend/package.json
```diff
@@ -24,19 +24,19 @@
...
-    "fga:init": "bash ../../scripts/init-openfga-store.sh",
     "db:seed:roar-pa": "tsx seeds/roar-pa.seed.ts",
     "db:seed:roar-swr": "tsx seeds/roar-swr.seed.ts",
     "db:seed:roar-sre": "tsx seeds/roar-sre.seed.ts",
     "db:seed:roar-letter": "tsx seeds/roar-letter.seed.ts",
     "db:seed:roar-multichoice": "tsx seeds/roar-multichoice.seed.ts",
     "db:seed:roar-survey": "tsx seeds/roar-survey.seed.ts",
-    "db:seed:roar-levante-tasks": "tsx seeds/roar-levante-tasks.seed.ts"
+    "db:seed:roar-levante-tasks": "tsx seeds/roar-levante-tasks.seed.ts",
+    "dev:seed": "tsx ./seeds/index.ts",
```
mikekirsten: This `dev:seed` does the full local-dev setup and test fixture right? Do you think we can rename it to be clearer in what it does (it doesn't just seed)? It would also be pretty cool to eventually consolidate all the assessment seeders behind something like `db:seed -- --task roar-swr`, which means we can get of the individual seeder scripts in this file, while retaining the same control.

mikekirsten (commented):
Awesome @maximilianoertel! Two small questions below

apps/backend/package.json
maximilianoertel: Good catch! Fixed in bb92ca634: `dev:reset` now just truncates all tables which is all it needs to do to offer a way to reset the local environment.

apps/backend/package.json
maximilianoertel: Same here, great catch! 

`dev:seed` has been split into three focused scripts (bb92ca634): 
- `dev:setup` for infrastructure only: FDW, migrations, FGA store + model, persists IDs to .env 
- `dev:seed` for data only: seed fixtures, sync FGA tuples, auth emulator users, write fixture files 
- `dev:reset` for truncating all tables 
- `dev:init` as convenience orchestrator that runs setup + seed 

Assessment seeders have been consolidated (909d0ce7b): All 9 per-assessment seed scripts are now behind a single `dev:seed -- --task <name>` entrypoint (e.g. `dev:seed -- --task roar-swr`). A config registry in _task-seed-configs.ts_ defines the task metadata, allowed params, and validation per assessment

apps/backend/seeds/configs/roar-letter.config.ts
```diff
@@ -0,0 +1,67 @@
+import { letter } from '@roar-platform/assessment-schema';
+
+import type { TaskSeedConfig } from '../task-seed-configs';
+
+const { LETTER_TASK_IDS, PHONICS_TASK_IDS, LETTER_LANGUAGES, LETTER_SCORING_VERSION } = letter;
+
+const VALID_SCORING_VERSIONS = new Set(Object.values(LETTER_SCORING_VERSION));
```
ksmontville:
```suggestion
const VALID_SCORING_VERSIONS = new Set<number>(Object.values(LETTER_SCORING_VERSION));
```
My editor was throwing error: `"Argument of type 'number' is not assignable to paremeter of type '1'"` further down in the .has() check on line 50 prior to this fix. 

The same error occurs in the other task configs which declare `VALID_SCORING_VERSIONS`.

apps/backend/seeds/configs/roar-sre.config.ts
```diff
@@ -0,0 +1,68 @@
+import { sre } from '@roar-platform/assessment-schema';
+
+import type { TaskSeedConfig } from '../task-seed-configs';
+
+const { SRE_TASK_IDS, SRE_LANGUAGES, SRE_SCORING_VERSION } = sre;
+
+const VALID_SCORING_VERSIONS = new Set(Object.values(SRE_SCORING_VERSION));
+const SUPPORTED_LANGUAGES = new Set(Object.values(SRE_LANGUAGES).map((l) => l.code));
```
ksmontville:
```suggestion
const SUPPORTED_LANGUAGES = new Set<string>(Object.values(SRE_LANGUAGES).map((l) => l.code));
```
Same class of error as the `VALID_SCORING_VERSIONS` set.

apps/backend/seeds/configs/roar-swr.config.ts
```diff
@@ -0,0 +1,76 @@
...
+    'scoringVersion',
+    'skipInstructions',
+    'storyOption',
+    'userMode',
+  ]),
+  validateVariant(loc, params) {
+    const lng = params.lng as string | undefined;
+    if (!lng) throw new Error(`${loc}: "lng" is required`);
+
+    const validLanguages = new Set(Object.values(SWR_LANGUAGES).map((l) => l.code));
```
ksmontville:
```suggestion
    const validLanguages = new Set<string>(Object.values(SWR_LANGUAGES).map((l) => l.code));
```

apps/backend/seeds/configs/roav-apps.config.ts
```diff
@@ -0,0 +1,30 @@
+import { roavApps } from '@roar-platform/assessment-schema';
+
+import type { TaskSeedConfig } from '../task-seed-configs';
+
+const { ROAV_MP_TASK_ID, ROAV_RVP_TASK_ID } = roavApps;
+
+const KNOWN_TASK_IDS = new Set([ROAV_MP_TASK_ID, ROAV_RVP_TASK_ID]);
```
ksmontville:
```suggestion
const KNOWN_TASK_IDS = new Set<string>([ROAV_MP_TASK_ID, ROAV_RVP_TASK_ID]);
```

apps/backend/seeds/configs/roar-letter.config.ts
```diff
@@ -0,0 +1,67 @@
...
+      nameTechnical: 'Rapid Online Assessment of Reading — Phonics',
+    },
+  },
+  validateVariant(loc, params) {
+    const task = params.task as string | undefined;
+    if (task !== 'letter' && task !== 'phonics') {
+      throw new Error(`${loc}: "task" must be "letter" or "phonics", got "${task}"`);
+    }
+
+    if (task === 'letter') {
```
ksmontville: Non-blocking:

These repeated `"letter"` and `"phonics"` strings could be extracted into a local constant like:

```ts
/**
 * Valid values of the `task` variant parameter — the task-family discriminator.
 * Distinct from the seeded task IDs (LETTER_TASK_IDS / PHONICS_TASK_IDS): a
 * Spanish variant has `task: 'letter'` but resolves to the `letter-es` task ID.
 */
const TASK_PARAM = {
  LETTER: 'letter',
  PHONICS: 'phonics',
} as const;
```

They appear multiple times and as strings embedded in error messages, so there is some value.

This class of refactor is likely true for the other task config files, as well.

ksmontville (changes requested):
Looks good, just a couple of small findings.

I also want to make sure that we have fully considered the (new) storage emulator functionality that was introduced when `roar-readaloud` went into the monorepo. The storage emulator is also used by `roav-apps` and `roav-ran`.

Is this something that we should consider here? Our main stack doesn't use any storage operations as far as I know. The assessments use the emulator storage to upload their audio/video outputs (the above assessments use audio and eyetracking data in their trials which researachers need to inspect directly in storage objects). 

In other words, should this local dev stack also define emulator constants and infra? There are two `firebase.json` files which are of concern here: `apps/assessments/shared/firebase.json` and `docker/firebase-auth-emulator/firebase.json`.

Just want to ensure that we've explored all avenues regarding emulation. If unification is easier for the stack as a whole, then maybe should consider it. If it's better to let the assessment environment handle storage emulation itself, then that's what we should do.

maximilianoertel:
> Looks good, just a couple of small findings.
> 
> I also want to make sure that we have fully considered the (new) storage emulator functionality that was introduced when `roar-readaloud` went into the monorepo. The storage emulator is also used by `roav-apps` and `roav-ran`.
> 
> Is this something that we should consider here? Our main stack doesn't use any storage operations as far as I know. The assessments use the emulator storage to upload their audio/video outputs (the above assessments use audio and eyetracking data in their trials which researachers need to inspect directly in storage objects).
> 
> In other words, should this local dev stack also define emulator constants and infra? There are two `firebase.json` files which are of concern here: `apps/assessments/shared/firebase.json` and `docker/firebase-auth-emulator/firebase.json`.
> 
> Just want to ensure that we've explored all avenues regarding emulation. If unification is easier for the stack as a whole, then maybe should consider it. If it's better to let the assessment environment handle storage emulation itself, then that's what we should do.

Good catch, @ksmontville! 

From what I could see: 
- Storage usage is upload-only and funnels entirely through the Assessment SDK's `uploadFile()` 
- Only `readaloud** and `ran` upload recordings
- On this branch, `resolveStorageBucket()` within the Assessment SDK assumes the standalone assessment context — it reads `process.env.FIREBASE_AUTH_EMULATOR_HOST` (invisible to the dashboard's Vite bundle) and calls `getApp()` (the dashboard only creates *named* Firebase apps). The failure is swallowed (`videoCapture.js` catches and returns `null`), so dashboard-launched readaloud/RAN would play fine but **silently drop recordings** — in local dev and, once this stack ships, likely in staging/prod too.

So, you're right that the local dev stack should rely on the storage emulator the same way it relies on the auth emulator — for the dashboard as well as the assessments. But provisioning the emulator in this PR alone wouldn't change anything observable: the SDK can't reach it from the dashboard until the storage wiring is fixed. 

Given the size of this PR and that the SDK change affects upload behaviour in all environments, I'll file tickets for two follow-up PRs:

1. Dashboard storage wiring + platform-stack emulator. For that, we need to extend the SDK context so the host app can inject a Firebase app/storage handle (or a browser-safe emulator signal — e.g. derived from `VITE_FIREBASE_EMULATOR_AUTH_HOST`, same host, port 9199, mirroring the existing auth-emulator pattern), and add the storage emulator + `storage.rules` to the platform dev stack.
2. A second follow-up that will apply the same class of local-dev environment changes to the assessment development environment.

So for this PR I'd recommend we don't introduce any further changes for the emulator. Does that work for you?

The other requested changes have been addressed in [b975592](https://github.com/yeatmanlab/roar-dashboard/pull/1890/commits/b9755929ea0891fbc363e6d82f87ba1082abf214) and [e9577fe](https://github.com/yeatmanlab/roar-dashboard/pull/1890/commits/e9577fe16118667f23dc0f960c890d725744c533)

ksmontville (dismissed):
Thanks for the robust follow-up, and I agree with the scope of work that you've outlined as a future enhancement.

Looks good to me.

Emily-ejag (approved):

PR #1894, opened 2026-06-22
Migrate district read composables to the ts-rest backend client

richford:
## Summary

This PR migrates the three district *read* composables off the legacy Firestore data layer onto the typed ts-rest backend client, making the backend (which already access-filters every response) the single source of truth for which districts and schools a user can see:

- `useDistrictsListQuery` → `GET /districts` (walks all pages, returns the full array)
- `useDistrictsQuery` → `GET /districts/:id` per id via `Promise.all` (input order preserved)
- `useDistrictSchoolsQuery` → `GET /districts/:districtId/schools` (walks all pages)

All three mirror the already-migrated `useAdministrationsListQuery`: a `getRoarApiClient()` call, a structured `Error` (carrying `.status` + `.body`) thrown on any non-200, `enabled` gated through `computeQueryOverrides`, and a `retry` policy that bails on terminal-auth / rostering-ended errors and is placed after `...options` so callers can't silently override it.

The query keys (`DISTRICTS_LIST_QUERY_KEY`, `DISTRICTS_QUERY_KEY`, `DISTRICT_SCHOOLS_QUERY_KEY`) are unchanged, and the parameterized queries keep the reactive `id` / `districtId` in the key array.

## Dropped client-side authorization

The legacy `useDistrictsListQuery` (and `useDistrictSchoolsQuery`) reconstructed the visible org set in the browser: they read `userClaims.minimalAdminOrgs`, derived `isSuperAdmin` via `useUserType`, and handed both to `orgFetcher`, which filtered Firestore client-side.

`GET /districts` and `GET /districts/:districtId/schools` already scope results to the caller (super admin → all; supervisory roles → their accessible org tree; everyone else → only what they belong to). So this PR **removes** the
`useUserClaimsQuery` + `useUserType` + `minimalAdminOrgs` + `orgFetcher` machinery from the list and schools composables. Enablement is now gated only on `authStore.accessToken` (plus `districtId` presence / `hasArrayEntries(ids)` for the parameterized queries). No `embed` is requested — no consumer of these three queries reads counts.

## Response-shape mapping

The backend `DistrictDetailSchema` / `SchoolDetailSchema` nest external identifiers under `identifiers` (`mdrNumber`, `ncesId`, `stateId`, `schoolNumber`) and address fields under `location`. The legacy Firestore documents exposed those flat. A new shared helper, `@/helpers/mapOrg` (`mapDistrictToOrg` / `mapSchoolToOrg`), spreads the raw backend object and flattens `location` + `identifiers` to top-level fields, so consumers keep working without edits.

Verified consumer reads:

|---|---|---|---|
| list | `OrgPicker`, `OrgsList`, `CreateOrgs` | `id`, `name` | yes |
| by-id | `useOrgQuery` → `ScoreReport` / `ProgressReport` | `name`, `clever` | `name` yes; `clever` **no** |
| schools | `OrgsList`, `CreateOrgs`, `ProgressReport`, `ScoreReport` | `id`, `name`, `lowGrade` | `id`/`name` yes; `lowGrade` **no** |

### Fields the backend does not yet provide

Two legacy fields have no equivalent on the current backend org schemas. Both are read defensively in the consumers and degrade gracefully (no crash):

- **`district.clever`** — `ScoreReport` adds a "State ID" CSV column only when `orgData.value?.clever === true`. Absent → the column is omitted, which is the correct default for non-Clever orgs.
- **`school.lowGrade`** — `ScoreReport` derives a per-school grade label via `getGrade(school.lowGrade ?? 0)`. Absent → it falls back to grade 0.

These are documented with a `@TODO` in `@/helpers/mapOrg` to re-map once the backend surfaces them. No consumer required edits in this PR.

## Tests

Each co-located unit test was rewritten to the ts-rest pattern (mock `getRoarApiClient`, mock the auth store, call through to the real `useQuery`, and capture `queryFn` / `retry` per test). Coverage per composable: correct query key, `enabled` gating (no token / no ids / no districtId), mapped items on 200, pagination walk across multiple pages (list + schools), by-id order preservation across out-of-order resolution, structured error on non-200, and the retry predicate. Legacy claims/`orgFetcher` assertions were removed.

## Verification

- Lint not run locally (the repo's ESLint flat-config crashes in this environment); CI is authoritative.
- The dashboard is plain JS — imports were sanity-checked against the resolved symbols (constants, helpers, and the `client.districts.{list,get,listSchools}` contract methods).

Resolves yeatmanlab/roar-project-management#1925


richford:

# Migrate district read composables to the ts-rest backend client

## Background

The dashboard's three district *read* composables still talk to Firestore via the legacy `@/helpers/query/orgs` / `@/helpers/query/utils` data layer:

- `useDistrictsListQuery` — lists all districts the caller can see.
- `useDistrictsQuery` — fetches one or more districts by ID.
- `useDistrictSchoolsQuery` — lists the schools within a district.

The new backend now serves these reads:

- `GET /districts` (paginated, access-scoped)
- `GET /districts/:id`
- `GET /districts/:districtId/schools` (paginated, access-scoped)

The backend already access-filters every response (super admin → all; everyone else → only what they belong to / their accessible org tree), so the dashboard should treat the backend as the source of truth and drop the client-side claims-based filtering it currently performs.

## Problem

`useDistrictsListQuery` (and the schools query) replicate authorization in the browser: they read `userClaims.minimalAdminOrgs`, derive `isSuperAdmin`, and hand both to `orgFetcher`, which then assembles the visible district/school set client-side from Firestore. This duplicates the backend's authorization logic, keeps the Firestore dependency alive, and can drift from what the backend actually returns.

## Goal

Migrate all three composables onto the typed ts-rest client (`getRoarApiClient()` from `@/clients/roar-api`), mirroring the already-migrated `useAdministrationsListQuery` pattern:

- `do/while` pagination walk for the two list composables; `Promise.all` (input order preserved) for the by-IDs composable.
- Throw a structured `Error` (carrying `.status` + `.body`) on any non-200.
- Gate `enabled` via `computeQueryOverrides` on `authStore.accessToken` (plus `hasArrayEntries(ids)` / `districtId` presence for the parameterized queries).
- `retry` policy gated on `isRosteringEndedError` / `isTerminalAuthError` (`MAX_RETRIES = 3`), placed after `...options` so callers can't override it.

Drop the claims/`isSuperAdmin`/`minimalAdminOrgs`/`orgFetcher` machinery from the list composable entirely — the backend scopes results.

## Constraints

- Preserve the consumer-facing return shape. Consumers (`OrgPicker`, `OrgsList`, `CreateOrgs`, `ProgressReport`, `ScoreReport`, and `useOrgQuery`) must keep working without edits. The backend nests `identifiers` and `location`; the legacy Firestore docs exposed them flat, so the mapping must flatten them back.
- Keep the existing query keys (`DISTRICTS_LIST_QUERY_KEY`, `DISTRICTS_QUERY_KEY`, `DISTRICT_SCHOOLS_QUERY_KEY`), and keep the `id`/`districtId` ref in the key array for the parameterized queries.
- Rewrite the co-located unit tests to the ts-rest pattern (mock `getRoarApiClient`, the auth store, and `useQuery`; capture `queryFn`/`retry`).

## Out of scope

- Org *write* paths (create/edit/delete orgs) — still Firestore.
- `useOrgsTableQuery`, `orgFetchAll`, `fetchDocById`, and the activation-code lookup — these remain on the legacy layer for now.
- Backend changes. The endpoints already exist.

richford:
@claude, please re-review. Regarding your previous review:

- **`mapOrg.js` dual nested/flat representation** — `mapDistrictToOrg` and `mapSchoolToOrg` now destructure `{ location, identifiers, ...rest }` and spread only `rest` + the flattened fields, so the result carries a single flat representation instead of both. This also makes the drop of `location.coordinates` explicit (no consumer reads it), and the JSDoc now says so. Tests were unaffected — they assert flat fields (`mdrNumber`, `city`), not the nested objects.
- **All-or-nothing semantics** — surfaced the constraint inline at the `Promise.all` (it was only in the JSDoc before).
- Left as-is: the `tags`/`clever`/`lowGrade` gaps and the `schoolNumber`-on-districts note are documented `@TODO`s tied to backend schema, not this PR.

apps/dashboard/src/helpers/mapOrg.js
```diff
@@ -0,0 +1,98 @@
...
+const flattenLocation = (location) => {
+  if (!location) return {};
+  return {
+    addressLine1: location.addressLine1,
+    addressLine2: location.addressLine2,
+    city: location.city,
+    stateProvince: location.stateProvince,
+    postalCode: location.postalCode,
+    country: location.country,
+  };
```
Emily-ejag: Can we do something like this to avoid adding keys with `undefined`. SO that we include fields that are present in the backend response?

```suggestion
return Object.fromEntries(
    ['addressLine1', 'addressLine2', 'city', 'stateProvince', 'postalCode', 'country']
      .filter((key) => Object.hasOwn(location, key) && location[key] !== undefined)
      .map((key) => [key, location[key]]),
  );
```

apps/dashboard/src/helpers/mapOrg.js
```diff
@@ -0,0 +1,98 @@
...
+ * @returns {Object} The flattened identifier fields (empty when identifiers are absent).
+ */
+const flattenIdentifiers = (identifiers) => {
+  if (!identifiers) return {};
+  return {
+    mdrNumber: identifiers.mdrNumber,
+    ncesId: identifiers.ncesId,
+    stateId: identifiers.stateId,
+    schoolNumber: identifiers.schoolNumber,
+  };
```
Emily-ejag: Same here: 

```suggestion
return Object.fromEntries(
    ['mdrNumber', 'ncesId', 'stateId', 'schoolNumber']
      .filter((key) => Object.hasOwn(identifiers, key) && identifiers[key] !== undefined)
      .map((key) => [key, identifiers[key]]),
  );
```

Emily-ejag (changes requested):

richford:
@claude, please re-review

Emily-ejag (approved):

Emily-ejag (approved):

PR #1896, opened 2026-06-22
Migrate group and class read composables to the ts-rest backend client

richford:
## Summary

This PR migrates the three group/class *read* composables off the legacy
Firestore data layer onto the typed ts-rest backend client, making the backend
(which already access-filters every response) the single source of truth for
which groups and classes a user can see:

- `useGroupsListQuery` → `GET /groups` (walks all pages, returns the full array)
- `useGroupsQuery` → `GET /groups/:groupId` per id via `Promise.all` (input order preserved)
- `useClassesQuery` → `GET /classes/:classId` per id via `Promise.all` (input order preserved)

It is FE-PR3 in the dashboard read-path migration, stacked on the school read
migration (FE-PR2), and mirrors it exactly: a `getRoarApiClient()` call, a
structured `Error` (carrying `.status` + `.body`) thrown on any non-200,
`enabled` gated through `computeQueryOverrides`, and a `retry` policy that bails
on terminal-auth / rostering-ended errors and is placed after `...options` so
callers can't silently override it.

The query keys (`GROUPS_LIST_QUERY_KEY`, `GROUPS_QUERY_KEY`, `CLASSES_QUERY_KEY`)
are unchanged, and the by-id queries keep the reactive ids in the key array.

## Endpoints

| Composable | Method + params |
|---|---|
| `useGroupsListQuery` | `client.groups.list({ query: { page, perPage: 100, sortBy: 'name', sortOrder: 'asc' } })`, paginated |
| `useGroupsQuery` | `client.groups.get({ params: { groupId } })`, one call per id |
| `useClassesQuery` | `client.classes.get({ params: { classId } })`, one call per id |

## Dropped client-side authorization

The legacy `useGroupsListQuery` reconstructed the visible org set in the
browser: it read `userClaims.minimalAdminOrgs`, derived `isSuperAdmin` via
`useUserType`, and handed both to `orgFetcher`, which filtered Firestore
client-side. `useGroupsQuery` and `useClassesQuery` used `fetchDocumentsById`,
which silently dropped ids the caller couldn't read.

`GET /groups`, `GET /groups/:groupId`, and `GET /classes/:classId` already scope
results to the caller (super admin → all; supervisory roles → their accessible
org tree; everyone else → only what they belong to). So this PR **removes** the
`useUserClaimsQuery` + `useUserType` + `minimalAdminOrgs` + `orgFetcher`
machinery from `useGroupsListQuery`. Enablement is now gated only on
`authStore.accessToken` (plus `hasArrayEntries(ids)` for the by-id queries). No
`embed` is requested — no consumer of these queries reads counts (and the
groups schema has no `embed` option).

Because the by-id queries are now all-or-nothing (a non-200 for any id rejects
the whole query), the sole `useGroupsQuery` / `useClassesQuery` caller
(`useOrgQuery`, which passes a single id and selects `data[0]`) surfaces a
bad/inaccessible id as a query error rather than a silent `data[0] === undefined`.

### Groups — new `mapGroupToOrg`

The backend `GroupDetailSchema` nests address fields under `location`; the
legacy Firestore documents exposed those flat. Unlike districts/schools, groups
are flat standalone entities — they carry **no** `identifiers` block (no
MDR/NCES/state ids) and **no** `orgType` / `parentOrgId`. So a new
`mapGroupToOrg` is added to `@/helpers/mapOrg` that spreads the raw backend
object and flattens `location` only.

`mapGroupToOrg` produces these fields:

- Pass-through from `GroupDetailSchema`: `id`, `name`, `abbreviation`,
  `groupType`, `rosteringEnded` (when present), and the raw nested `location`.
- Flattened from `location`: `addressLine1`, `addressLine2`, `city`,
  `stateProvince`, `postalCode`, `country`.

Used in both `useGroupsListQuery` and `useGroupsQuery`.

### Classes — `mapClassToOrg` reused (decision)

`useClassesQuery` reuses the existing `mapClassToOrg` unchanged.

`mapClassToOrg` was originally written (FE-PR2) for `SchoolClassSchema`
(`GET /schools/:schoolId/classes`). The by-id endpoint mapped here returns
`ClassDetailSchema` instead. Comparing the two:

| | `SchoolClassSchema` | `ClassDetailSchema` |
|---|---|---|
| Flat core | `id`, `name`, `schoolId`, `districtId`, `classType`, `courseId`, `number`, `period`, `subjects`, `grades`, `schoolLevels` | same flat core |
| `location` | (absent) | present — a **plain string** (room/location label), NOT a nested address object |
| `rosteringEnded` | (absent) | present (optional) |
| timestamps | `createdAt`, `updatedAt` | omitted |

Both shapes are entirely flat. The only new field on `ClassDetailSchema`,
`location`, is a scalar string (explicitly *not* an assembled `{ city, ... }`
address object — see the schema's own JSDoc), so it passes through the spread
unchanged and needs no flattening. **Decision: reuse `mapClassToOrg`
unchanged** — `{ ...schoolClass }` yields the correct flat record for either
shape, and adding a dedicated detail mapper would be misleading. The mapper's
JSDoc was extended to document that it now covers both shapes.

### Verified consumer reads

|---|---|---|---|
| `useGroupsListQuery` | `CreateOrgs` (Tags autocomplete) | `id`, `name`, `tags` (per group) | `id`/`name` yes; `tags` **no** |
| `useGroupsQuery` | `useOrgQuery` → `ScoreReport` / `ProgressReport` | `id`, `name`, `clever` | `id`/`name` yes; `clever` **no** |
| `useClassesQuery` | `useOrgQuery` → `ScoreReport` / `ProgressReport` | `id`, `name`, `clever` | `id`/`name` yes; `clever` **no** |

`useGroupsListQuery` flows only into `CreateOrgs`, whose only read off the group
objects is `org.tags` (CreateOrgs.vue:342, folded into the Tags autocomplete
suggestions). `useGroupsQuery` and `useClassesQuery` flow only through
`useOrgQuery` (groups / classes branches); the result (`orgData`, selected to
`data[0]`) is read for `name` (ScoreReport PDF/CSV filenames + header,
`ProgressReport` header, `useProgressExport` filename) and `clever` (ScoreReport
CSV column toggle, ScoreReport.vue:1583).

Note: the `class.tags` read in `CreateOrgs.vue:341` is fed by
`useSchoolClassesQuery` (migrated in FE-PR2), **not** by the by-id
`useClassesQuery` mapped here — so it is not a consumer of this PR's class query.

### Fields the backend does not yet provide

Three legacy reads have no equivalent on the current backend group/class shapes.
All are read defensively and degrade gracefully (no crash). All are
**secondary** (autocomplete-only / CSV-only) — none regresses a primary
feature:

- **`group.tags`** — `CreateOrgs.vue:342` folds group `tags` into the Tags
  autocomplete suggestion list. `GroupDetailSchema` has no `tags` field, so
  group tags contribute nothing to the suggestions. Users can still type tags
  freely; autocomplete-only, no data loss.
- **`group.clever`** — `ScoreReport.vue:1583` adds "Student ID" / "State ID"
  CSV columns when `orgData.clever === true`. `GroupDetailSchema` has no
  `clever` field, so for a group report `orgData.clever` is `undefined` → the
  columns are omitted, which is the correct default for non-Clever orgs. CSV
  export only; the on-screen report is unaffected.
- **`class.clever`** — same `ScoreReport.vue:1583` read fires for a class
  report. `ClassDetailSchema` has no `clever` field → same CSV-only effect.

All three are now documented in the `@TODO` gap block in `@/helpers/mapOrg.js`,
to be re-mapped once the backend exposes them. (The `clever` note there already
covered districts/schools; this PR broadens it to note that the same
`orgData.clever` read fires for groups and classes too, since `useOrgQuery`
routes all of them to `ScoreReport`.)

## Consumers

No consumer required edits — the mapping keeps the flat shape stable.
`useOrgQuery.test.js` mocks `useGroupsQuery` and `useClassesQuery` wholesale, so
it is unaffected.

## Tests

Co-located unit tests follow the district/school pattern (mock
`getRoarApiClient`, the auth store, and `useQuery`; capture `queryFn` / `retry`
per test):

- `useGroupsListQuery.test.js` (rewritten from the Firestore stub): query key;
  mapped items on 200 (location flattened, no identifiers); pagination walk;
  structured non-200 throw; disabled without accessToken; retry predicate.
- `useGroupsQuery.test.js` (rewritten): query key + ids ref; per-id fetch with
  order preservation + mapping; structured non-200 throw; disabled without ids /
  without accessToken; becomes enabled once ids populate; retry predicate.
- `useClassesQuery.test.js` (rewritten): query key + ids ref; per-id fetch with
  order preservation + mapping (asserts the scalar `location` string passes
  through unchanged); structured non-200 throw; disabled without ids / without
  accessToken; becomes enabled once ids populate; retry predicate.

## Verification

- Vitest runs in CI (cannot run in the authoring sandbox).

Resolves yeatmanlab/roar-project-management#1927


richford:

# Migrate group and class read composables to the ts-rest backend client

## Summary

Migrate the three group/class *read* composables off the legacy Firestore data layer onto the typed ts-rest backend client (`getRoarApiClient`), mirroring the district and school read migrations. The backend already access-filters every response, so it becomes the single source of truth for which groups and classes a user can see.

## Scope

- `useGroupsListQuery` (list) → `GET /groups` (paginated; walks every page and returns the full array).
- `useGroupsQuery` (fetch-by-IDs) → `GET /groups/:groupId` per id via `Promise.all`, preserving input order, all-or-nothing on any non-200.
- `useClassesQuery` (fetch-by-IDs) → `GET /classes/:classId` per id via `Promise.all`, preserving input order, all-or-nothing on any non-200.

## Motivation

This is PR3 in the dashboard read-path migration, stacked directly on the school read migration. It applies the identical pattern to the group and class read composables so the dashboard stops reaching into Firestore for those org reads and the backend's authorization model is the only access boundary.

## Acceptance criteria

- All three composables call the ts-rest client and throw a structured `Error` (carrying `.status` + `.body`) on any non-200.
- Enablement is gated only on `authStore.accessToken` (plus ids presence). The legacy `useUserClaimsQuery` / `useUserType` / `minimalAdminOrgs` / `orgFetcher` machinery is removed from `useGroupsListQuery`.
- The existing query keys (`GROUPS_LIST_QUERY_KEY`, `GROUPS_QUERY_KEY`, `CLASSES_QUERY_KEY`) are unchanged, with the reactive ids kept in the key array (by reference).
- Response shapes are mapped to the flat shape consumers read via `@/helpers/mapOrg`: a new `mapGroupToOrg` is added (flattens `location`); `mapClassToOrg` is reused for the by-id `ClassDetailSchema`.
- Co-located unit tests are rewritten to the ts-rest pattern.
- Any consumed legacy field the backend group/class shape lacks is documented (in `mapOrg.js` and the PR description) rather than silently dropped.

## Out of scope

- Backend schema changes (no new fields added to the group/class shapes).
- `OrgsList`'s `useOrgsTableQuery` (the OrgsList table is fed by a separate Firestore composable on its own migration track) and `useSchoolClassesQuery` (classes-in-a-school).

richford:
@claude, please re-review. The `undefined`-keys-from-`flattenLocation` note is deferred because it's a shared helper. I think if we do that, it belongs in a separate PR.

Emily-ejag (dismissed):
LGTM

Emily-ejag (approved):

PR #1897, opened 2026-06-22
Migrate useFamiliesQuery to the ts-rest backend client

richford:
## Summary

This PR migrates the `useFamiliesQuery` (fetch-by-IDs) read composable off the
legacy Firestore data layer onto the typed ts-rest backend client, making the
backend (which already authorizes every response) the single source of truth for
which families a user can read:

- `useFamiliesQuery` → `GET /families/:familyId` per id via `Promise.all` (input order preserved)

It is FE-PR4 in the dashboard read-path migration — the smallest of the set (a
single by-id composable) — stacked on the groups/classes read migration
(FE-PR3), and mirrors it exactly: a `getRoarApiClient()` call, a structured
`Error` (carrying `.status` + `.body`) thrown on any non-200, `enabled` gated
through `computeQueryOverrides`, and a `retry` policy that bails on
terminal-auth / rostering-ended errors and is placed after `...options` so
callers can't silently override it.

The `FAMILIES_QUERY_KEY` is unchanged, and the by-id query keeps the reactive
`familyIds` in the key array (by reference).

## Endpoint

| Composable | Method + params |
|---|---|
| `useFamiliesQuery` | `client.families.get({ params: { familyId } })`, one call per id |

## Parent-only authorization

`GET /families/:familyId` is **parent-only** (FGA `can_read: parent`): only a
user with the parent role on the family (its caretaker) or a super admin may
read it. Unlike org and group reads, **no** admin or supervisory role grants
access — admins, the family's own children, and all other users receive 403.

That matches the only legitimate consumer of this query: a caretaker viewing
their own family (and super admins). It is a tighter boundary than the legacy
`fetchDocumentsById`, which silently dropped ids the caller couldn't read.
Because the by-id query is now all-or-nothing (a non-200 for any id rejects the
whole query), an inaccessible id surfaces as a query error rather than a silent
`data[0] === undefined`.

## Dropped client-side data layer

The legacy `useFamiliesQuery` used `fetchDocumentsById(FIRESTORE_COLLECTIONS.FAMILIES, …)`,
reading family documents directly from Firestore and silently dropping missing
or inaccessible ids. This PR **removes** that path. Enablement is now gated only
on `authStore.accessToken` (plus `hasArrayEntries(familyIds)`). No `embed` is
requested — no consumer reads counts, and the families schema has no `embed`
option.

## Response-shape mapping — new `mapFamilyToOrg`

The backend `FamilyDetailSchema` is the leanest org shape on the platform:
`{ id, location?, rosteringEnded? }`. Families are flat, standalone entities
outside the org hierarchy, keyed by UUID — there is **no** `name`,
`abbreviation`, `identifiers`, or `orgType` / `parentOrgId`.

A new `mapFamilyToOrg` is added to `@/helpers/mapOrg`. Like districts, schools,
and groups, `FamilyDetailSchema.location` is an assembled address **OBJECT**
(`FamilyDetailLocationSchema`: `addressLine1`, `city`, `stateProvince`, …) —
**not** the free-text room-label string that classes carry — so the mapper
flattens it. There is no `identifiers` block to flatten.

`mapFamilyToOrg` produces these fields:

- Pass-through from `FamilyDetailSchema`: `id`, `rosteringEnded` (when present),
  and the raw nested `location`.
- Flattened from `location`: `addressLine1`, `addressLine2`, `city`,
  `stateProvince`, `postalCode`, `country`.

Used in `useFamiliesQuery`.

## The missing `name` lands on a vestigial path

`useFamiliesQuery`'s **sole** consumer is `useOrgQuery` (the `families` case) →
`ScoreReport.vue` / `ProgressReport.vue`. The result (`orgData`, selected to
`data[0]`) is read for `name` (ScoreReport PDF/CSV filenames + header,
ProgressReport header). `FamilyDetailSchema` has no `name`, so this read would
resolve to `undefined`.

This is **inconsequential** because the family-report path is **vestigial**:

- A family-scoped report renders only when `useOrgQuery` is called with
  `orgType === 'families'`, which requires an administration assigned to a
  `family` org.
- Administrations only assign to orgs (districts/schools), classes, and groups —
  the administration upsert body has **no** `families` field.
- Therefore no administration is ever assigned to a `family` org, and
  family-scoped reports are never generated in practice.

So the missing `name` lands on a dead path, not a live feature. **Reviewed and
accepted by the team.** Unlike the other documented gaps (`tags`, `clever`,
`lowGrade`), this one is *not* "re-map once the backend exposes it": the backend
family model has no `name` to expose.

### Verified consumer reads

|---|---|---|---|
| `useFamiliesQuery` | `useOrgQuery` → `ScoreReport` / `ProgressReport` (vestigial) | `id`, `name`, `clever` | `id` yes; `name` **no** (vestigial path); `clever` **no** (vestigial path) |

Both `name` and `clever` reads are reached only via the vestigial family-report
path described above, so neither regresses a live feature. The gap is recorded
in the `@TODO` block of `@/helpers/mapOrg.js` and in `mapFamilyToOrg`'s JSDoc.

## Consumers

No consumer required edits — the mapping keeps the flat shape stable.
`useOrgQuery.test.js` mocks `useFamiliesQuery` wholesale, so it is unaffected.

## Tests

The co-located unit test is rewritten from the Firestore stub to the
district/school/groups/classes pattern (mock `getRoarApiClient`, the auth store,
and `useQuery`; capture `queryFn` / `retry` per test):

- `useFamiliesQuery.test.js`: query key + ids ref; per-id fetch with order
  preservation + mapping (slow-resolves the 2nd id and asserts its `location`
  object is flattened to `city`, while the 1st id's `rosteringEnded` scalar
  passes through); structured non-200 throw (a 403, matching the parent-only
  authz); disabled without ids / without accessToken; becomes enabled once ids
  populate; retry predicate (false on `auth/required` + `auth/rostering-ended`,
  true transient at 0, false at MAX_RETRIES).

Legacy `fetchDocumentsById` assertions are removed.

## Verification

- Vitest runs in CI (cannot run in the authoring sandbox).

Resolves yeatmanlab/roar-project-management#1928


richford:

# Migrate useFamiliesQuery to the ts-rest backend client

## Summary

Migrate the `useFamiliesQuery` (fetch-by-IDs) read composable off the legacy Firestore data layer onto the typed ts-rest backend client (`getRoarApiClient`), mirroring the district, school, and groups/classes read migrations. This is the smallest of the set — a single by-id composable. The backend already authorizes every response, so it becomes the single source of truth for which families a user can read.

## Scope

- `useFamiliesQuery` (fetch-by-IDs) → `GET /families/:familyId` per id via `Promise.all`, preserving input order, all-or-nothing on any non-200.

## Motivation

PR4 in the dashboard read-path migration, stacked directly on the
groups/classes read migration. It applies the identical by-id pattern to the family read composable so the dashboard stops reaching into Firestore for family reads and the backend's authorization model is the only access boundary.

## Backend family model is minimal (context)

Unlike orgs and groups, `FamilyDetailSchema` (`packages/api-contract/src/v1/families/schema.ts`) is intentionally minimal: `{ id, location?, rosteringEnded? }`. Families are flat, standalone entities outside the org hierarchy, keyed by UUID. There is **no** `name`, `abbreviation`, `identifiers`, or `orgType`/`parentOrgId`.

The absence of `name` is inconsequential in practice. `useFamiliesQuery`'s **sole** consumer is `useOrgQuery` (the `families` case) → `ScoreReport.vue` / `ProgressReport.vue`, which read `orgData.name` (via `select: (data) => data[0]`).  That read sits on a **vestigial** path: a family-scoped report requires an administration assigned to a `family` org, but administrations only assign to orgs/classes/groups (the upsert body has no `families` field), so family-scoped reports are never generated in practice. The missing `name` therefore lands on a dead path, not a live feature.

## Authorization (context)

`GET /families/:familyId` is **parent-only** (FGA `can_read: parent`): only the family's caretaker or a super admin may read it. Unlike org and group reads, no admin or supervisory role grants access — admins, the family's own children, and all other users receive 403. That matches the legitimate consumer (a caretaker viewing their own family) and super admins.

## Acceptance criteria

- `useFamiliesQuery` calls the ts-rest client (`client.families.get`) and throws a structured `Error` (carrying `.status` + `.body`) on any non-200.
- Enablement is gated on `authStore.accessToken` + `hasArrayEntries(familyIds)`.  The legacy `fetchDocumentsById` / `FIRESTORE_COLLECTIONS` machinery is removed.
- The existing `FAMILIES_QUERY_KEY` is unchanged, with the reactive `familyIds` kept in the key array (by reference).
- Response shapes are mapped to the flat shape consumers read via `@/helpers/mapOrg`: a new `mapFamilyToOrg` is added (flattens the `location` OBJECT; passes `id` / `rosteringEnded` through unchanged).
- The co-located unit test is rewritten to the ts-rest pattern.
- The minimal family shape (no `name`) and its landing on the vestigial consumer path are documented (in `mapOrg.js` and the PR description) rather than silently dropped.

## Out of scope

- Backend schema changes (no new fields added to `FamilyDetailSchema`).
- `POST /families`, `POST /families/:familyId/users`, and `GET /families/:familyId/users` (caretaker registration / child enrollment / member listing) — only the by-id read is migrated here.

richford:
@claude, please re-review. For the **403 mock code** finding, this is verified: `auth/forbidden` is actually the real backend code (`ApiErrorCode.AUTH_FORBIDDEN = 'auth/forbidden'`). It's just not in the frontend's retry-relevant `API_ERROR_CODES` set, which is why it reads as transient (correct for a 403). So the mock is realistic as-is.

Emily-ejag (approved):

PR #1899, opened 2026-06-22
Migrate OrgsList to the ts-rest backend and drop unsupported columns/tabs

richford:
## Summary

This PR migrates `apps/dashboard/src/containers/OrgsList/OrgsList.vue` off
Firestore onto the already-merged ts-rest read composables (FE-PR1 through
FE-PR4), mirroring the OrgPicker rewire (FE-PR5). It is the final, meatiest PR of
the dashboard org read-path migration. Alongside the data-source swap, it applies
a set of deliberate, team-approved feature drops for fields and surfaces the
backend org schemas don't expose. `useOrgsTableQuery` — the polymorphic Firestore
query that only OrgsList consumed — is deleted.

## How the table data is now sourced

The single polymorphic `useOrgsTableQuery(activeOrgType, selectedDistrict,
selectedSchool, …)` (Firestore via `orgPageFetcher`, page-size 100000) is
replaced with the same page-walking composables OrgPicker uses, decomposed into a
per-tab `computed`:

| Tab | Before | After |
|---|---|---|
| `districts` | `useOrgsTableQuery(…)` → `orgPageFetcher('districts', …)` | `useDistrictsListQuery()` data (`allDistricts`) |
| `schools` | `useOrgsTableQuery(…)` → `orgPageFetcher('schools', …)` | `useDistrictSchoolsQuery(selectedDistrict)` data (`allSchools`) |
| `classes` | `useOrgsTableQuery(…)` → `orgPageFetcher('classes', …)` | `useSchoolClassesQuery(selectedSchool)` data (`allClasses`) |
| `groups` | `useOrgsTableQuery(…)` → `orgPageFetcher('groups', …)` | `useGroupsListQuery()` data (`allGroups`) |

```js
const orgData = computed(() => {
  switch (activeOrgType.value) {
    case ORG_TYPES.DISTRICTS:
      return allDistricts.value ?? [];
    case ORG_TYPES.SCHOOLS:
      return allSchools.value ?? [];
    case ORG_TYPES.CLASSES:
      return allClasses.value ?? [];
    case ORG_TYPES.GROUPS:
      return allGroups.value ?? [];
    default:
      return [];
  }
});
```

Each composable walks the backend pagination and returns the full set
(equivalent to the legacy page-size-100000 fetch), so the table keeps doing
client-side sort/paginate with no server-side pagination. `districts` and
`schools` reuse the data already fetched for the selectors (no duplicate fetch).
`isLoading` / `isFetching` are likewise per-tab `computed`s that reflect only the
query feeding the active tab, so the table spinner stays accurate.

### Enablement is preserved / minimal

- `schools`: `useDistrictSchoolsQuery` gates internally on `accessToken` + a
  truthy `districtId`; the existing `schoolQueryEnabled`
  (`claimsLoaded && selectedDistrict`) is passed through.
- `classes`: `useSchoolClassesQuery(selectedSchool)` gates internally on
  `accessToken` + a truthy `selectedSchool`; `classQueryEnabled`
  (`claimsLoaded && selectedSchool`) is passed through.
- `groups`: gated to the Groups tab (`claimsLoaded && activeOrgType === 'groups'`),
  mirroring OrgPicker, so the list isn't fetched on other tabs.
- `districts`: gates on `accessToken` plus the existing `claimsLoaded` option.

### 1. Four columns removed — Address, Tags, Clever, ClassLink

`useOrgTableColumns.js` no longer emits Address (`address.formattedAddress`),
Tags (`tags`), Clever (`clever`), or ClassLink (`classlink`). The backend org
schemas don't expose these (documented in `mapOrg.js`), so the columns would
always render empty. Remaining columns: **Name, Abbreviation**, **MDR Number /
NCES ID** (districts & schools only), and the action columns **Users / Edit /
SignUp Code / Export Users**.

### 2. Group invitation codes — kept for groups, dropped for orgs/classes

`showCode` previously did `fetchDocById(activeOrgType, id)` and read
`currentActivationCode`. It now calls the groups-only invitation-code endpoint:

```js
const showCode = async (selectedOrg) => {
  if (activeOrgType.value !== ORG_TYPES.GROUPS) return; // defensive backstop
  const res = await getRoarApiClient().groups.getInvitationCode({
    params: { groupId: selectedOrg.id },
  });
  if (res.status === StatusCodes.OK) {
    activationCode.value = res.body.data.code; // InvitationCodeSchema.code
    isDialogVisible.value = true;
    return;
  }
  // non-200 → error toast
};
```

(`GET /groups/:groupId/invitation-code`, response `InvitationCodeSchema` →
`{ …, code }`.) Non-200 surfaces a "No invitation code" toast; a thrown error
surfaces an "Unexpected error" toast and is reported to Sentry. `fetchDocById` is
no longer imported.

### 3. SignUp Code action dropped for non-groups

The SignUp Code column (previously super-admin-only) is now gated to
`isSuperAdmin && activeOrgType === 'groups'`. On districts/schools/classes the
column is absent, so no SignUp Code button renders and `@show-activation-code`
never fires. `showCode`'s org-type guard is a defensive backstop.

### 4. Families tab dropped

`orgHeaders` no longer includes `families` — removed from both the static
super-admin headers and the `adminOrgs.families` conditional. Families have no
list endpoint and aren't admin-managed here. Because `activeOrgType` is derived
from `Object.keys(orgHeaders)[activeIndex]`, it can never resolve to `families`
once the entry is gone, so the default tab is always a remaining org type.

## `exportAll` reworked off `orgFetchAll`

The per-tab table-data `computed` already holds the full set for the active tab,
so `exportAll` builds the CSV from `tableData.value` instead of a separate
`orgFetchAll(activeOrgType, …)` fetch. The `orgFetchAll` import is removed from
OrgsList.

## Shape compatibility — verified, no live field gap

The table reads `name`, `abbreviation`, `mdrNumber`, `ncesId` (flattened by
`mapDistrictToOrg` / `mapSchoolToOrg`), `id`, and `parentOrgId`. All are present
on the migrated/mapped shapes. The groups subgroup filter previously tested
`!org.parentOrgId && !org.parentOrgType`; the legacy `parentOrgType` is gone from
the backend shape, and a subgroup is fully identified by `parentOrgId` alone, so
the filter now tests `!org.parentOrgId`. The 4 dropped columns' fields
(`address.formattedAddress`, `tags`, `clever`, `classlink`) leave with the
columns; OrgsList reads none of them elsewhere (the export composables don't read
them either). No remaining OrgsList code reads `address` / `tags` / `clever` /
`classlink` / `parentOrgType`.

## Removed imports / deletions

- `useOrgsTableQuery` import removed; `apps/dashboard/src/composables/queries/useOrgsTableQuery.js`
  deleted (OrgsList was its only consumer; it had no co-located test).
- `orgFetchAll` (from `@/helpers/query/orgs`) and `fetchDocById` (from
  `@/helpers/query/utils`) imports removed — no longer referenced in OrgsList.
- Added: `useSchoolClassesQuery`, `useGroupsListQuery`, `getRoarApiClient`,
  `StatusCodes`, and `ORG_TYPES` (in `useOrgTableColumns.js`).
- `orderBy` / `orderByDefault` / `exportCsv` are retained — `orderBy` still drives
  the export orchestrator and `exportCsv` is still used by `exportAll`.

`@/helpers/query/orgs.js` is **not** deleted: `RegisterStudents.vue` still imports
`orgFetchAll` from it, so the helper survives. Only OrgsList's import is removed.
Migrating `RegisterStudents` and then removing the helper is a noted follow-up.

`ORGS_TABLE_QUERY_KEY` (in `constants/queryKeys.js`) is now unused after the
composable deletion, but is left in place because removing a shared export is
flagged "ask first" in the repo guardrails; it is harmless and can be pruned in a
follow-up.

## Tests

- `useOrgTableColumns.test.js` updated: base-column assertions drop Address/Tags;
  the SSO (Clever/ClassLink) describe block is removed and replaced with a
  "Dropped Columns" block asserting Address/Tags/Clever/ClassLink never appear for
  any org type; the SignUp Code tests now assert super-admin **and** groups-only
  visibility (present on groups, absent on districts/schools/classes, absent for
  non-super-admins on groups); the full-configuration and column-order
  expectations are updated to the new column set.
- No co-located OrgsList component/e2e test (`OrgsList.cy.js` / `OrgsList.test.js`)
  exists, so there is no OrgsList test harness to update. The migrated composables
  each carry their own unit test from FE-PR1 through FE-PR4.
- `useOrgsTableQuery` had **no** co-located test file (it never existed), so there
  was no `useOrgsTableQuery.test.js` to delete.

## Verification

- `npx prettier --check` passes on all changed files (`OrgsList.vue`,
  `useOrgTableColumns.js`, `useOrgTableColumns.test.js`).
- `useOrgsTableQuery`, `orgFetchAll`, `fetchDocById`, and the 4 dropped column
  fields are no longer referenced in OrgsList.
- `orgFetchAll` remains exported from `@/helpers/query/orgs.js` (only OrgsList's
  import was removed).
- Vitest / Cypress run in CI (cannot run in the authoring sandbox).

Resolves yeatmanlab/roar-project-management#1929


richford:

# Migrate OrgsList to the ts-rest backend and drop unsupported columns/tabs

## Summary

Migrate `apps/dashboard/src/containers/OrgsList/OrgsList.vue` off Firestore onto the ts-rest read composables, mirroring the OrgPicker rewire. This is the final, meatiest PR of the dashboard org read-path migration. Alongside the data-source swap, it applies a set of deliberate, team-approved feature drops for fields and surfaces the backend org schemas don't expose. `useOrgsTableQuery` (the polymorphic Firestore query that only OrgsList consumed) is deleted.

## Scope

### 1. Table data → per-tab migrated composables

Replace `useOrgsTableQuery(activeOrgType, selectedDistrict, selectedSchool, …)` (polymorphic Firestore via `orgPageFetcher`, page-size 100000) with the same page-walking composables OrgPicker uses, plus a per-tab `computed`:

- `districts` → `useDistrictsListQuery()` (`GET /districts`)
- `schools` → `useDistrictSchoolsQuery(selectedDistrict)` (`GET /districts/:districtId/schools`)
- `classes` → `useSchoolClassesQuery(selectedSchool)` (`GET /schools/:schoolId/classes`)
- `groups` → `useGroupsListQuery()` (`GET /groups`), gated to the Groups tab

Each composable walks the backend pagination and returns the full set (equivalent to the legacy page-size-100000 fetch), so the table needs no server-side pagination. The existing district/school selectors that drive `selectedDistrict` / `selectedSchool` are preserved. Delete `useOrgsTableQuery.js` and remove its import.

### 2. Drop 4 columns (deliberate)

Remove Address (`address.formattedAddress`), Tags (`tags`), Clever (`clever`), and ClassLink (`classlink`) column definitions from `useOrgTableColumns.js`. The backend org schemas don't expose these fields (`mapOrg.js` documents the gaps), so the columns would always render empty. Keep Name, Abbreviation, MDR Number, NCES ID (districts/schools), and the action columns (Users / Edit / SignUp Code / Export Users).

### 3. Group invitation codes (keep groups, drop orgs/classes)

Rewrite `showCode(selectedOrg)` from `fetchDocById(activeOrgType, id)` + `currentActivationCode` to call the backend invitation-code endpoint, which exists only for groups: `getRoarApiClient().groups.getInvitationCode({ params: { groupId } })`, then read `res.body.data.code` (`InvitationCodeSchema`). Non-200 / errors surface a toast.

### 4. Drop the activation-code (SignUp Code) action for non-groups

Gate the SignUp Code column (previously super-admin-only) to `isSuperAdmin && activeOrgType === 'groups'`, so districts/schools/classes show no SignUp Code action.

### 5. Drop the Families tab

Remove the `families` entry from `orgHeaders` (both the static super-admin headers and the `adminOrgs.families` conditional). Families have no list endpoint and aren't admin-managed here. With families gone from `orgHeaders`, `activeOrgType` (derived from `Object.keys(orgHeaders)[activeIndex]`) can never be `families`.

### 6. Rework `exportAll` off `orgFetchAll`

The per-tab table-data `computed` already holds the full set, so `exportAll` builds the CSV from that instead of a separate `orgFetchAll` fetch. Remove the `orgFetchAll` import from OrgsList.

## Motivation

OrgsList was the last live consumer of `useOrgsTableQuery` and one of the last consumers of `orgFetchAll` / `fetchDocById` in the org UI, reaching directly into Firestore and re-implementing claims-based client-side filtering. The migrated composables already encapsulate the ts-rest calls, pagination walking, response mapping, retry policy, and defer authorization to the backend. The dropped columns/tabs/codes are surfaces the backend deliberately does not model.

## Shape compatibility

The table reads `name`, `abbreviation`, `mdrNumber`, `ncesId` (flattened by the mappers), `id`, and `parentOrgId` (the groups subgroup filter). All are present on the migrated/mapped shapes. The legacy `parentOrgType` field is gone; a subgroup is identified by `parentOrgId` alone, so the subgroup filter drops the `parentOrgType` check. The 4 dropped columns' fields go away with the columns.

## Out of scope

- Deleting `@/helpers/query/orgs.js` — `RegisterStudents.vue` still imports `orgFetchAll`, so the helper survives. Only OrgsList's import is removed. A follow-up can migrate `RegisterStudents` and then remove the helper.
- Any backend or contract changes (dashboard-only; consumes endpoints already shipped).

richford:
@claude, please re-review. `error.message` in the toast matches the existing pattern in the same file so I left it.

Emily-ejag (approved):

PR #1901, opened 2026-06-22
Migrate the org EDIT flow off Firestore/roarfirekit to the ts-rest backend

richford:
## Summary

This PR migrates the dashboard's org **edit** flow off Firestore/`roarfirekit`
and onto the typed ts-rest backend client, using the org `PATCH` update
endpoints. The edit form now reads through the already-migrated by-id read
composables and the Save action writes through a new `useUpdateOrgMutation`,
which invalidates the org reads so the OrgsList tables and the edited org both
refetch.

## Read path

`EditOrgsForm.vue` previously read the org via
`useQuery(['org', orgType, orgId], () => fetchDocById(orgType, orgId))`, gated
on `roarfirekit.restConfig`. It now reads through `useOrgQuery`, the migrated
by-id dispatcher (ts-rest backend), selecting the single org out of the
returned array (`select: (data) => data?.[0]`).

`useOrgQuery` dispatches by the **singular** org type, while the form receives
the **plural** type from OrgsList's active tab, so the form converts it with
`singularizeFirestoreCollection`. The read composables gate internally on
`authStore.accessToken`, so the firekit init machinery
(`roarfirekit`/`restConfig` `$subscribe`/`onMounted` gate/`initialized`) has
been removed from the form entirely.

The address is shown via the `GMapAutocomplete` placeholder, built from the
org's structured `location` (which `mapOrg` flattens to top-level fields):
`[addressLine1, city, stateProvince, postalCode].filter(Boolean).join(', ')`.
`localOrgData.address` starts `null` and is only set when the user picks a new
place via `setAddress` (unchanged).

## Write path

The Save button in `OrgsList.vue` previously called
`roarfirekit.value.createOrg(activeOrgType, localOrgData, testData, demoData, id)`.
It now builds a typed PATCH body and calls
`useUpdateOrgMutation().mutateAsync({ orgType, orgId, body })`.

`useUpdateOrgMutation` mirrors `useUpsertAdministrationMutation`: it grabs the
client via `getRoarApiClient()`, dispatches by org type to the correct
`client.<resource>.update({ params, body })` (mapping the differing path-param
keys — districts `id`, schools `schoolId`, classes `classId`, groups
`groupId`), throws a structured error on a non-200 response, and returns
`result.body.data`. On success it invalidates every org read key
(`DISTRICTS`/`DISTRICTS_LIST`/`DISTRICT_SCHOOLS`/`SCHOOLS`/`SCHOOL_CLASSES`/
`CLASSES`/`GROUPS`/`GROUPS_LIST`) so both the tables and the by-id reads
refetch.

## Google Places → structured location parser

`parseGooglePlaceToLocation` (`apps/dashboard/src/helpers/`) is a pure function
mapping the Google Places object `setAddress` builds (carrying
`addressComponents`, the raw Google `address_components[]`) to the backend's
structured `location` (`{ addressLine1, addressLine2, city, stateProvince,
postalCode, country }`):

- `addressLine1` = `street_number` + ' ' + `route` (long names)
- `addressLine2` = `subpremise` (long name)
- `city` = `locality`, falling back to `postal_town` then `sublocality`
- `stateProvince` = `administrative_area_level_1` (short name)
- `postalCode` = `postal_code`
- `country` = `country` (short name)

It returns a **partial** object (missing components are omitted, never emitted
as empty strings) and `{}` for a missing/empty `addressComponents`. Unit-tested
with a representative `address_components` array, the partial cases, the
city-fallback chain, and the empty cases.

## Per-org-type PATCH body

The body includes only the fields valid for each org type (per each resource's
`Update<Resource>RequestSchema`) and only when present/changed:

- `name`: every org type, when present.
- `abbreviation`: districts/schools/groups only — **never for classes** (classes
  have no abbreviation column).
- `location` (structured): districts/schools/groups only, and only when the user
  picked a new address — **excluded for classes**, whose `location` is a
  free-text room label string, incompatible with the structured object the
  autocomplete produces.
- `identifiers.ncesId`: districts/schools only, when present.

The retired `tags`, `testData`, and `demoData` are never sent, and their inputs
were removed from the form (template + `localOrgData` + `setupOrgData`).

## Field visibility in the form

- `abbreviation` input hidden for classes.
- NCES ID input shown for districts only (unchanged from the legacy form).

## Tests

- `parseGooglePlaceToLocation.test.js` — full mapping, long/short name choice,
  partial fields, city fallback, empty input.
- `useUpdateOrgMutation.test.js` — per-org-type dispatch with the correct path
  param, unsupported-type throw, non-200 structured error, and the full
  invalidation set on success.

CI runs the Vitest suites, lint, and type-check (the local sandbox can't run
them); `npx prettier --check` passes on every changed file.

Resolves yeatmanlab/roar-project-management#1931


richford:

# Migrate the org EDIT flow off Firestore/`roarfirekit` to the ts-rest backend

## Problem

The org **edit** flow in the dashboard still reads and writes through the legacy Firestore stack:

- `EditOrgsForm.vue` reads the org with `useQuery(['org', orgType, orgId], () => fetchDocById(orgType, orgId))`, gated behind `roarfirekit.restConfig`, and edits a local copy that includes retired fields (`tags`, `testData`, `demoData`).
- `OrgsList.vue`'s Save button calls `roarfirekit.value.createOrg(...)` — a Firestore upsert keyed by the org id.

The backend now exposes typed PATCH update endpoints for all four admin-managed org types (`PATCH /districts/:id`, `/schools/:schoolId`, `/classes/:classId`, `/groups/:groupId`), so the edit flow should move onto the ts-rest client like the rest of the org surface (reads, OrgPicker, OrgsList tables) already has.

## Scope

- Read: `EditOrgsForm` should read the org via the migrated by-id read composables (`useOrgQuery`), not Firestore.
- Write: the Save action should PATCH via a new `useUpdateOrgMutation`, not `roarfirekit.createOrg`.
- Drop the retired `tags`, `testData`, and `demoData` inputs.
- Respect each org type's update schema: classes have no `abbreviation`; only districts/schools carry NCES identifiers; classes' `location` is a free-text room label (a string), not a structured address.

## Out of scope

- The org **create** flow (`CreateOrgs.vue`) — separate migration.
- Any backend changes — the PATCH endpoints already exist.
- The Families tab (already dropped from OrgsList; no list endpoint).

## Constraints

- Follow the container/presentational + composable patterns (`frontend-layer-architecture.md`, `frontend-composable-patterns.md`, `frontend-state-management.md`): the mutation invalidates the org reads; presentational form receives data through the read composable, container owns the write.
- Pure address-parsing logic must be a unit-tested helper.

richford:
@claude, please re-review

richford:
@claude, please re-review

kellyel (approved):

PR #1905, opened 2026-06-22
Fix emulator first-navigation bounce by reading super-admin from /me

richford:
## Summary

This PR fixes the local-emulator bug where the first in-app navigation bounces back to the
current page and only works on the second click. The router's `beforeEach` guard now recognizes
super admins from the `/me` payload it already awaits, instead of depending on a separately-timed
claims source.

## Root cause

The guard (`apps/dashboard/src/router/index.js`) gates routes via `userCan(to.meta.permission)`.
On the emulator, `userCan` grants access only when `authStore.isUserSuperAdmin` is true (the
emulator token has no role claim, so `PermissionsService` otherwise denies as `guest`).
`isUserSuperAdmin` derives from `authStore.userClaims`, which is set by a **separate** async
`resolveUserClaims('/me')` call in `App.vue`'s `onBeforeMount` — not the same `/me` the guard
`await`s (`ensureQueryData([ME_QUERY_KEY])`). On the first navigation the two async `/me` flows
race: the guard runs the permission check before `userClaims` lands, evaluates the super admin as
forbidden, and the navigation is bounced. The second click — claims now settled — passes.

## Change

The guard already awaits the `/me` payload into a local `meData`, and `/me` returns
`isSuperAdmin`. The permission check now reads super-admin from that already-awaited source
(falling back to `store.isUserSuperAdmin`) and short-circuits the check for super admins:

```js
const isSuperAdmin = Boolean(meData?.isSuperAdmin) || store.isUserSuperAdmin;
if (Object.keys(to?.meta).includes('permission') && !isSuperAdmin && !userCan(to.meta.permission)) {
  // …Unauthorized
}
```

This removes the dependency on the late `userClaims` from the navigation-critical path. Non-super
-admin gating and production behavior are unchanged.

## Testing

Verified manually on `dev:local` (there is no router-guard unit-test harness in the repo). The
fix is scoped to the guard; `usePermissions.js` and `App.vue` are untouched.

## Follow-up (out of scope)

The deeper fix is to retire the second `/me` source: derive `userClaims` / `isUserSuperAdmin`
from the canonical `useMeQuery` rather than a separate `resolveUserClaims` fetch, so the two can
never disagree on timing. That's a larger refactor tracked separately.

Resolves yeatmanlab/roar-project-management#1938


richford:

## Problem

On `npm run dev:local` (Firebase Auth emulator + `server-test` backend), the first time you
navigate to a different route the app routes you **back to your current page**. The navigation
only succeeds when you click the routing control a **second** time.

## Steps to reproduce

1. `npm run dev:local`.
2. Sign in as a **super admin**.
3. From any page, click a nav link/button to a different (permission-gated) route.
4. Observe: you stay on / are returned to the current page.
5. Click the same control again → the navigation now works.

## Root cause

The router's global `beforeEach` guard (`apps/dashboard/src/router/index.js`) gates routes on
`userCan(to.meta.permission)`. On the emulator, `userCan` grants access only when
`authStore.isUserSuperAdmin` is true — otherwise the emulator-issued token carries no role claim
and `PermissionsService` treats the user as `guest` and denies. `isUserSuperAdmin` reads
`authStore.userClaims.claims.super_admin`, which is populated by a **separate** async call,
`resolveUserClaims(uid)` (→ `/me`) in `App.vue`'s `onBeforeMount` — distinct from the `/me` the
guard itself awaits (`ensureQueryData([ME_QUERY_KEY])`). On the first navigation those two async
`/me` flows haven't both settled, so the guard evaluates the super admin as forbidden and the
navigation is bounced. By the second click `userClaims` has landed and it passes.

## Expected

The first navigation should succeed; a super admin should never be transiently treated as
unauthorized.

## Acceptance criteria

- On `dev:local`, the first navigation to a permission-gated route works for a super admin (no
  second click needed).
- Non-super-admin route gating is unchanged.
- No change to production behavior.

kellyel (approved):
Nice catch!

PR #1912, opened 2026-06-24
Migrate admin user view + edit to GET/PATCH /v1/users/:id

richford:
## Summary

This PR migrates the admin user view + edit surface (`UserInfoView` + `EditUsersForm`) off
Firestore/Firekit onto the API — read via a new `useUserProfileQuery` (`GET /v1/users/:id`) and write
via `PATCH /v1/users/:id` — in a single PR so the edit form never reads from the API while saving to
Firestore. It also reworks the free-reduced-lunch control to the contract enum. Stacked on #1941.

## What changed

- **`composables/queries/useUserProfileQuery.js`** (new, + tests) — typed ts-rest client (`GET /v1/users/:id`)
  reshaped via `mapUser`, keyed `'user-profile'` to avoid colliding with the Firestore `'user'` cache.
- **`composables/mutations/useUpdateUserMutation.js`** — `roarfirekit.updateUserData` →
  `client.users.update({ params: { id }, body })`; 204-aware; invalidates the profile (and `'user'`) keys.
- **`components/views/UserInfoView.vue`** — reads via `useUserProfileQuery`.
- **`components/EditUsersForm.vue`** — submit maps the form model to `UpdateUserRequestBodySchema` via the
  new `helpers/mappers/mapUserFormToUpdateBody.js`; the FRL control is now a `Free | Reduced | Paid`
  (+ None → `null`) select instead of a boolean, and `statusFrl` flows through the mapper.
- `tags` / `testData` / `demoData` are not written.

## Notes

- `UserInfoView` is admin-permission-gated; the admin holds FGA `can_read` / update rights on the target
  user, so the API calls are authorized. The orphaned `modals/EditUsers.vue` is handled in #1944.

Resolves yeatmanlab/roar-project-management#1942


richford:

## Summary

Migrate the admin user-management surface (`UserInfoView` + `EditUsersForm`) off Firestore/Firekit
onto the API: read via a new `useUserProfileQuery` (`GET /v1/users/:id` + `mapUser`) and write via
`PATCH /v1/users/:id`. **Read and write migrate in the same PR** — splitting them would leave the
edit form reading from the API (Postgres) while saving to Firestore (Firekit), so every save would
display stale data.

## Scope

In scope:

- New `composables/queries/useUserProfileQuery.js` (+ test) — API + `mapUser`, keyed on a new
  `USER_PROFILE_QUERY_KEY` (`'user-profile'`) to avoid colliding with the Firestore `'user'` cache.
- `composables/mutations/useUpdateUserMutation.js` — `roarfirekit.updateUserData` →
  `client.users.update({ params: { id }, body })`; invalidate `USER_PROFILE_QUERY_KEY` (and `'user'`
  for any Firestore consumers still mounted).
- `components/views/UserInfoView.vue` — read via `useUserProfileQuery` instead of `useUserDataQuery`.
- `components/EditUsersForm.vue` — display from the mapped shape (already compatible); on submit, map
  the form model to `UpdateUserRequestBodySchema` (nested → flat, snake_case → camelCase).
- Tests for the query, mutation, and form mapping.

Out of scope: org-user lists (F3), export (F4), CSV upload (F5). The Firestore `useUserDataQuery`
remains for the portal/consent/SSO/report consumers (see F1 scope boundary).

## Notes

- `UserInfoView` is permission-gated to admins (`Permissions.Users.EDIT`); the admin holds FGA
  `can_read` / update rights on the target user, so the API calls are authorized.
- Retired fields: `EditUsersForm` currently renders super-admin `testData` / `demoData` checkboxes —
  drop these from the write body (they are being removed platform-wide). Confirm the field is gone
  from `UpdateUserRequestBodySchema` before wiring.
- Keep PR under the size limit; if read + write together exceed it, ship as two branches stacked and
  merged together (never read-only alone).

## Acceptance criteria

- [ ] `useUserProfileQuery` fetches via the typed client (key `'user-profile'`); unit-tested
- [ ] `UserInfoView` reads via `useUserProfileQuery`; no `useUserDataQuery` import
- [ ] `useUpdateUserMutation` calls `client.users.update`; invalidates the profile (and `'user'`) keys
- [ ] `EditUsersForm` submit produces a body valid against `UpdateUserRequestBodySchema`
- [ ] `tags` / `testData` / `demoData` not written
- [ ] No API-read / Firestore-write split: editing a user reflects immediately
- [ ] Tests updated; lint + format + type-check clean

## Branch

`enh/<ticket>/user-view-edit-migration`, stacked on F1.

Parent: #1897
Blocked by: #1941

richford:
@claude, please re-review.

apps/dashboard/src/components/EditUsersForm.vue
```diff
@@ -292,17 +258,18 @@ const setupUserData = () => {
...
+      // through serializeDob (local components) as the wrong date — corrupting it on save.
+      dob: parseDateLocal(props.userData?.studentData?.dob),
       grade: props.userData?.studentData?.grade || '',
       gender: props.userData?.studentData?.gender || '',
       race: props.userData?.studentData?.race || [],
       hispanic_ethnicity: props.userData?.studentData?.hispanic_ethnicity || false,
       ell_status: props.userData?.studentData?.ell_status || false,
-      frl_status: props.userData?.studentData?.frl_status || false,
+      frl_status: props.userData?.studentData?.frl_status || null,
       iep_status: props.userData?.studentData?.iep_status || false,
```
Emily-ejag: the API returns strings ('true'/'false'), but the form dropdown expects boolean values. Coerce explicitly to avoid type mismatches.

```
      ell_status: String(props.userData?.studentData?.ell_status) === 'true' ? true : false,
      frl_status: props.userData?.studentData?.frl_status || null,
      iep_status: String(props.userData?.studentData?.iep_status) === 'true' ? true : false,`
```

apps/dashboard/src/components/views/UserInfoView.vue
```diff
@@ -74,31 +71,40 @@ onMounted(() => {
...
-      toast.add({
-        severity: 'error',
-        summary: 'Unexpected Error',
-        detail: 'An unexpected error has occurred.',
-        life: 3000,
-      });
+  try {
+    // Map the form's nested model to the flat `UpdateUserRequestBodySchema`
+    // body before writing, so the read and write stay on the same API source.
+    const body = mapUserFormToUpdateBody(localUserData.value);
```
Emily-ejag, line 95: User clicking "Update" without editing anything sends empty {} body, which API rejects silently

```suggestion
    const body = mapUserFormToUpdateBody(localUserData.value);
    
    if (!body || Object.keys(body).length === 0) {
      toast.add({
        severity: 'warn',
        summary: 'No Changes',
        detail: 'Please edit at least one field before updating.',
        life: 3000,
      });
      isSubmitting.value = false;
      return;
    }
```

apps/dashboard/src/helpers/mappers/mapUserFormToUpdateBody.js
```diff
@@ -0,0 +1,153 @@
...
+/**
+ * Coerces a boolean status flag from the form to the nullable string the
+ * contract expects for `statusEll` / `statusIep`. Leaves nullish values as
+ * `null` so they don't clobber an existing value with the string `"null"`.
+ *
+ * @param {boolean|string|null|undefined} value – The status value from the form.
+ * @returns {string|null} The string representation, or `null`.
+ */
+function coerceBooleanStatus(value) {
+  if (value === null || value === undefined) return null;
```
Emily-ejag, line 135: The form expects boolean values (true/false), and the API stores them as strings ('true'/'false'). This function normalizes boolean input to the string format the API expects. On load, `EditUsersForm.setupUserData `must normalize the returned strings back to booleans for form binding.

```suggestion
  if (value === null || value === undefined) return null;
  // Explicitly handle boolean → 'true'/'false' string conversion
  if (typeof value === 'boolean') return value ? 'true' : 'false';
```

Emily-ejag (changes requested):

apps/dashboard/src/components/EditUsersForm.vue
richford: Good catch. I coerced both with String(...) === 'true' (dropped the redundant ? true : false). Left frl_status as-is since it's the Free|Reduced|Paid enum, and hispanic_ethnicity since mapUser already returns it as a boolean. (See 3d842b559)

apps/dashboard/src/helpers/mappers/mapUserFormToUpdateBody.js
richford: Added the explicit typeof value === 'boolean' → 'true'/'false' conversion. I kept the trailing return String(value) as a fallback so a value that's already a string passes through rather than returning undefined (the previous String(value) produced the right result for booleans too, but the explicit branch makes the intent clear). (See 7bef81eaa)

apps/dashboard/src/components/views/UserInfoView.vue
richford: Added the guard. One thing to note: mapUserFormToUpdateBody always returns a full body (every field, defaulting to the loaded values), so Object.keys(body).length === 0 only happens when the form isn't initialized. It prevents the empty-{} 400, but a true no-op "Update" still sends the unchanged values (which the API accepts as a no-op). I also relied on the existing finally to reset isSubmitting on the early return rather than resetting it inline. (See 3397bfc97)

Emily-ejag (approved):
LGTM

PR #1915, opened 2026-06-24
Migrate org-users CSV export to listUsers with ?embed=demographics

richford:
## Summary

This PR migrates the org-users CSV export off Firestore onto the typed `listUsers` endpoint with the new
`?embed=demographics` option, paging through all users server-side and preserving the exact CSV column
set, order, and formatting. Stacked on #1943.

## What changed

- **`containers/OrgsList/composables/useOrgExport.js`** — rewritten onto the typed client: `fetchAllOrgUsers`
  dispatches by org type (shared `helpers/orgUsersResolvers.js`, extracted from F3), requests `perPage: 100`
  + `embed: 'demographics'`, and loops all pages (no N+1); the count comes from `pagination.totalItems`
  (replacing the Firestore aggregation).
- **`helpers/mappers/mapEnrolledUserForExport.js`** (new, + tests) — maps an API row (plus its
  `demographics` sub-object) to the 14-column CSV object; `race` is kept as the comma-joined scalar string
  and `hispanic_ethnicity` as the raw boolean to match the prior Firestore output byte-for-byte.
- Deletes the now-dead Firestore helpers (`fetchUsersByOrg` / `countUsersByOrg` / `getUsersRequestBody`).

## Notes

- Runtime-depends on the demographics embed (#1946) being deployed; the branch parent is #1943.
  `tags` / `testData` / `demoData` are not exported. Batch/progress/cancel UX and warning thresholds are
  unchanged — only the data source changed.

Resolves yeatmanlab/roar-project-management#1945


richford:

## Summary

Migrate the org-users CSV export (`useOrgExport` + `useOrgExportOrchestrator`) off Firestore
onto the same `listUsers` endpoints F3 adopts. Export is just a full-table read, so it reuses
F3's query path — paging through (or streaming) all accessible users rather than reading
Firestore documents directly.

## Scope

In scope:

- `containers/OrgsList/composables/useOrgExport.js` — source rows from the typed `listUsers`
  endpoint instead of Firestore; preserve the existing CSV column set and ordering.
- `containers/OrgsList/composables/useOrgExportOrchestrator.js` — orchestrate paged fetches to
  assemble the full export (respect the endpoint's max `perPage`; loop pages or use the streamed
  variant).
- Tests for the orchestrator's paging/assembly.

Out of scope: the list view itself (F3), bulk import (F5).

## Notes

- Performance: page through with bulk reads — do not fetch users one-by-one
  (`performance-avoid-quadratic.md`). Map each page through `mapUser`, accumulate, then serialize.
- Column parity: keep the current CSV header order and field formatting so downstream consumers
  of the export are unaffected. Diff a sample export against the Firestore version before merge.

## Acceptance criteria

- [ ] Export sources rows from the typed `listUsers` endpoint; no Firestore reads
- [ ] Full export assembled via paged/streamed reads (no N+1)
- [ ] CSV columns, order, and formatting match the pre-migration output
- [ ] `tags` / `testData` / `demoData` not exported
- [ ] Orchestrator paging tested; lint + format + type-check clean

## Branch

`enh/<ticket>/org-users-export-migration`, stacked on F3.

Parent: #1897
Blocked by: #1943

richford:
@claude, please re-review.

apps/dashboard/src/containers/OrgsList/composables/useOrgExport.js
```diff
@@ -1,30 +1,134 @@
...
+   * @returns {Promise<{ items: object[], pagination: object }>} The page payload.
+   */
+  const fetchUsersPage = async (resolver, orgId, page) => {
+    const result = await resolver.action({
+      params: { [resolver.paramKey]: orgId },
+      query: {
+        page,
+        perPage: EXPORT_FETCH_PER_PAGE,
+        // Wire format is a comma-separated string; this is the single embed option.
+        embed: 'demographics',
```
Emily-ejag, line 85: I think the API contract has NO support for: EnrolledUserDemographicsSchema (the schema the API should return)
demographics field on EnrolledUserSchema embed parameter on the listUsers query schema

SO, I think this silent data loss, and the export completes successfully but some CSV columns will be empty like UserType, ell_status , iep_status , frl_status , etc.

Emily-ejag (changes requested):

apps/dashboard/src/containers/OrgsList/composables/useOrgExport.js
richford: You're right. This PR depends on #1916. So as long as we merge that one first, this should take care of your concern.

Emily-ejag (dismissed):

Emily-ejag (dismissed):

Emily-ejag (approved):

PR #1916, opened 2026-06-24
Add ?embed=demographics to org user-list endpoints

richford:
## Summary

This PR adds an opt-in `?embed=demographics` to the org user-list endpoints
(`GET /v1/{districts|schools|classes|groups}/:id/users`) so the CSV export (#1945) can fetch demographic
columns the lean base list omits, without bloating the default response. The demographics are student PII
returned only when explicitly requested. Off `project/backend-refactor`.

## What changed

- **Contract** (`common/user.ts`) — `createEmbedQuerySchema(['demographics'])` on the org
  `EnrolledUsersQuerySchema`; an optional `demographics` sub-object on `EnrolledUserSchema` (`userType`,
  `statusEll`, `statusFrl`, `statusIep`, `race`, `hispanicEthnicity`, `homeLanguage`), sourced from the
  user-profile field set/naming. Scoped to the org query — families reuse the base query but return a
  demographics-free shape, so the embed is org-only.
- **Repository** — shared `buildEnrolledUserSelection(embedDemographics)` selects the demographic columns
  only when embedded (the default query never fetches them); bulk-selected with the page (no N+1).
- **Service** (all four org types) — resolves the embed after the main query; authorization unchanged.
- **Controller** — includes `demographics` in the transform only when present.

## Tests & verification

Unit (transform, query-helper, four org services, district controller) + integration (district repo +
route) asserting the base list omits demographics, the embed includes them null-safe, and **the embed does
not widen the result set**. Backend `tsc --noEmit` passes.

## Notes

CRITICAL-tier: the embed exposes student PII only on explicit `?embed=demographics`; the FGA org scoping is
unchanged and remains the sole authority over which rows are returned.

Resolves yeatmanlab/roar-project-management#1946


richford:

## Summary

Add an opt-in `?embed=demographics` to the org user-list endpoints
(`GET /v1/{districts|schools|classes|groups}/:id/users`) so the CSV export can fetch demographic columns
the lean base list omits, without bloating the default response. The base `EnrolledUser` returns identity
+ per-org roles; the demographics embed adds a `demographics` sub-object with `userType`, `statusEll`,
`statusFrl`, `statusIep`, `race`, `hispanicEthnicity`, `homeLanguage` (student PII), present only when
requested.

## Scope

In scope:

- Contract (`packages/api-contract/src/v1/common/user.ts`): `createEmbedQuerySchema(['demographics'])`
  on the org `EnrolledUsersQuerySchema`; an optional `demographics` sub-object on `EnrolledUserSchema`,
  sourced from the existing user-profile field set/naming. Scoped to the org query — families reuse the
  base query but return a demographics-free shape, so the embed is org-only.
- Repository: select the demographic columns only when the embed is requested (default query unchanged);
  bulk-selected with the page (no N+1).
- Service: resolve the embed after the main query; authorization unchanged (the embed only adds columns
  to already-authorized rows).
- Controller: include `demographics` in the transform when present.
- Unit + integration tests (base omits / embed includes / null-safe / no result-set widening).

Out of scope: the frontend export consumer (F4); families.

## Notes

- CRITICAL-tier: the embed exposes student PII only on explicit `?embed=demographics`; the FGA org
  scoping is the sole authority over which rows are returned and is unchanged.
- Backend foundation for the CSV export (F4), which requests this embed at runtime.

## Acceptance criteria

- [ ] Base list shape unchanged without the embed
- [ ] `?embed=demographics` returns the demographic sub-object; field names/nullability match the user profile
- [ ] Repository selects demographics only when embedded; no N+1
- [ ] Authorization unchanged; an integration test asserts the embed does not widen the result set
- [ ] Scoped to org endpoints (families unaffected)
- [ ] Tests pass; lint + format + type-check clean

## Branch

`enh/<ticket>/demographics-embed`, off `project/backend-refactor`.

Parent: #1897

richford:
@claude, please re-review.

Emily-ejag (changes requested):
I don't see a `home_language: user.homeLanguage` on `mapUser.js` so I think it needs to be added as part of the scope of this PR

richford:
@Emily-ejag, thanks for catching that. I fixed it in this PR rather than a follow-up. `home_language` now round-trips end-to-end on the single-user path, and empty free-text demographics save as `null`.

On scope: `mapUser.js` is the mapper for the single-user `GET /v1/users/:id` response, not the org-list `?embed=demographics` this PR adds. #1946's embed already carries `homeLanguage` (it's in `EnrolledUserDemographicsSchema`, and `mapEnrolledUserForExport` already maps `home_language: demographics.homeLanguage`). So the gap was in the adjacent single-user read mapper, which mapped every other demographic but dropped `home_language`. Since it's small and clearly related, I closed it here:

- `mapUser` now maps `home_language` on read.
- `EditUsersForm` had no Home Language control, so adding `homeLanguage` to `mapUserFormToUpdateBody` alone would have sent `homeLanguage: null` on every save and silently cleared it. So I added a Home Language input to the student edit section (mirrors the Gender field).
- `mapUserFormToUpdateBody` maps `home_language` to `homeLanguage` on write.
- Empty free-text fields now normalize to `null` (cleared) instead of `''`, via a small `emptyToNull` helper applied to both `gender` and `homeLanguage`; booleans like `hispanicEthnicity` keep `?? null` so a real `false` survives.
- Unit tests added on both mappers.

No contract change was needed since `UpdateUserRequestBodySchema` already accepts `homeLanguage`.

@claude, please re-review this PR.

Emily-ejag (approved):

PR #1928, opened 2026-06-24
Cut the Progress Report off Firestore onto the progress API

richford:
## Summary

This PR migrates the dashboard **Progress Report** off Firestore/firekit onto the
existing ts-rest progress endpoints. The backend (`report.service` /
`report.repository`, the `/reports/progress/students` + `/overview` routes, and
their unit + integration tests) already lives on `project/backend-refactor`, so this
is a frontend cutover — no new backend report logic. Score Reports are untouched.

## What changed

**New query composables** (`apps/dashboard/src/composables/queries/`)

- `useAdministrationProgressQuery` — `GET /v1/administrations/:id/reports/progress/students`,
  scoped by `scopeType` + `scopeId`, follows the response pagination and returns
  domain data `{ students, tasks, exclusions }`.
- `useAdministrationProgressOverviewQuery` — `GET .../reports/progress/overview`,
  returns the aggregated overview object.

Both follow the canonical pattern: internal token gating
(`Boolean(authStore.accessToken)`), `computeQueryOverrides`, success-envelope
unwrapping, structured `.status`/`.body` error throws, and the retry policy pinned
after `...options` (`isRosteringEndedError` / `isTerminalAuthError`, `MAX_RETRIES = 3`).
Keys are centralized in `constants/queryKeys.js`.

**Container + presentational composables** (`containers/ProgressReport/`)

- `ProgressReport.vue` now drives off the two new queries; `orgType`/`orgId` map to
  `scopeType`/`scopeId`.
- `useProgressData` consumes the API student rows and maps the backend's 7-level
  `ProgressStatus` to the report's display tags. Progress is keyed by task **UUID**
  (the legacy Firestore path keyed by slug).
- `useProgressColumns` builds task columns from the self-describing `tasks[]`
  metadata (UUID field paths; friendly headers still via the task dictionary by slug).
- `useProgressExport` resolves per-task CSV headers from the `tasks[]` metadata.
- `ProgressStatsOverview` consumes the overview response (`byTask` + per-student
  assignment buckets); its now-unused `administrationData` prop was dropped.

**Seed + e2e**

- `server-test.ts` surfaces a ready-made `progress.schoolA` scenario in the Cypress
  fixture (administration + School A scope + the seeded completed/started student
  IDs), reusing the completion runs already seeded for local dev.
- `cypress/e2e/smoke/progressReport.cy.js` — backend-only spec that signs in as the
  seeded School A admin and asserts `progress/students` + `progress/overview` reflect
  the seeded completion, plus a 401 for an unauthenticated request.

## Status mapping

The backend's orthogonal status (`assigned`/`started`/`completed` × `required`/`optional`)
collapses to the report's single per-task tag, preserving the legacy precedence —
**any optional task shows "Optional"**, otherwise the progress state:

| Backend status | Report tag |
|---|---|
| `*-optional` | Optional |
| `completed-required` | Completed |
| `started-required` | Started |
| `assigned-required` | Assigned |

## District roll-up enabled

All scopes — district, school, class, and group — load per-student progress. The
legacy Firestore report fetched no per-student data at the district level (a
client-side fetch-everything limitation, not a product decision); with server-side
pagination there's no reason to exclude it, so the district view now renders the full
per-school roll-up. The School column, school filter, and school-first sort were
already wired for district and light up now that the table has rows.

## Testing

- Unit tests for both query composables (key composition, pagination follow,
  structured error throw, retry short-circuit, enablement) mirroring
  `useTaskVariantsByTaskQuery.test.js`.
- Updated `useProgressData` / `useProgressExport` tests and a new `useProgressColumns`
  test (priority ordering, UUID field paths, presence/PID/school columns).
- Backend-only Cypress smoke spec over the seeded stack.
- `prettier --check` and `node --check` clean; relying on CI for `check-types` and the
  Vitest/Cypress runs (no arm64 native binaries locally).

Resolves [1956](https://github.com/yeatmanlab/roar-project-management/issues/1956)


richford:

## Summary

The dashboard **Progress Report** still reads assignment + completion data from
Firestore (`useAdministrationAssignmentsQuery` → `assignmentFetchAll`) and firekit
(`useAdministrationsStatsQuery` → `roarfirekit.getAssignmentStats`). The backend
progress endpoints already exist on `project/backend-refactor`
(`report.service.ts`, `report.repository.ts`, full unit + integration tests), so
this is a **frontend cutover** onto already-built endpoints — no backend report
work required.

Cut the Progress Report's two data queries over to:

- `GET /v1/administrations/:id/reports/progress/students` — paginated per-student
  progress, scoped via `scopeType` + `scopeId`, with self-describing `tasks[]`
  metadata and a 6-level visible `ProgressStatus` per task.
- `GET /v1/administrations/:id/reports/progress/overview` — aggregated per-task and
  per-student completion counts.

## Scope

In scope:

- New query composables (`useAdministrationProgressQuery`,
  `useAdministrationProgressOverviewQuery`) following the canonical token-gated /
  envelope-unwrapping / structured-error / pinned-retry pattern.
- Re-point `ProgressReport.vue` and its `useProgressData` / `useProgressColumns` /
  `useProgressExport` composables and the `ProgressStatsOverview` component at the
  new API shape (progress keyed by task **UUID**, not slug; status remapped from the
  backend's 7-level enum to the report's display tags).
- A progress-report scenario surfaced in the `server-test` Cypress fixture plus a
  backend-only smoke spec asserting the endpoints reflect the seeded completion.
- **Enable district-scope roll-up.** The legacy report fetched no per-student data at
  the district level (a client-side fetch-everything limitation); with server-side
  pagination there's no reason to exclude it, so all scopes — district, school, class,
  and group — load. The School column, school filter, and school-first sort were
  already wired for district and now render.

Out of scope (explicitly):

- **Score Reports** and any shared score-report consumers — they stay on firekit.
- The `restConfig` auth-bootstrap gate in the container — removed platform-wide when
  auth migrates.

## Acceptance criteria

- [ ] Progress Report renders per-student progress + the stats overview for district,
      school, class, and group scopes from the ts-rest API (no Firestore/firekit reads).
- [ ] District scope renders the per-school roll-up (School column + school filter),
      not just a header.
- [ ] Per-task status tags (Assigned / Started / Completed / Optional) match the
      prior behavior, including the rule that optional tasks always show "Optional".
- [ ] CSV export and tag-based column filtering keep working with UUID-keyed progress.
- [ ] Unit tests cover the new composables (key, pagination, error throw, retry,
      enablement) and the status remap.
- [ ] A backend-only Cypress spec exercises `progress/students` + `progress/overview`
      against the seeded stack.

richford:
@claude, please re-review

richford:
@claude, please re-review

richford:
@claude, please re-review. I believe your issue number 6 was resolved on the progress report but not yet on the score report. The score report is out of scope for this PR.

apps/dashboard/src/containers/ProgressReport/ProgressReport.vue
```diff
@@ -17,20 +17,15 @@
...
         />
 
-        <div v-if="assignmentData?.length">
-          <ProgressStatsOverview
-            v-if="adminStats"
-            :admin-stats="adminStats"
-            :administration-data="administrationData"
-            :tasks-dictionary="tasksDictionary"
-          />
+        <div v-if="studentRows.length">
```
amyyeung17: Minor nitpick. Since we're not styling this, what about using a template?

apps/dashboard/src/containers/ProgressReport/composables/useProgressExport.test.js
```diff
@@ -32,8 +34,8 @@ describe('useProgressExport', () => {
           schoolName: 'Test School',
         },
         progress: {
-          swr: { value: 'completed' },
-          pa: { value: 'started' },
+          't-swr': { value: 'completed' },
```
amyyeung17: Very minor nitpick: According line 24, progress keys are keyed by task UUIDs. Should we generate dummy UUIDs to align with the comment?

apps/dashboard/src/containers/ProgressReport/composables/useProgressExport.js
```diff
@@ -6,27 +7,38 @@ import { exportCsv } from '@/helpers/query/utils';
...
 ) {
-  const buildExportRow = (user, progress) => {
+  // Map each task UUID to a friendly header for the CSV columns.
+  const buildTaskNameByUuid = () =>
+    (toValue(tasks) ?? []).reduce((acc, task) => {
+      acc[task.taskId] = tasksDictionary.value?.[task.taskSlug]?.nameSimple ?? task.taskName;
+      return acc;
+    }, {});
+
+  const buildExportRow = (user, progress, taskNameByUuid) => {
```
amyyeung17, line 41: Are we planning to use `buildExportRow` elsewhere that requires a different `taskNameByUuid`? If not, what are your thoughts about generating the values inside the function itself? (deleting the argument)

apps/dashboard/src/containers/ProgressReport/composables/useProgressColumns.js
```diff
@@ -4,39 +4,79 @@ import { Permissions } from '@bdelab/roar-firekit';
...
+    if (isLoadingTasksDictionary.value || students.value === undefined) return [];
 
     const tableColumns = [];
 
     // Add launch button if user has permission and administration is open
     const { userCan } = usePermissions();
-    const isAdministrationOpen = administrationData.value?.dateClosed
-      ? new Date(administrationData.value?.dateClosed) > new Date()
-      : false;
+    // Open when there's no close date (open indefinitely) or the close date is still in
```
amyyeung17: Hm, maybe I missed something, but I thought administrations had to define a `dateClosed` with the new backend. `dateEnd` is defined as `notNull()` in the schema.

amyyeung17 (changes requested):

apps/dashboard/src/containers/ProgressReport/ProgressReport.vue
richford: Good call. Switched the unstyled studentRows.length wrapper to <template v-if> so it doesn't emit an extra DOM node.

apps/dashboard/src/containers/ProgressReport/composables/useProgressExport.test.js
richford: Done. The fixtures now key progress and the matching mockTasks[].taskId by UUID-shaped ids via shared consts, so they mirror the real shape the comment describes. Assertions are on the resolved friendly headers, so they're unchanged.

apps/dashboard/src/containers/ProgressReport/composables/useProgressExport.js
richford: It's not reused elsewhere, but I'd keep it as an argument deliberately: `buildTaskNameByUuid()` is built once per export in `exportSelected`/`exportAll` and shared across every row. Generating it inside `buildExportRow` would rebuild the UUID→name map for every row (O(rows × tasks), quadratic). So the arg here is hoisting the map build out of the loop, not enabling reuse.

apps/dashboard/src/containers/ProgressReport/composables/useProgressColumns.js
richford: OMG. This revealed an even bigger issue. I was still using the old query that uses Firestore, which is why it was referencing `dateClosed` instead of `dateEnd`. I'll switch it over to use the new query and `dateEnd`. Good catch.

apps/dashboard/src/containers/ProgressReport/composables/useProgressExport.js
amyyeung17: Good point. I glossed over the fact that we're calling this in a mapping function. We could use a closure, but I think that would be an overkill.

apps/dashboard/src/containers/ProgressReport/composables/useProgressColumns.test.js
```diff
@@ -0,0 +1,132 @@
...
+const buildStudents = () =>
+  ref([{ user: { username: 'student1', email: 's1@test.com', firstName: 'John', lastName: 'Doe', grade: '3' } }]);
+
+// Intentionally out of display order so the composable's ordering is exercised.
+const buildTasks = () =>
+  ref([
+    { taskId: 't-mep', taskSlug: 'mep', taskName: 'MEP', orderIndex: 3 },
+    { taskId: 't-pa', taskSlug: 'pa', taskName: 'PA', orderIndex: 2 },
+    { taskId: 't-swr', taskSlug: 'swr', taskName: 'SWR', orderIndex: 0 },
+    { taskId: 't-sre', taskSlug: 'sre', taskName: 'SRE', orderIndex: 1 },
```
amyyeung17: Should we switch these `taskIds` to UUIDs as well?

richford: Ugh. Yes. You are right, of course.

richford: Resolved.

amyyeung17 (approved):

PR #1984, opened 2026-07-07
Score report: cut the individual report (parent path) to the guardian endpoint

richford:
## Summary

This PR completes the individual student report migration by cutting the **parent/guardian path** over to the backend `getGuardianStudentReport` endpoint — the sixth slice of the score-report frontend migration, and the follow-on to the administrator-path PR.

The `StudentScoreReport` container is reached by two routes with two different endpoints. The admin path (org scopes) landed previously; this PR wires the parent path (`orgType === 'family'`) to the longitudinal guardian endpoint and makes the container **uniform** — both paths now feed the score cards from the backend, so the legacy Firestore run-page + longitudinal queries are removed from this container.

> Stacked on `enh/score-report-individual-admin` → `subscores` → `facets` → `students` → `overview`. Merge bottom-up.

## Endpoint

`GET /v1/users/:userId/reports/scores` (`client.users.scoreReports.getGuardianStudentReport`), **no scope params** (authorization is the guardian → child link). Returns `{ student, administrations, longitudinalScores }`: `administrations[]` each carry the same per-task shape as the admin report **minus** per-task `historicalScores`, and the top-level `longitudinalScores` is `Record<taskSlug, HistoricalScore[]>`.

## Changes

- **`useGuardianStudentReportQuery`** (new) — non-paginated composable (token + userId gating, no scope; envelope unwrap, structured errors, pinned retry). Unit test.
- **`StudentScoreReport` container** — for `isParentPath`, fetches the guardian report, selects the administration being viewed from `administrations[]`, and attaches `longitudinalScores[taskSlug]` to each task as `historicalScores` so the guardian tasks match the admin report's per-task shape. Both paths now produce a unified `reportTasks` (→ `useReportCardData` via `ScoreList`) and a `taskData` derived from the completed report tasks. The legacy `useUserRunPageQuery` + `useUserLongitudinalRunsQuery` calls are removed (no longer used here).

## Notes

- **`useReportCardData` is reused unchanged** — the guardian tasks are the same per-task shape (with `historicalScores` re-attached), so the cards render identically to the admin path.
- **`taskData` mapping matches the admin-path PR** (all report tasks, `scores` present only when completed) — kept consistent across both paths rather than narrowing.
[...]

PR #1977, opened 2026-07-07
Score report: add per-task display descriptor to reports/scores

richford:
## Summary

This PR adds a server-computed per-task **`display` descriptor** to the score-report responses so the dashboard can paint scores without knowing anything about scoring versions or per-task display rules. It implements `.do-not-track/ticket-report-display-descriptor.md` and unblocks the admin-metadata swap (the dashboard reads `display` instead of deriving the displayed score type / range from `administrationData.assessments[].params.scoringVersion`).

All per-task/per-version display logic stays in the backend scoring config, where the version/norm logic already lives — the response surfaces the *result*.

## Changes

- **Scoring config (`scoring.config-schema.ts`)** — new optional `displayCategory` (versioned: `normed` | `percentCorrect` | `rawOnly` | `gradeEstimate`) and `displayRanges` (per display score type) blocks.
- **`getScoreDisplay()` (`scoring.service.ts`)** — resolves the display category for the task's scoring version, then picks the score type + value: `percentCorrect` (the percentile field holds the percent for letter/phonics), `rawOnly`/`gradeEstimate` → raw, `normed` → percentile below `percentileBelowGrade` else standard, falling back to raw when the resolved version has no normed fields. Attaches the configured range. Returns `null` for tasks with no display config. Unit tests added.
- **Config data** — `displayCategory` + `displayRanges` authored for `swr`, `sre`, `pa` (normed), `letter`, `phonics` (percent-correct), and `swr-es`, `sre-es` (version-conditional: percent-correct at v0, normed at v1+). Ranges ported from the dashboard's `getRawScoreRange` / `processTaskScores` constants.
- **Contract (`schema.ts`)** — `ScoreDisplaySchema` + an optional `display` on `StudentScoreEntrySchema` and `IndividualStudentReportTaskSchema` (so it flows to the guardian report too via the shared base).
- **Report service** — populates `display` in `buildBaseAssessedTaskEntry` (individual + guardian) and the student-scores builder, from the same rounded scores the response carries. The controller already returns the service result structurally, so no controller change.

## Additive + safe

The `display` field is optional and new; existing `scores` / `supportLevel` are untouched. Tasks without a `displayCategory` simply omit `display`, and the frontend keeps its legacy path for them.

## Flags for review / verification

- **`swr-es` / `sre-es` v0** resolve to `percentCorrect`, whose value comes from the `percentile` field — which doesn't exist at v0, so `value` may be null. The v0 percent-correct value source needs confirmation against the norming tables (flagged in the config comments).
[...]

Issue #1958, opened 2026-07-07
[Epic] Dashboard frontend ts-rest wiring: score reports, participant home, child proxy-launch


richford:
## Summary

Tracks the remaining **dashboard frontend** cutover off firekit/Firestore onto the ts-rest backend, delivered as the stacked/related PRs below (opened from `richford/roar-dashboard` into `project/backend-refactor`). Extends the earlier migration phases (parent portal, student dashboard, consent, progress report, restConfig readiness gate); see the frontend migration status for the full picture.

## Workstreams

**A. Score-report frontend cutover** (continues #1897) — port `ScoreReport.vue` + `StudentScoreReport.vue` off Firestore onto `reports/scores`: overview, student table, distribution facets, subscore tables, individual (admin + parent) reports; move `scoringVersion` display logic into backend `display` / support-threshold descriptors; delete the legacy Firestore composables + the ScoreList legacy path.

**B. Participant home off Firestore** — add `GET /v1/users/:id/memberships` and read the participant homepage profile from `GET /users/:id`, removing the homepage's last Firestore data read.

**C. Child proxy-launch** — wire `can_read_child` into the two administration read endpoints so a parent proxy-launching their child resolves instead of 403-ing, and resolve the participant identity from the child's `launchId`.

## Follow-on / out of scope (tracked separately, not in these PRs)

- **Score-report table base** — `computeAssignmentAndRunData` + the `useAdministrationsQuery → useAdministrationQuery` swap: the last Firestore read in `ScoreReport.vue`, backend-blocked on the student-scores response carrying the row-set special-task fields + capped percentile string.
- Legacy `Task*.vue` assessment players (externally blocked on `@bdelab/*` SDK republish).

## Sub-issues — ordered dependency graph (merge left-to-right within each chain)

**A · score-report backend descriptors:** ~~#1959~~ → ~~#1960~~
**A · score-report frontend:** ~~#1961~~ → ~~#1962~~ → ~~#1963~~ → ~~#1964~~ → #1965 → #1966 → #1967 → ~~#1968~~ → ~~#1969~~ → ~~#1970~~ → ~~#1971~~ → ~~#1972~~
**B · participant home:** ~~#1973~~ → ~~#1974~~
**C · child proxy-launch:** ~~#1975~~, then ~~#1976~~

[...]

PR #1985, opened 2026-07-07
Score report: consume the display descriptor in the report cards

### LINKED ISSUE #1967: Score report: consume the display descriptor in the report cards
## Summary

Consume the backend per-task `display` descriptor (added by the display-descriptor PR) in the individual-report cards, so the **primary** displayed score — which score type, its value, label, and range — comes from the server instead of being re-derived in the browser from `scoringVersion` / task slug.

## Scope

- `containers/StudentScoreReport/components/ScoreList/useReportCardData.js`: when a report task carries `display`, map `display.scoreType` → the card's score slot (`percentCorrect` → the percentile slot), set that score's value, localized name (from `display.label`), and `min`/`max` (from `display.range`), and use it as `scoreToDisplay` — applied before the breakdown rows so dial and breakdown stay consistent. Fall back to the existing `getScoreToDisplay` / `getRawScoreRange` derivation when `display` is absent (not-yet-authored tasks, or older backend).

## Acceptance

- Cards render the same primary score/label/range as before across swr/sre/pa/letter/phonics; not-yet-authored tasks still render via the fallback. Unit tests cover both the display-driven and fallback paths.

> [!IMPORTANT]
> **Dependencies / merge order:** stacked on #1966 (merge that first); runtime dependency on #1959 (must merge first). Part of epic #1958 — see it for the full ordered graph.

richford:
## Summary

First frontend consumption of the backend score **`display`** descriptor (added in `enh/score-report-display-descriptor`). The individual-report cards (`useReportCardData`) now use `task.display` to drive the **primary** displayed score — which score type, its value, label, and range — instead of deriving it client-side from `scoringVersion` / task slug. This starts retiring the dashboard's `scoringVersion` coupling ahead of the admin-metadata swap.

> Stacked on `enh/score-report-individual-parent`. The dashboard is JS, so it reads `task.display` dynamically; when the backend descriptor is absent it falls back to the existing client derivation, so this is safe to land before the backend PR ships.

## Changes

- **`useReportCardData.js`** — when a backend task carries `display`, map `display.scoreType` to the card's score slot (`percentCorrect` → the percentile slot), set that score's `value`, localized `name` (from `display.label`), and `min`/`max` (from `display.range`), and use it as `scoreToDisplay`. Applied **before** the score-breakdown rows are built so the dial and the breakdown stay consistent. Falls back to `getScoreToDisplay` / `getRawScoreRange` when `display` is absent. Unit tests added (display-driven standard score overriding the grade rule; `percentCorrect` → percentile slot with the localized label; the existing grade-≥6 fallback test still passes).

- **Only the primary card score** moves to `display` in this slice. The score-**breakdown** rows for the non-primary score types, and `getTaskDescription` (which still passes `scoringVersion` to the service's swr-es branch), keep the legacy path for now — a tight follow-up once the descriptor proves out.
- **Not-yet-authored tasks** (`fluency`, `roam-alpaca`, etc.) have no backend `display`, so they keep the client derivation via the fallback.
- This does **not** yet swap `useAdministrationsQuery` → `useAdministrationQuery`; that needs the remaining `scoringVersion` users (the `ScoreReport.vue` table columns, the `TaskReport` description, the distribution-chart image — backend ticket item #2, deferred) cut over first.

## Testing

- `useReportCardData` unit tests (display-driven + fallback). `prettier --check` and JS parse pass.
- ⚠️ Manual card QA once the backend `display` PR is deployed: confirm the primary dial score/label/range matches the prior cards across swr/sre/pa/letter/phonics, and that not-yet-authored tasks still render via the fallback.

> [!NOTE]

> [!IMPORTANT]
> **Merge order:** stacked on `enh/score-report-individual-parent` (#1984) — merge #1984 first; until it lands, this PR's diff sits on top of that branch's commits.
>
> **Runtime dependency:** merge #1977 first.

Resolves #1967
Part of #1958

### DIFF
--- a/apps/dashboard/src/components/reports/DistributionChartFacet.vue
+++ b/apps/dashboard/src/components/reports/DistributionChartFacet.vue
@@ -54,9 +54,13 @@ const props = defineProps({
       return { name: 'Grade', key: 'grade' };
     },
   },
-  runs: {
-    type: Array,
-    required: true,
+  facets: {
+    // Per-task facet aggregation from `getScoreFacets` (scoreBinsByGrade /
+    // scoreBinsBySchool). Each bin is `{ binStart, binEnd, count }`, already
+    // computed server-side. Null while the query is loading.
+    type: Object,
+    required: false,
+    default: null,
   },
   minGradeByRuns: {
     type: Number,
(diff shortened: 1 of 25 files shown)


Emily-ejag:
Closing in favor of #2086

PR #1986, opened 2026-07-07
Score report: drive table column choice from the display descriptor

### LINKED ISSUE #1968: Score report: drive table column choice from the display descriptor
## Summary

Drive the score-report **table** column choice from the backend `display` descriptor instead of `scoringVersion`. Specifically, the `swr-es` "is it normed?" decision in `scoreReportColumns` (which picks the standard/raw column vs. percent-correct) should read the students-table `display.scoreType`, not `getScoringVersions`.

## Scope

- `pages/ScoreReport.vue`: derive a per-task `display.scoreType` map from the students response and decide `swr-es` normed-vs-percent-correct from it, falling back to the `scoringVersion` rule when the descriptor isn't present. No change to any other column.

## Acceptance

- The `swr-es` column in the Standard-Score and Raw-Score views matches today's behavior for both a normed (v1+) and an un-normed (v0) administration; all other tasks' columns unchanged. Byte-for-byte equivalent to the old logic under the fallback.

> [!IMPORTANT]
> **Dependencies / merge order:** stacked on #1967 (merge that first). Part of epic #1958 — see it for the full ordered graph.

richford:
## Summary

Wires the score-report **table columns** to the backend `display` descriptor: the `swr-es` "is it normed?" decision in `scoreReportColumns` (which picks standard/raw score vs. percent-correct in the standard/raw views) now comes from the students-table `display.scoreType` instead of `scoringVersion`, with a fallback to `scoringVersion` when the backend descriptor isn't available. This is the table's counterpart to the card consumption, and a step toward dropping the `administrationData.assessments` (`scoringVersion`) dependency that blocks the admin-metadata swap.

> Stacked on `enh/score-report-display-consumption`. JS, so it reads `entry.display` dynamically and falls back when absent — safe before the backend `display` PR ships.

## Changes

- **`ScoreReport.vue`** — adds `displayScoreTypeBySlug` (per-task backend display score type, uniform per administration, taken from any student's `display` descriptor) and an `isSwrEsNormed(taskId)` helper (`display.scoreType !== 'percentCorrect'`, falling back to `getScoringVersions['swr-es'] >= 1`). The two `scoreReportColumns` `swr-es` checks now call `isSwrEsNormed(taskId)`.

- **Only the column-field decision** moves to `display`. The CSV export (`createExportData`) and the soon-to-be-deleted `computeAssignmentAndRunData` client path still read `getScoringVersions` for the same `swr-es` check — tight follow-ups (the export) and cleanup (the client path). So this PR does **not** yet remove `getScoringVersions` / the `useAdministrationsQuery` dependency; that also needs the export, the `TaskReport` description, and the distribution-chart image (backend ticket #2) cut over.

## Behavior / fallback

`swr-es` scoring version is uniform per administration, so taking `display.scoreType` from any completed student's entry is well-defined. When no student has a `display` for `swr-es` (backend not deployed, or no completed `swr-es` run), `isSwrEsNormed` falls back to `getScoringVersions['swr-es'] >= 1` — identical to today. So this is a no-op until the backend ships, then becomes backend-authoritative.

## Testing

- `prettier --check` and `@vue/compiler-sfc` compile pass.
- ⚠️ Manual QA once the backend `display` PR is deployed: for an administration with `swr-es`, confirm the `swr-es` column in the **Standard Score** and **Raw Score** views shows the normed score vs. percent-correct correctly for both a normed (v1+) and an un-normed (v0) administration, and that other tasks' columns are unchanged.

> [!NOTE]

> [!IMPORTANT]
> **Merge order:** stacked on `enh/score-report-display-consumption` (#1985) — merge #1985 first; until it lands, this PR's diff sits on top of that branch's commits.

Resolves #1968
Part of #1958

### DIFF
--- a/apps/dashboard/src/components/reports/DistributionChartFacet.vue
+++ b/apps/dashboard/src/components/reports/DistributionChartFacet.vue
@@ -54,9 +54,13 @@ const props = defineProps({
       return { name: 'Grade', key: 'grade' };
     },
   },
-  runs: {
-    type: Array,
-    required: true,
+  facets: {
+    // Per-task facet aggregation from `getScoreFacets` (scoreBinsByGrade /
+    // scoreBinsBySchool). Each bin is `{ binStart, binEnd, count }`, already
+    // computed server-side. Null while the query is loading.
+    type: Object,
+    required: false,
+    default: null,
   },
   minGradeByRuns: {
     type: Number,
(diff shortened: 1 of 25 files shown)


Emily-ejag:
Close in favor of #2089

Emily-ejag:
Close in favor of https://github.com/yeatmanlab/roar-dashboard/pull/2089

PR #1987, opened 2026-07-07
Score report: drive CSV export swr-es/sre-es from the display descriptor

### LINKED ISSUE #1969: Score report: drive CSV export swr-es/sre-es from the display descriptor
## Summary

Cut the score-report **CSV export** off `scoringVersion`: the `swr-es` and `sre-es` normed-vs-percent-correct branch choices in `createExportData` should come from the backend `display` descriptor (generalized `isSpanishNormed` helper), not `getScoringVersions`. After this, `createExportData` no longer reads `getScoringVersions` at all.

## Scope

- `pages/ScoreReport.vue`: generalize the table-display helper to `isSpanishNormed(taskId)` (guarded to `['swr-es','sre-es']`), and use it at both `createExportData` branch guards (percent-correct / correct-incorrect-difference), falling back to `scoringVersion` when the descriptor is absent.

## Acceptance

- Export of an administration containing `swr-es`/`sre-es` emits the same columns as today for both normed and un-normed admins; `letter`/`phonics` percent-correct columns unchanged. Equivalent to the old logic under the fallback.

> [!IMPORTANT]
> **Dependencies / merge order:** stacked on #1968 (merge that first). Part of epic #1958 — see it for the full ordered graph.

richford:
## Summary

Cuts the score-report **CSV export** off `scoringVersion`: the `swr-es` and `sre-es` "is it normed?" decisions in `createExportData` now come from the students-table backend `display` descriptor instead of `getScoringVersions`, with a fallback to `scoringVersion` when the descriptor isn't available. After this slice `createExportData` no longer reads `getScoringVersions` at all — another step toward dropping the `administrationData.assessments` dependency that blocks the admin-metadata swap.

> Stacked on `enh/score-report-table-display`. JS, so it reads `entry.display` dynamically and falls back when absent — safe before the backend `display` PR ships.

## Changes

- **`ScoreReport.vue`** — generalizes the table-display helper `isSwrEsNormed` → **`isSpanishNormed(taskId)`**, guarded to `SPANISH_NORMED_TASKS = ['swr-es', 'sre-es']` (the two tasks whose normed-vs-percent-correct presentation is version-conditional; both carry a backend `display` descriptor). The helper reads `displayScoreTypeBySlug[taskId]` (`!== 'percentCorrect'` ⇒ normed) and falls back to `getScoringVersions[taskId] >= 1`.
  - The two `createExportData` branch guards now call `!isSpanishNormed(taskId)`: the percent-correct branch (`swr-es`) and the correct/incorrect-difference branch (`sre-es`).
  - The two `scoreReportColumns` sites are renamed `isSwrEsNormed` → `isSpanishNormed` (behaviour-identical — see below).

## Why generalize instead of swr-es only

`createExportData` had **two** `scoringVersion` guards two lines apart: `swr-es` (percent-correct branch) and `sre-es` (correct/incorrect-difference branch). They share the exact same "normed when version ≥ 1" semantics that `useReportCardData` already encodes as `useSpanishNorms = (swr-es | sre-es) && version >= 1`, and both tasks get a `display` descriptor from the backend (`swr-es.ts` / `sre-es.ts` configs). Migrating only `swr-es` would leave the export still bound to `getScoringVersions` for `sre-es` — an arbitrary half-measure. Doing both clears `getScoringVersions` out of the export in one coherent change.

## Equivalence (provable, all four call sites)

`isSpanishNormed` returns the **same value as the old `isSwrEsNormed` for every task except `sre-es`**, and `sre-es` only reaches the new branch where the old code already keyed on `scoringVersion`:

- **Column sites** are only evaluated when `tasksToDisplayPercentCorrect.includes(taskId)`. `sre-es` is **not** in that list, so the helper there is only ever called for `swr-es` (identical logic) or non-Spanish percent-correct tasks (`letter`, `phonics`, … → `false`, exactly as the old `taskId !== 'swr-es'` guard). No column behaviour changes.
- **Export `swr-es` branch**: `!isSpanishNormed('swr-es')` under fallback = `!(getScoringVersions['swr-es'] >= 1)` = old `!(taskId === 'swr-es' && version >= 1)`. Other percent-correct tasks → `!false` = `true`, unchanged.
- **Export `sre-es` branch**: `tasksToDisplayCorrectIncorrectDifference = ['sre-es']`, so only `sre-es` reaches it. `!isSpanishNormed('sre-es')` under fallback = `!(getScoringVersions['sre-es'] >= 1)` = old guard.

So this is a **no-op until the backend `display` ships**, then becomes backend-authoritative. The non-Spanish percent-correct tasks (`letter`, `phonics`, `morphology`, …) that are *always* percent-correct never consult `display` and never flip — `SPANISH_NORMED_TASKS` is the guard that preserves that.

- This slice removes `getScoringVersions` from `createExportData` only. The remaining dashboard `scoringVersion` users that still block dropping `useAdministrationsQuery` → `useAdministrationQuery`: the `TaskReport` task description (`:task-scoring-versions` prop), the soon-to-be-deleted `computeAssignmentAndRunData` client path (removed in the cleanup PR), and the distribution-chart image (backend ticket #2, deferred).
- Export **values** (`score.percentCorrect` / `numAttempted` / `numCorrect` / `correctIncorrectDifference`) still come from the client-computed special-task fields the overlay preserves; only the **branch selection** moves to `display`. Same scope boundary as the table-display slice.

## Testing

- `prettier --check` and `@vue/compiler-sfc` compile pass.
- ⚠️ Manual QA once the backend `display` PR is deployed: export an administration containing `swr-es` and `sre-es` and confirm a normed (v1+) admin emits the standard/raw columns while an un-normed (v0) admin emits the percent-correct / correct-incorrect-difference columns, matching today; confirm `letter`/`phonics` percent-correct columns are unchanged.

> [!NOTE]

> [!IMPORTANT]
> **Merge order:** stacked on `enh/score-report-table-display` (#1986) — merge #1986 first; until it lands, this PR's diff sits on top of that branch's commits.

Resolves #1969
Part of #1958

### DIFF
--- a/apps/dashboard/src/components/reports/DistributionChartFacet.vue
+++ b/apps/dashboard/src/components/reports/DistributionChartFacet.vue
@@ -54,9 +54,13 @@ const props = defineProps({
       return { name: 'Grade', key: 'grade' };
     },
   },
-  runs: {
-    type: Array,
-    required: true,
+  facets: {
+    // Per-task facet aggregation from `getScoreFacets` (scoreBinsByGrade /
+    // scoreBinsBySchool). Each bin is `{ binStart, binEnd, count }`, already
+    // computed server-side. Null while the query is loading.
+    type: Object,
+    required: false,
+    default: null,
   },
   minGradeByRuns: {
     type: Number,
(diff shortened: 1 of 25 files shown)


Emily-ejag:
Closing in favor of #2090

PR #1993, opened 2026-07-07
Wire can_read_child into the two administration read endpoints

### LINKED ISSUE #1975: Wire can_read_child into the two administration read endpoints
> ⚠️ **SECURITY-CRITICAL — REQUIRES LINE-BY-LINE HUMAN REVIEW.** This widens who can read a child's administration and agreement data. Per `culture-leverage-ai.md`, authorization changes are exactly the class of AI-generated code that looks right but can be subtly wrong (missing super-admin bypass, wrong FGA relation, 404-vs-403 ordering, embed leakage). The unit tests mock FGA, so they verify the wiring, not the policy. **Integration tests against the real OpenFGA model are required.**

When a parent launches an assessment on behalf of their child (the "proxy-launch" flow), the dashboard renders the child's homepage and consent gate by calling two backend read endpoints **as the parent, for the child's `userId`**:

- `GET /v1/users/:userId/administrations` (`getUserAdministrations`)
- `GET /v1/users/:userId/administrations/:administrationId/agreements` (`listUserAdministrationAgreements`)

Both currently authorize only three ways: super admin, self (`requesterUserId === userId`), or a requester who holds administration-scoped `can_read` **and** is an admin/teacher whose accessible-administration set intersects the target's. A parent is none of these for their child, so both endpoints return `403` and the child's portal cannot load under a parent session.

Run creation and run events are already authorized for this flow via `can_create_run_for_child`, and parent **consent recording** is already authorized via `can_consent_for_child` (`user.service.recordUserAgreement`). The gap is purely the two **reads** that the homepage and consent gate depend on.

### Proposed change
Accept a fourth authorization path on these two reads: a requester who holds the FGA relation **`can_read_child`** on the target child's family. This mirrors the existing parent-consent authorization already in `recordUserAgreement`.

### Scope
- `getUserAdministrations`: add the guardian path to the non-super-admin / non-self branch, ahead of the admin/teacher intersection fallback.
- `listUserAdministrationAgreements`: accept the guardian path in place of the administration-level `can_read` fallback.
- No contract, route, controller, or schema changes — service-layer authorization only.
- `recordUserAgreement` is **out of scope** (already supports `can_consent_for_child`).

### Acceptance criteria
- [x] A parent with `can_read_child` on the child's family can read that child's administrations and per-administration agreements.
- [x] A parent **without** `can_read_child` on that family still gets `403` (no cross-family leakage).
- [x] Super-admin and self paths are unchanged; the admin/teacher intersection path is unchanged.
- [x] Embed output on `getUserAdministrations` remains scoped to the **child** (no requester-scoped stats or per-student task state).
- [x] **Integration tests against the real FGA model** cover allow (guardian-on-family) and deny (non-guardian, wrong family) for both endpoints.
- [x] Lint, format, type-check, unit + integration tests pass.

Parent: #1897

> [!IMPORTANT]
> **Dependencies / merge order:** first in its stack — targets `project/backend-refactor` directly. Part of epic #1958 — see it for the full ordered graph.

richford:
## Summary

This PR wires the FGA `can_read_child` relation into the two administration **read** endpoints a parent hits when proxy-launching their child, so the child's homepage and consent gate resolve under a parent session instead of `403`-ing. It is a service-layer authorization change only — no contract, route, controller, or schema changes.

> ⚠️ **This is a security-sensitive authorization change. Please review the three points below line-by-line and rely on the integration tests, not the (FGA-mocked) unit tests, for the policy guarantee.**

## What was missing

`getUserAdministrations` and `listUserAdministrationAgreements` authorized only super-admin, self, or an admin/teacher whose accessible-administration set intersects the target's. A parent reading their own child is none of those, so the proxy-launch homepage and consent gate returned `403`. Run creation (`can_create_run_for_child`) and parent consent recording (`can_consent_for_child`) were already authorized; only these two reads were not.

## Changes

- **`hasGuardianReadAccess(requesterUserId, targetUserId)`** — a new private, **non-throwing** helper in `administration.service.ts`. It reads the target user's entity memberships, keeps the `family` ones, and asks FGA whether the requester has `can_read_child` on any of those family objects (`hasAnyPermission`). It returns a boolean (rather than throwing) so callers can fall through to other authorization paths. This mirrors the parent block already in `user.service.recordUserAgreement`.
- **`getUserAdministrations`** — in the non-super-admin / non-self branch, the guardian path is tried **before** the existing admin/teacher intersection fallback. When the requester is an authorized guardian, the target child's administrations are returned directly (by IDs), bypassing the requester-intersection logic that exists for admins/teachers.
- **`listUserAdministrationAgreements`** — inside the existing `!isSuperAdmin && requesterUserId !== userId` block, an authorized guardian is accepted in place of the administration-level `can_read` requirement.

## Authorization behavior (please verify each)

1. **Super-admin bypass is unchanged and still checked first.** The guardian path lives only in the non-super-admin, non-self branch.
2. **404-before-403 ordering is preserved.** `listUserAdministrationAgreements` still verifies the child and the administration exist before the guardian check runs.
3. **Correct FGA relation and object.** Access is `can_read_child` on `family:<id>` for each family the **child** belongs to — not an administration-scoped relation, and keyed to the child's family, not the requester's.

## Why the guardian sees the child's data, not their own (embed scoping)

`getUserAdministrations` resolves embeds after the authorization decision. Stats are gated to super-admins only (a guardian never receives them). The tasks embed is fetched unfiltered (its `userId` argument is only error-logging context), and per-student task state is attached **keyed to the target user** (the child). So the guardian path returns the child's administrations and the child's per-task state. Confirm this in review against the embed-resolution code.

## Testing

- **Unit tests added** for both endpoints (guardian-allowed). These mock FGA, so they assert the **wiring**: that `hasAnyPermission` is called with `('<requester>', 'can_read_child', ['family:<childFamily>'])`, that the self/admin paths are not taken, and that the child's data is returned.
- ⚠️ **Integration tests against the real OpenFGA model are still required and were not run in the authoring environment.** Before merge, please add/verify integration coverage for: (a) a guardian with `can_read_child` on the child's family is **allowed** on both endpoints; (b) a user **without** that relation, and a guardian of a **different** family, are **denied** (`403`); (c) the super-admin, self, and admin/teacher paths are unaffected.
- Lint, format, and `check-types` were run for the changed backend files.

## Out of scope

- `recordUserAgreement` (already authorizes parents via `can_consent_for_child`).
- The dependent **frontend** proxy-launch cutover (`enh/proxy-launch-frontend`), which builds on this PR.

> [!NOTE]

> [!IMPORTANT]
> **Merge order:** first in its stack — targets `project/backend-refactor` directly.

Resolves #1975
Part of #1958

### DIFF
--- a/apps/backend/src/services/administration/administration.service.test.ts
+++ b/apps/backend/src/services/administration/administration.service.test.ts
@@ -3107,6 +3107,48 @@ describe('AdministrationService', () => {
       expect(result.totalItems).toBe(2);
     });
 
+    it('returns the target child administrations for an authorized guardian (can_read_child), bypassing the requester intersection', async () => {
+      // Proxy-launch: a parent with `can_read_child` on the child's family lists the
+      // child's administrations. The parent holds no administration-level access of
+      // their own, so the admin/teacher intersection would be empty — the guardian
+      // path returns the child's full accessible set instead.
+      const child = UserFactory.build({ id: 'child-1', rosteringEnded: null });
+      mockUserRepository.getById.mockResolvedValue(child);
+
+      const childAdmins = AdministrationFactory.buildList(2);
+      mockAuthorizationService.listAccessibleObjects.mockResolvedValue(
+        childAdmins.map((a) => `administration:${a.id}`),
+      );
+      mockUserRepository.getUserEntityMemberships.mockResolvedValue([{ entityType: 'family', entityId: 'family-1' }]);
+      mockAuthorizationService.hasAnyPermission.mockResolvedValue(true);
+      mockAdministrationRepository.getByIds.mockResolvedValue({ items: childAdmins, totalItems: 2 });
+
+      const service = AdministrationService({
+        administrationRepository: mockAdministrationRepository,
+        userRepository: mockUserRepository,
+        authorizationService: mockAuthorizationService,
+      });
+
+      const result = await service.getUserAdministrations({ userId: 'parent-1', isSuperAdmin: false }, 'child-1', {
+        page: 1,
+        perPage: 25,
+        sortBy: 'createdAt',
+        sortOrder: 'desc',
+      });
+
+      // Guardianship is checked via `can_read_child` on the child's family.
+      expect(mockAuthorizationService.hasAnyPermission).toHaveBeenCalledWith('parent-1', 'can_read_child', [
+        'family:family-1',
+      ]);
+      // Only the target's accessible-objects lookup runs — no requester intersection.
+      expect(mockAuthorizationService.listAccessibleObjects).toHaveBeenCalledTimes(1);
+      expect(mockAdministrationRepository.getByIds).toHaveBeenCalledWith(
+        childAdmins.map((a) => a.id),
+        expect.objectContaining({ page: 1, perPage: 25 }),
+      );
+      expect(result.items).toHaveLength(2);
+    });
+
     it('throws 404 when a rostering-ended user requests their own administrations (#1742 defense-in-depth)', async () => {
       // The auth guard (#1735) blocks rostering-ended users end-to-end, but
       // the service-layer check runs BEFORE the `requesterUserId === userId`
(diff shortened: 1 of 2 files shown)


amyyeung17:
Closed in favor of https://github.com/yeatmanlab/roar-dashboard/pull/2069

PR #2083, opened 2026-07-28
enh(dashboard): Seed a default administration with dashboard-launchable tasks in dev environment

ksmontville:
## Summary

This PR makes a greenfield local environment capable of launching an assessment from the dashboard. Two seeding paths existed and never met: `dev:seed` creates administrations whose tasks come from `TaskFactory` with generated slugs (`task-3-a3f9k2`), while `dev:seed:tasks` creates real tasks and variants but never writes `administration_task_variants`. No seeded administration had ever contained a launchable task, so exercising the dashboard launch path — and therefore the run, trial, and score write path — required hand-written SQL.

## What changed

- **`defaultVariant` on `TaskSeedConfig`**, set for SWR, SRE, Letter, and PA. Naming
  the variant in the config rather than taking the first entry of
  `taskVariantParameters.json` keeps the choice deterministic and platform-owned.
- **`dev:seed:tasks` assigns that variant** to a new **Assessment Launch Sandbox**
  administration. Guarded on the fixture existing, so the assessment e2e stack
  (migrations + task seed, no fixture) is unaffected. `--no-assign` opts out.
- **`dev:init` seeds the four defaults**, copying `taskVariantParameters.example.json`
  where absent, so the documented setup command lands on a launchable environment.
- **`dev:assign:variant`** for any other administration/variant pair, including
  non-fixture databases.
- **`--refresh-params`** pushes edited parameters into an existing variant, which
  skip-if-present seeding otherwise made impossible.
- **Task metadata is now reconciled** rather than insert-only, so config changes reach
  rows seeded earlier. Variants keep skip-if-present semantics — they carry runs.

## Fixture fixes

- The sandbox is a **separate administration**, so the existing ones keep their synthetic tasks, which drive the progress and score fixtures.
- Consent/assent versions pointed at placeholder GitHub coordinates that 404, so the content endpoint 500'd and the participant's task list never rendered. Now pinned to real public documents.
- Seeded tasks carry their production tile artwork, from a host already allowlisted in `img-src`.

## Verified

`dev:init` from a truncated database: four tasks assigned, agreements resolve, tasks launch and persist runs, trials, and scores. Re-runs are idempotent.


ksmontville:
@claude Let's review.

ksmontville:
@claude Fixed, let's review again.

apps/backend/seeds/task-seed.ts
```diff
@@ -148,32 +169,55 @@ const db = drizzle(pool, { schema: CoreDbSchema, casing: 'snake_case' });
...
-      nameTechnical: meta.nameTechnical,
-      taskConfig: {},
-    })
-    .onConflictDoNothing()
-    .returning();
+  const metadata = {
+    name: meta.name,
+    nameSimple: meta.nameSimple,
+    nameTechnical: meta.nameTechnical,
+    ...(meta.image ? { image: meta.image } : {}),
```
amyyeung17, line 194: I noticed there were few more nullable fields like `description` and `tutorialVideo` in `backend/db/schema/core/task.ts`. Do we want insert them here as well?

kellyel (dismissed):

amyyeung17 (changes requested):
Are we allowing users to assign multiple variants of the same task to an administration? I might have missed where the code handles this or when we decided to change this behavior.

apps/backend/seeds/task-seed.ts
ksmontville: Sure, we can add those.

ksmontville:
> Are we allowing users to assign multiple variants of the same task to an administration? I might have missed where the code handles this or when we decided to change this behavior.

Not that I'm aware of. Is there an example of that happening in this PR?

Emily-ejag (dismissed):

amyyeung17 (dismissed):

amyyeung17:
> Are we allowing users to assign multiple variants of the same task to an administration? I might have missed where the code handles this or when we decided to change this behavior.

Update: established out-of-scope at scrum. For now, we let the task picker enforce this constraint.

Emily-ejag (approved):

